PR #2, opened 2016-12-13
Basic users

2e2a:
Test-URLs: `/accounts/login/`,`/accounts/signup/`,`/accounts/logout/`

### DIFF
--- a/apps/users/models.py
+++ b/apps/users/models.py
@@ -1,0 +1,56 @@
+from django.contrib.auth import models as auth_models
+from django.core import validators
+from django.db import models
+from django.utils import timezone
+from django.utils.translation import ugettext_lazy as _
+
+from . import USERNAME_INVALID_MESSAGE
+from . import USERNAME_REGEX
+
+
+class User(auth_models.AbstractBaseUser, auth_models.PermissionsMixin):
+    username = models.CharField(
+        _('username'),
+        max_length=60,
+        unique=True,
+        help_text=_(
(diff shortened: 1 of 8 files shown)


meinberlin/settings.py
```diff
@@ -57,8 +57,10 @@
     'rest_framework',
     'allauth',
     'allauth.account',
+    'allauth.socialaccount',
 
     'apps.cms',
+    'apps.users.apps.UsersConfig',
```
xi: why not simply `apps.users`?

meinberlin/urls.py
```diff
@@ -11,6 +11,7 @@
     url(r'^django-admin/', include(admin.site.urls)),
 
     url(r'^admin/', include(wagtailadmin_urls)),
+    url(r'^accounts/', include('allauth.urls')),
```
xi: I think I would prefer to to have them on top level, e.g. `/login` instead of `/accounts/login`.

apps/users/models.py
```diff
@@ -0,0 +1,56 @@
...
+        error_messages={
+            'unique': _('A user with that email address already exists.')}
+    )
+
+    is_staff = models.BooleanField(
+        _('staff status'),
+        default=False,
+        help_text=_(
+            'Designates whether the user can log into this admin site.')
+    )
```
xi: What exactly is the use case for this? It sounds very specific.

PR #7, opened 2017-01-16
setup webpack

xi:
This sets up webpack, similar to how it is done in https://github.com/liqd/euth_wagtail/.

I am not totally sold on webpack yet.

- Pros
  - used in euth_wagtail
  - good support for es6, jsx, sass, postcss and many other tools
  - powerful dependency management
- Cons
  - complicated
  - slow

My proposal is to merge this now to get started and to switch to something else if/when we find a better solution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,9 @@
 *__pycache__/
 db.sqlite3
 node_modules
-meinberlin/static/style.css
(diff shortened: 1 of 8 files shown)


xi:
I forgot on detail: This does not yet work because adhocracy4 still has some hardcoded occurances of `opin`.

PR #15, opened 2017-01-18
use a4 permission view

2e2a:
The a4 core permission view does not redirect to login, if permissions are denied.


xi:
Does it never redirect to login or only if the user is not yet logged in?

2e2a:
> Does it never redirect to login or only if the user is not yet logged in?

It still does redirect if the user is not logged in.

PR #30, opened 2017-01-31
Add python docstring linter

xi:
http://www.pydocstyle.org/


setup.cfg
```diff
@@ -1,2 +1,3 @@
 [flake8]
+max-complexity = 10
```
2e2a: Interresting

PR #31, opened 2017-01-31
refactor CSS

xi:
- use contrast-color
- add some variables
- style some minor elements


2e2a:
There is still a hard-coded font-size in `_base.scss` and a padding in `_label.scss`.
Rest looks fine.

xi:
The font-size in `_base.scss` defines the root font size which the others are relative to. `$padding` is meant for layout (e.g. gutters; maybe it should be renamed) so it is not a problem if there are non-variable paddings somewhere.

PR #32, opened 2017-02-01
pagination

xi:
Note that ddfdf04 uses the `contrast-color()` function from #31. There is no build error, but the styling looks a bit off without it.

The UI can be expanded later (e.g. previous/next links) but I went with a minimal solution for now.

To test, it is easiest to reduce `IdeaListView.paginate_by` to 1.


2e2a:
looks good

PR #33, opened 2017-02-02
basic document commenting

2e2a:


apps/documents/models.py
```diff
@@ -0,0 +1,58 @@
...
+    comments = GenericRelation(comment_models.Comment,
+                               related_query_name='document',
+                               object_id_field='object_pk')
+
+    def __str__(self):
+        return "{}_document_{}".format(str(self.module), self.pk)
+
+    def get_absolute_url(self):
+        from django.core.urlresolvers import reverse
+        return reverse('project-detail', args=[str(self.project.slug)])
```
xi: Why `project-detail`?

apps/documents/rules.py
```diff
@@ -0,0 +1,16 @@
...
+from adhocracy4.modules.predicates import is_public_context
+
+
+rules.add_perm('meinberlin_documents.view',
+               is_superuser | is_context_moderator | is_context_initiator |
+               is_context_member | is_public_context)
+
+
+rules.add_perm('meinberlin_documents.create',
+               is_superuser | is_context_moderator | is_context_initiator)
```
xi: Why does a moderator have this permission?

apps/documents/templates/meinberlin_documents/document_detail.html
```diff
@@ -0,0 +1,39 @@
...
+<div class="document">
+    <article>
+        <header>
+            <h1>{{ object.name }}</h1>
+        </header>
+
+        {% for paragraph in object.paragraphs.all %}
+
+        <section class="paragraph">
+            <h3 class="paragraph-title">
```
xi: needs to be `<h2>`

apps/documents/templates/meinberlin_documents/paragraph_detail.html
```diff
@@ -0,0 +1,28 @@
...
+                </li>
+            </ul>
+        </nav>
+    </div>
+
+    <article class="paragraph">
+        <h1 class="paragraph-title">
+            {{ paragraph.name }}
+        </h1>
+        <p>{{ paragraph.text | safe }}</p>
```
xi: Should not be wrapped in `<p>` because it contains its own markup.

apps/documents/models.py
xi: Nevermind, I found out myself: The document is displayed on the project page.

apps/documents/rules.py
2e2a: > Why does a moderator have this permission?

Can think about it. Makes sense, somehow, as there could be one initiator that created the project, but multiple user contributing to the document.

PR #34, opened 2017-02-06
pagination

xi:
note that the styling uses the `contrast-color()` function that is introduced
in #31. Sass will just ignore it and output it as string, so there is no build
error, only broken styling.


apps/documents/templates/meinberlin_documents/document_detail.html
```diff
@@ -8,24 +8,25 @@
...
-            <h2 class="paragraph-title">
-                {{ paragraph.name }}
-            </h2>
-            <div class="paragraph-comments">
+        <section class="commenting">
+            <div class="commenting-content">
+                <h2 class="commenting-title">
+                    {{ paragraph.name }}
+                </h2>
+                <div class="paragraph-body">
```
2e2a: Should be `commenting-body`? But `commenting-body` is also not defined.

2e2a: > Should be commenting-body? But commenting-body is also not defined.

I'll remove it, as not needed.

PR #39, opened 2017-02-08
Document edit UI

xi:
This adds UI to create/edit a document. The editor is implemented as a react app.

Note that the editor is not avilable in a special edit view. Instead, it replaces the document detail view in the `create_document` phase.

The code is very similar to what is used in euth_wagtail. I mostly adapted the UI and did some minor refactoring to increase readability. I also tried to reduce flickering by only reloading ckeditor in certain cases (see a4fc734). @slomo, @vellip what do you think about the approach?

### DIFF
--- a/apps/documents/api.py
+++ b/apps/documents/api.py
@@ -1,0 +1,18 @@
+from rest_framework import mixins
+from rest_framework import permissions
+from rest_framework import viewsets
+
+from adhocracy4.api.permissions import IsModerator
+
+from .models import Document
+from .serializers import DocumentSerializer
+
+
+class DocumentViewSet(mixins.CreateModelMixin,
+                      mixins.UpdateModelMixin,
+                      viewsets.GenericViewSet):
+    queryset = Document.objects.all()
+    serializer_class = DocumentSerializer
+    permission_classes = (
+        permissions.IsAuthenticated, IsModerator,
+    )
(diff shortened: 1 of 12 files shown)


slomo:
The version proposed by @xi works on my device. I think the only issue here, is that we are exploiting how the react virtual dom / dom sync works right now. Since this is probably undefined behaviour, it could beak on any react update.

In my opinion we should do the following: We use this optimazation, but try to watch out for breakage after react updates. I would also like to put the widget it self into core.

@xi Do you think it makes sense to spend more time on trying to isolate the CKEditor into its own component? We could also look on other approaches like this [stackoverflow post](http://stackoverflow.com/questions/36535234/), which basically loads the CKEditor only after the user clicked on the paragraph.

slomo:
I just tested with @vellip the inline ckeditor (a content editable div), maybe that would be an even better alternative, since it doesn't flicker at all (because it only attaches to the preexisting diff).

xi:
> In my opinion we should do the following: We use this optimazation, but try to watch out for breakage after react updates.

Yes, we definitely have to check on every update.

> Do you think it makes sense to spend more time on trying to isolate the CKEditor into its own component?

I tried putting it into its own component with `shouldComponentUpdate` always returning `false`. Did not work.

> We could also look on other approaches like this stackoverflow post, which basically loads the CKEditor only after the user clicked on the paragraph.

That could be beneficial also in terms of UX: Having all these editors all the time is very distracting. On the other hand, I usually find these kinds of UI frustrating to use (e.g. taiga task descriptions). It may be even better to have a separate "edit" button for each paragraph that turns it into an editor.

I also found http://alloyeditor.com/ which is based on CKEditor, but have not looked into it.

Going forward, I would suggest to merge this now (after review) and find an even better solution separately.

PR #40, opened 2017-02-08
Implement full-width layout

xi:
This makes first steps in implementing a new layout where many parts of the UI stretch the full width of the UI. This style requires that layout is not only handled in the base.html file, but scattered throughout the application. So the main challange here is maintainability.

The new layout consists of rows that stretch the full width of the screen. Inside of each row, there is a `div.l-wrapper` that sets the general width. Inside of that, you can optionally use the grid (e.g. using the `.l-center-8` or `.l-center-6` classes). Finally, the content is added inside of that.

The rows tend to have limited semantic value and are mostly used to supply a different background color or a large image. A classic example of a row would be the resource header with a full-width image.

### DIFF
--- a/apps/cms/templates/meinberlin_cms/home_page.html
+++ b/apps/cms/templates/meinberlin_cms/home_page.html
@@ -10,8 +10,11 @@
 {% endblock %}
 
 {% block content %}
-    <h1>{{ self.title }}</h1>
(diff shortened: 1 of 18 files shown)


xi:
Note: This breaks the layout on the login/register views

PR #43, opened 2017-02-13
Adapt to (proposed) HTML changes in core

xi:
I did two pull request to core that change HTML:

- https://github.com/liqd/adhocracy4/pull/25 (ratings, already merged)
- https://github.com/liqd/adhocracy4/pull/23 (comments, not yet merged)

Both require adjustments in our CSS, which is done in this pull-request.

Note that I added the `dropdown-item` class from bootstrap4 in 0cb42db. This is not yet used (because it would also need to be added in core) but may allow us to simplify the code in the future.

### DIFF
--- a/apps/users/templates/meinberlin_users/indicator.html
+++ b/apps/users/templates/meinberlin_users/indicator.html
@@ -20,7 +20,7 @@
         {% for organisation in request.user.organisations %}
 
         <li>
-            <a href="{% url 'dashboard-project-list' organisation_slug=organisation.slug %}">
+            <a class="dropdown-item" href="{% url 'dashboard-project-list' organisation_slug=organisation.slug %}">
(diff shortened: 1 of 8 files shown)


apps/users/templates/meinberlin_users/indicator.html
```diff
@@ -31,7 +31,7 @@
             <form class="form--inline" action="{% url 'account_logout' %}" method="post">
                 {% csrf_token %}
                 <input type="hidden" name="next" value="{{ redirect_field_value }}">
-                <button type="submit" class="button--as-link">{% trans "Logout" %}</button>
+                <button type="submit" class="dropdown-item button--as-link">{% trans "Logout" %}</button>
```
xi: `button--as-link` should not be required here anymore.

PR #44, opened 2017-02-13
Initiator dashboard

2e2a:


xi:
Nice! I have not done in-depth testing yet. Instead I will merge this so we can continue working on it without major conflicts.

PR #53, opened 2017-02-14
Improve error reporting in document edit app

xi:
@slomo maybe some of these changes can be ported to euth_wagtail?


slomo:
We decided in the scrum of scrums to move this widget to core. So instead of doing extra iterations here, just let move that guy to core.

xi:
In that case, my request is simply to consider these changes when creating the unified core version.

For meinberlin these changes are still valuable because they fix existing bugs.

2e2a:
Merging this. Will have to take a look at all changes we have done, when moving the widget to core.

slomo:
I was rather suggesting to move your updated version to core, because we didn't continue working on this.

PR #56, opened 2017-02-15
phase indicator and info

2e2a:


apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
```diff
@@ -0,0 +1,43 @@
...
+{% with project.active_phase as active_phase %}
+
+<div class="phaseList">
+    <ul>
+        {% with project.phases|length as phase_count %}
+        {% for phase in project.phases %}
+            {% includeTemplateString "meinberlin_projects/includes/phase_popover.html" phase=phase phase_num=forloop.counter phase_count=phase_count as popover_content %}
+            <li>
+                <button
+                    title="{% trans phase.name %}"
```
xi: Is this marked as translatable somewhere else? Otherwise this will not be extracted.

apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
```diff
@@ -0,0 +1,43 @@
...
+                    {{ phase.name }}
+                </button>
+            </li>
+        {% endfor %}
+        {% endwith %}
+    </ul>
+</div>
+
+<div class="phaseInfo">
+    <div class="l-center-8">
```
xi: should be `l-center-6`

meinberlin/assets/scss/components/_phases.scss
```diff
@@ -0,0 +1,32 @@
...
+    text-align: center;
+
+    ul {
+        margin: 0;
+        padding: 0;
+    }
+
+    li {
+        display: inline-block;
+    }
```
xi: These descendanct selectors are problematic because they also apply to all lists inside of the popup. We should refactor this later.

meinberlin/assets/scss/components/_phases.scss
```diff
@@ -0,0 +1,32 @@
...
+
+    button {
+        text-decoration: none;
```
xi: why is this needed?

xi:
Nice!

Next steps:

- hide popover on outside click (see the [solution in euth_wagtail](https://github.com/liqd/euth_wagtail/blob/master/euth/projects/static/euth_projects/timeline-popover.js))
- advanced styling (e.g. transition, shadow, ...)

PR #59, opened 2017-02-15
phase indicator fixup

2e2a:
Fixup for #56:
- small fixes
- use dedicated `popover.js` that is loaded only when needed
- close popover on a click somewhere else on page


xi:
LGTM

PR #61, opened 2017-02-15
implement same-height tiles

xi:
It just came to my mind that using flexbox here might actually work quite well, so I quickly implemented a frist version.

### DIFF
--- a/meinberlin/assets/scss/_layout.scss
+++ b/meinberlin/assets/scss/_layout.scss
@@ -23,4 +23,18 @@
     .l-tile-2 {
         @include grid-same-width(2);
     }
+
+    @supports (display: flex) {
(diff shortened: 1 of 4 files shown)


meinberlin/assets/scss/_layout.scss
```diff
@@ -23,4 +23,18 @@
...
     }
+
+    @supports (display: flex) {
+        .l-tile-wrapper {
+            display: flex;
+            flex-wrap: wrap;
+            justify-content: space-between;
+        }
+
+        // HACK to make this more specific than the grid-same-width selector
```
2e2a: Need to explain it to me, but will already merge.

2e2a:
lgtm

PR #62, opened 2017-02-16
Error indicator in project initiator tabs

2e2a:


xi:
Works fine. The UI could be more prominent, but works for now.

PR #69, opened 2017-02-20
simple project list

xi:
This adds a very simple project list.

Since the project-detail view is already in core, it may be a good idea to move the project-list view to core, too. But I guess we can first see what we have to add to it here and move it later if it still seems like a good idea.

### DIFF
--- a/apps/projects/urls.py
+++ b/apps/projects/urls.py
@@ -1,3 +1,4 @@
+from adhocracy4.projects.urls import urlpatterns as a4_projects_urls
 from django.conf.urls import url
 
 from . import views
@@ -6,3 +7,5 @@ urlpatterns = [
     url(r'^$', views.ProjectListView.as_view(),
         name='project-list'),
 ]
+
+urlpatterns += a4_projects_urls
--- a/apps/projects/urls.py
+++ b/apps/projects/urls.py
@@ -1,5 +1,6 @@
-from adhocracy4.projects.urls import urlpatterns as a4_projects_urls
 from django.conf.urls import url
+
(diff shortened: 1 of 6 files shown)


xi:
I just realized that there already was a project-list as part of the dashboard. We should either merge them or give them more specific names.

meinberlin/assets/scss/_layout.scss
```diff
@@ -36,5 +40,10 @@
             margin-left: 0;
             margin-right: 0;
         }
+
+        .l-tile-4.l-tile-4 {
```
vellip: this appears to be a bit hacky, i think?

xi: See comment above.

vellip:
@xi, @2e2a what do you think of [this](https://github.com/liqd/a4-meinberlin/pull/69/commits/bb122f64b19047e3c4a071b01824cd35e710d9c5) way to get project urls from a4 and our owns together?

2e2a:
> @xi, @2e2a what do you think of this way to get project urls from a4 and our owns together? At least for now, before we move it into a4 directly.

Makes sense for me.

2e2a:
But import order needs to be fixed.

apps/projects/urls.py
```diff
@@ -6,3 +7,5 @@
     url(r'^$', views.ProjectListView.as_view(),
         name='project-list'),
 ]
+
+urlpatterns += a4_projects_urls
```
xi: Definitely nicer than my version!

PR #70, opened 2017-02-21
update wagtail

vellip:


xi:
We do not use wagtail yet, so breaking changes should not affect us. Merging.

PR #73, opened 2017-02-22
add landingpage

vellip:


meinberlin/assets/scss/components/_hero_unit.scss
```diff
@@ -0,0 +1,40 @@
+.hero-unit {
+    position: relative;
+    min-height: 25em;
+
+    background: no-repeat center center;
```
xi: This implicitly overwrites all other background properties. I would prefer `background-size`/`background-position`

apps/cms/models.py
```diff
@@ -1,22 +1,68 @@
...
     content_panels = Page.content_panels + [
-        FieldPanel('body', classname="full")
+        edit_handlers.FieldPanel('subtitle'),
+        ImageChooserPanel('header_image'),
+        edit_handlers.StreamFieldPanel('body'),
     ]
 
+    @property
+    def featured_projects(self):
+        return Project.objects.order_by('-created')[:8]
```
xi: Why not use `adhocracy4.ProjectManager.featured()`?

apps/cms/templates/meinberlin_cms/home_page.html
```diff
@@ -1,20 +1,49 @@
...
 
 {% block header %}
     {% if self.search_description %}
-    <meta name="description" content="{{self.search_description}}" />
+        <meta name="description" content="{{ self.search_description }}"/>
     {% endif %}
 {% endblock %}
 
 {% block content %}
+    {% image self.header_image fill-1920x560-c100 as header_image %}
```
xi: Maybe it makes sense to use a [responsive image](https://responsiveimages.org/) here (to reduce download size on mobile).

xi: Forget what I said. I missed that this is used in CSS.

meinberlin/assets/scss/components/_blocks.scss
```diff
@@ -0,0 +1,12 @@
+.block--cta {
+    padding: 2.8em 0;
+    margin: 2.8em 0;
+
+    background-color: $brand-primary;
+    color: $text-color-inverted;
```
xi: Should be `contrast-color($brand-primary)`.

meinberlin/assets/scss/components/_hero_unit.scss
```diff
@@ -0,0 +1,40 @@
...
+
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    right: 0;
+}
+
+.hero-unit__title,
+.hero-unit__subtitle {
+    color: $text-color-inverted;
```
xi: I would prefer to set the color on `.hero-unit` along with the background.

meinberlin/assets/scss/components/_hero_unit.scss
```diff
@@ -0,0 +1,40 @@
...
+        background-image: linear-gradient(to bottom, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.8));
+    }
+
+    @media screen and (min-width: $breakpoint) {
+        min-height: 41em;
+    }
+}
+
+.hero-unit__wrapper {
+    width: 100%;
```
xi: This is implicit in `left: 0` and `right: 0`

meinberlin/assets/scss/components/_hero_unit.scss
```diff
@@ -0,0 +1,40 @@
+.hero-unit {
+    position: relative;
+    min-height: 25em;
+
+    background: no-repeat center center;
+    background-size: cover;
+
+    &:before {
+        content: ' ';
```
xi: In other places (e.g. `_tabs.scss`) we use `""` as a value here. I would prefer to have this consistent.

meinberlin/assets/scss/components/_blocks.scss
```diff
@@ -0,0 +1,12 @@
+.block--cta {
+    padding: 2.8em 0;
+    margin: 2.8em 0;
```
xi: It might be a good idea to use something like this instead:

```
.block {
    margin-bottom: 2.8em;
}
```

meinberlin/assets/scss/components/_hero_unit.scss
```diff
@@ -0,0 +1,40 @@
+.hero-unit {
+    position: relative;
+    min-height: 25em;
```
xi: I am not a big fan of magic numbers, but I guess in this specific context they are fine.

apps/cms/templates/meinberlin_cms/home_page.html
```diff
@@ -1,20 +1,49 @@
...
 
-        {{ self.body|richtext }}
+                {% has_perm 'a4projects.view_project' request.user project as show %}
+                {% if show %}
+                    {% include "a4projects/includes/project_list_tile.html" with project=project %}
+                {% endif %}
+
+            {% endfor %}
+        </div>
     </div>
```
xi: Can we convert this to a block?

apps/cms/templates/meinberlin_cms/home_page.html
```diff
@@ -1,20 +1,49 @@
...
+    <div class="hero-unit" style="background-image: url({{ header_image.url }})">
+        <div class="l-wrapper hero-unit__wrapper">
+            <h1 class="hero-unit__title">{{ self.title }}</h1>
+            <h2 class="hero-unit__subtitle">{{ self.subtitle }}</h2>
+        </div>
+    </div>
+
     <div class="l-wrapper">
-        <h1>{{ self.title }}</h1>
+        <h3>Aktuelle Beteiligungsverfahren</h3>
```
xi: Should be translatable.

apps/cms/models.py
```diff
@@ -1,22 +1,68 @@
...
+    )
+
+    class Meta:
+        template = 'meinberlin_cms/blocks/columns_block.html'
+
 
 class HomePage(Page):
-    body = RichTextField(blank=True)
+    body = fields.StreamField([
+        ('paragraph', blocks.RichTextBlock()),
```
xi: We should either restrict the size (e.g. `.l-center-6`) or not allow paragraph blocks at all.

apps/cms/templates/meinberlin_cms/home_page.html
vellip: You mean a wagtail block? Why would we want to do that? Do you mean for the editorial seöection of featured projects? I didn't do it yet because it was in another user story, but I talked with @MagdaN about it, it can be done. But i'd do that in a separate PR.

apps/cms/models.py
vellip: Why? I just thought it'd be a nice addition to also just be able add text. It shouldn't look broken or such anyway.

apps/cms/templates/meinberlin_cms/home_page.html
xi: I just thought that maybe editors want to change the order, e.g. have some text before the project list.

apps/cms/models.py
xi: I agree that it is a nice addition, but I doubt that it would work well in production without the `.l-center-6`. You can also just leave it and we get back to it if needed.

vellip: Do you mean typographically, because of the length of a line? Then I would understand your point and agree.

xi:
Looks very nice!

I added a lot of comments, but none of them are really blocking this from merging. Still you might want to implement some of my suggestions. Feel free to merge at any time!

PR #74, opened 2017-02-22
Create categories on dashboard

2e2a:
This currently uses a simple `FormSet` that allows to add only one categories at a time. This will be extended with a scrict making this dynamic.

Relies on: https://github.com/liqd/adhocracy4/pull/37


vellip:
working on ui

meinberlin/assets/scss/components/_formset.scss
```diff
@@ -24,7 +24,13 @@
...
 
 .formset__delete {
-    display: none;
+    @include sr-only();
+
+    &:focus ~ .formset__delete-label {
+        outline-width: 2px;
+        outline-style: solid;
+        outline-color: Highlight;
+    }
```
xi: AFAIK focus styles vary wildly across browsers. Also, the `Highlight` keyword is listed as deprecated on [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/color).

I guess you took this from some article (e.g. like [this one](https://ghinda.net/article/mimic-native-focus-css/)), but it should be clearly marked as a hack.

apps/contrib/templates/meinberlin_contrib/includes/inline_form_field.html
```diff
@@ -0,0 +1,13 @@
...
+        {% trans field.label %}
+    </label>
+    {% if field.help_text %}
+    <div class="form-hint">
+        {% trans field.help_text %}
+    </div>
+    {% endif %}
+    {{ field }}
+    {{ field.errors }}
+</div>
```
xi: This file is identical to `form_field.html`

xi: Also does not seem to be used. Maybe it was added by mistake?

apps/contrib/formset.py
```diff
@@ -0,0 +1,22 @@
+from django import forms
+
+
+class DynamicModelFormSet(forms.BaseModelFormSet):
+    def get_default_prefix(cls):
+        return 'test'
```
xi: This seems like it is unfinished.

apps/contrib/formset.py
```diff
@@ -0,0 +1,22 @@
...
+        return super().media + additional_js
+
+
+def dynamic_modelformset_factory(*args, **kwargs):
+    kwargs.update({
+        'formset': DynamicModelFormSet
+    })
+    modelformset = forms.modelformset_factory(*args, **kwargs)
+
+    return modelformset
```
xi: If I see correctly the whole purpose of this file is to inject `formst.js` and `get_default_prefix`, right? Is there no simpler way to do that?

apps/dashboard/forms.py
```diff
@@ -115,6 +140,18 @@ def __init__(self, blueprint, organisation, creator, *args, **kwargs):
                 get_module_settings_form(module_settings),
             ))
 
+        self.show_categories_form = \
+            self._show_categories_form(self.blueprint.content)
+        if self.show_categories_form:
+            kwargs['categories__queryset'] = \
+                category_models.Category.objects.none()
```
xi: Why is this done? A comment might be helpful here.

apps/contrib/formset.py
vellip: You could just add the js file inside the template, but the thought behind this was to make it more reusable and especially easier to use, because it takes care care of implementing the JS.

apps/contrib/templates/meinberlin_contrib/includes/inline_form_field.html
vellip: Good point! That was from early in the development.

meinberlin/assets/scss/components/_formset.scss
```diff
@@ -0,0 +1,54 @@
...
+        position: absolute;
+        right: 0.3em;
+        bottom: 0.3em;
+        color: $gray;
+    }
+
+    &:checked ~ * {
+        opacity: 0.3;
+        pointer-events: none;
+    }
```
xi: IMHO this looks very similar to an empty input with a placeholder, which is confusing. I would prefer a better distinction.

meinberlin/assets/scss/components/_formset.scss
```diff
@@ -0,0 +1,54 @@
...
+        background-color: $body-bg;
+    }
+}
+
+.formset--inline .formset__form--has-error {
+    border-bottom-color: $brand-danger;
+    border-bottom-width: 5px;
+}
+
+.formset__empty {
```
xi: Maybe rename this to `formset__template`? I took some time to understand its purpose.

apps/contrib/formset.py
xi: Why not just including the js in our bundle?

meinberlin/assets/scss/components/_formset.scss
vellip: That's the way it's named by Django, so I thought it would be clearer to have the same name

apps/contrib/formset.py
vellip: I didn't want to load it on every page when it's not necessary.

xi:
Code looks good (see my comments) and works fine. However, I am not completely convinced about the UI:

- categories marked for deletion look very similar to empty ones.
- It would be relatively simple to "un-mark" a category marked for deletion. The current UI hides the respective button though.
- It is currently possible to add a large number of empty subforms without a way to remove them again (except reloading, thus losing all other edits)
- There is no warning on duplicated category names.

apps/contrib/templates/meinberlin_contrib/includes/inline_form.html
```diff
@@ -0,0 +1,23 @@
+{% load i18n %}
+<div class="formset__form{% if form.name.errors %} formset__form--has-error{% endif %}"
+     data-container="body"
+     data-toggle="popover"
+     data-trigger="hover"
+     data-placement="top"
+     data-content="{% for error in form.name.errors %}{{ error }}{% endfor %}">
```
xi: When are these used? I could not find a way to trigger an error. (subforms with an empty name seem to be ignored.)

meinberlin/assets/scss/components/_formset.scss
xi: Yes, we use django's `empty_form` here, but we do use it as a template.

apps/contrib/templates/meinberlin_contrib/includes/inline_form.html
vellip: When you have a category that's already saved and you remove the text in the input for that respective category you get an error.

apps/contrib/formset.py
xi: `formset.js` is only few lines of code. The bigger issues is `popover.js`.

apps/contrib/templates/meinberlin_contrib/includes/inline_form.html
```diff
@@ -0,0 +1,23 @@
...
+    {% if can_delete %}
+        <input type="checkbox" name="{{ form.DELETE.html_name }}" id="{{ form.DELETE.id_for_label }}" class="formset__delete">
```
xi: Why is this not next to the label (visual order = tab order)?

apps/contrib/templates/meinberlin_contrib/includes/inline_form.html
xi: Ok, I tested that now. There seem to be some issues:

- Not accessible via keyboard
- The error is not counted in the tab

xi: The second issue is fixed in 6c4304a.

I think we can live with the keyboard issue for now.

PR #75, opened 2017-02-23
proper project tiles

vellip:
screen shot 2017-02-23 at 09 44 06

right now with placeholder images, obviously they should be replaced when we have project images.

### DIFF
--- a/apps/projects/templates/a4projects/includes/project_list_tile.html
+++ b/apps/projects/templates/a4projects/includes/project_list_tile.html
@@ -1,15 +1,36 @@
-{% load i18n project_tags thumbnail %}
+{% load i18n project_tags thumbnail static %}
 
 <div class="tile">
-    {% get_days project.days_left as days %}
-    {% if project.has_finished %}
-    <span class="label">{% trans "Finished" %}</span>
-    {% elif days %}
-    <span class="label">{{ days }}</span>
-    {% endif %}
+    <div class="tile__head">
+        <a href="{% url 'project-detail' project.slug %}">
+            <img src="http://placehold.it/285x170" alt="{{ project.name }}">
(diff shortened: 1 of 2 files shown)


apps/projects/templates/a4projects/includes/project_list_tile.html
```diff
@@ -1,15 +1,36 @@
...
 <div class="tile">
-    {% get_days project.days_left as days %}
-    {% if project.has_finished %}
-    <span class="label">{% trans "Finished" %}</span>
-    {% elif days %}
-    <span class="label">{{ days }}</span>
-    {% endif %}
+    <div class="tile__head">
+        <a href="{% url 'project-detail' project.slug %}">
+            <img src="http://placehold.it/285x170" alt="{{ project.name }}">
```
xi: The `alt` is redundant because `project.name` is also available from `list-item__title`

meinberlin/assets/scss/components/_tile.scss
```diff
@@ -1,5 +1,48 @@
...
+
+.tile__hint {
+    margin-bottom: 0;
+
+    color: $brand-danger;
+    font-size: $font-size-sm;
+}
+
+.tile__head {
+    position: relative;
```
xi: maybe add `min-height: 6em` here for safety

meinberlin/assets/scss/components/_tile.scss
```diff
@@ -1,5 +1,48 @@
...
+    height: 3.4em;
+    width: 3.4em;
+
+    position: absolute;
+    bottom: 0.8em;
+    right: 0.8em;
+}
+
+.tile__body {
+    flex-grow: 1;
```
xi: The [spec](https://www.w3.org/TR/css-flexbox-1/#flex-components) recommends to use the `flex` shorthand. I am not sure whether we should follow that advice.

PR #77, opened 2017-02-23
Show category of an idea

2e2a:


xi:
Works fine. I wonder whether we should use consistent styling in both cases (i.e. a `.category` selector).

PR #78, opened 2017-02-23
style platform header

vellip:
screen shot 2017-02-23 at 15 15 44


meinberlin/templates/header.html
```diff
@@ -0,0 +1,30 @@
...
+        </nav>
+        <div class="main-header__user">
+            {% userindicator %}
+        </div>
+    </div>
+    <div class="dialogue-box">
+        <div class="dialogue-box__left"></div>
+        <div class="dialogue-box__center"></div>
+        <div class="dialogue-box__right"></div>
+    </div>
```
xi: I would like to have a comment here explaining what it does

meinberlin/assets/scss/components/_header.scss
```diff
@@ -43,12 +47,34 @@
...
+.dialogue-box {
+    display: flex;
+    margin-top: -10px;
+
+    > div {
+        height: 43px;
+        background-image: url('/static/images/dialoguebox.png');
+        background-repeat: no-repeat;
+    }
+}
```
xi: It would be more robust to use a separate image for left/right that just contains the line and repeats along the x-axis.

meinberlin/assets/scss/components/_header.scss
```diff
@@ -1,15 +1,19 @@
...
-    top: 0;
+    bottom: 0;
     right: 0;
 }
 
+.main-header__toggle {
+    padding: 0.5em 1.5em;
+    border: solid 1px $border-color;
+    border-radius: 50px;
+}
```
xi: This should be a midifier on dropdown-toggle.

meinberlin/assets/scss/components/_header.scss
```diff
@@ -43,12 +47,34 @@
 
 .main-nav__logo {
     img {
-        width: 7em;
-        margin-top: -3em;
-        margin-bottom: -3em;
+        width: 6em;
+        margin-top: -2em;
+        margin-bottom: -2em;
```
xi: The idea here was to vertically center the logo: 1em (line-height) + 3em + 3em = 7em. So if the total height is reduced to 6em, the magins should be 2.5em.

xi:
Nice work! Again, it may seem like a lot of comments, but most of them are not critical.

- Somehow the vertical alignment between logo, links and user indicator is off
- as far as I remember, there should be a margin below the header to highlight the blue line.
- the "be berlin" logo is missing
- the user indicator is too far right (should have `$padding`)

vellip:
The be berlin logo is missing right now because @damusp hasn't come up with an idea to display it properly yet.
Since the blue line would not hide anymore with a blue header image (since we don't have the blue coloured image on the landing page anymore), I thought it was desirable to keep it that way. Maybe @damusp can enlighten us.

damusp:
I've talked with Jana and Swetlana about this. We can add the beBerlin logo like it is shown in the design (https://zpl.io/ZwlnbY), but it will just not show up on certain pages (e.g. dashboard)

vellip:
What about pages where there is no image in the header? A white logo on a white background might not work that well...

damusp:
I'd prefer to have the logo show up inverted in these cases. What do you think?

alternatively we could try to make sure that the background is always dark enough (colored backgrounds as placeholder for images)

vellip:
there will always be pages where it won't work just because there is no image at all in the header. In general I think the better way is to have one solution (image) for all.

xi:
> there will always be pages where it won't work just because there is no image at all in the header. In general I think the better way is to have one solution (image) for all.

I strongly agree.

We should go ahead with this change and add the missing logo later.

PR #81, opened 2017-02-28
change polylint.rc 

vellip:
fix polylint.rc to support commits without activated virtual env


xi:
Note that this expects the virtual env in the default location `.env`.

PR #83, opened 2017-02-28
footer

xi:
This adds a basic footer, including a social sharing widget.

Note that it is currently not guaranteed that the site fills the screen, i.e. there could be whitespace below the footer.

### DIFF
--- a/meinberlin/assets/js/app.js
+++ b/meinberlin/assets/js/app.js
@@ -6,7 +6,6 @@ window.Tether = require('tether/dist/js/tether.js')
 
 // social share
 var Shariff = require('shariff')
-var shariffCSS = require('shariff/build/shariff.min.css')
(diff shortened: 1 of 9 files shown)


meinberlin/assets/js/app.js
```diff
@@ -1,9 +1,16 @@
-/* eslint no-unused-vars: "off" */
+/* eslint no-unused-vars: "off", no-new: "off" */
 
 // make jquery available for non-webpack js
 var $ = window.jQuery = window.$ = require('jquery')
 window.Tether = require('tether/dist/js/tether.js')
 
+// social share
+var Shariff = require('shariff')
+var shariffCSS = require('shariff/build/shariff.min.css')
```
vellip: This should be included in the webpack `vendor` entry point

PR #86, opened 2017-03-06
Filter ideas by category

2e2a:
Requires: https://github.com/liqd/adhocracy4/pull/39

Open issues:

- The filtering/sorting parameters  do not work with pagination and will be fixed separately
- form needs to be styled
- Do not display empty filter (e.g. no categories)

### DIFF
--- a/apps/contrib/templates/meinberlin_contrib/includes/filter_and_sort.html
+++ b/apps/contrib/templates/meinberlin_contrib/includes/filter_and_sort.html
@@ -1,0 +1,5 @@
+{% load i18n %}
+<form action="" method="get">
+    {{ filter.form }}
+    <input type="submit" />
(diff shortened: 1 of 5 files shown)


xi:
LGTM.

xi:
I experimented a bit with the UI: 1cd36ba. This commit is not really ready for production, but it might be a useful reference.

xi:
I noticed that django-filter does have a `ModelMultipleChoiceFilter` class, so ORing several categories should not be that hard to implement.

2e2a:
> I noticed that django-filter does have a ModelMultipleChoiceFilter class, so ORing several categories should not be that hard to implement.

Thats good to know. Will keep it in mind, if we need to extend the filter.

PR #90, opened 2017-03-07
Filter UI

xi:
Building on #86, this changes UI to be more in line with the previous sorting UI. There is still room for improvement:

-   style the dropdown buttons (similar to select, withy max-width and text-overflow)
-   improve UI in cases where there are many filters (especially on mobile)

### DIFF
--- a/apps/contrib/templates/meinberlin_contrib/includes/filter_and_sort.html
+++ b/apps/contrib/templates/meinberlin_contrib/includes/filter_and_sort.html
@@ -1,5 +1,15 @@
 {% load i18n %}
-<form action="" method="get">
-    {{ filter.form }}
-    <input type="submit" />
-</form>
+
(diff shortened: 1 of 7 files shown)


2e2a:
If no ordering is selected 'All' is shown instead of the first choice.

2e2a:
The filter and ordering elements do not have the same vertical position.
We could fix this later, but I think having this fixed would make the initial version look already much better.

xi:
> The filter and ordering elements do not have the same vertical position.

I thought I took care of that. Are you sure that you did run webpack?

2e2a:
>> The filter and ordering elements do not have the same vertical position.

> I thought I took care of that. Are you sure that you did run webpack?

Apparently not ... Ignore this comment.

2e2a:
lgtm

PR #91, opened 2017-03-07
Fix rating in list item

xi:
there have been some changes to ratings, so the list items needed to be adapted.


2e2a:
This adds a larger space between the rating icons, which is also larger than the space to the comments icon.

xi:
I fixed some inline whitespace issues. The problem you describe still exists, but is now less grave.

IMHO this is ok for now. This is simply how our rating/list item components look at the moment. It might be a good idea to refactor them later though.

PR #94, opened 2017-03-07
Use button styles for filters

xi:
This adds a new button modifier `.button--form` which looks like something between a button and an input (specifically, `select`).

This new kind of button is used for the filter dropdowns.

### DIFF
--- a/apps/contrib/templates/meinberlin_contrib/widgets/dropdown_link.html
+++ b/apps/contrib/templates/meinberlin_contrib/widgets/dropdown_link.html
@@ -1,11 +1,12 @@
 <div class="dropdown {% if right %}dropdown--right{% endif %}" {{ attrs }}>
     <button
         type="button"
-        class="dropdown-toggle"
+        class="dropdown-toggle button button--select"
         data-toggle="dropdown"
         aria-haspopup="true"
         aria-expanded="false"
         id="{{ id }}"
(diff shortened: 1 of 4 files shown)


vellip:
I find the name a bit confusing. It's not really saying what it is/does - or rather: it needs an explanation - I feel like when you have to explain names the name is bad. I'd prefer something like `.button--dropdown` or `.button--select` or `

apps/contrib/templates/meinberlin_contrib/widgets/dropdown_link.html
```diff
@@ -1,11 +1,12 @@
 <div class="dropdown {% if right %}dropdown--right{% endif %}" {{ attrs }}>
     <button
         type="button"
-        class="dropdown-toggle"
+        class="dropdown-toggle button button--select"
         data-toggle="dropdown"
         aria-haspopup="true"
         aria-expanded="false"
         id="{{ id }}"
+        title="{{ label }}: {{ value_label }}"
```
vellip: Why'd you add the title here? The button has the same value as text anyway two lines further down.

xi: Yes, but it may be truncated. I found that I personally expect a tooltip on truncated buttons.

PR #95, opened 2017-03-08
Use mixin to add module to request for category filter

2e2a:


apps/ideas/views.py
```diff
@@ -57,20 +57,19 @@ class Meta:
...
     def dispatch(self, *args, **kwargs):
-        # TODO: Refactor duplicate to ProjectMixin.dispatch()
-        self.project = kwargs['project']
-        self.phase = self.project.active_phase or self.project.past_phases[0]
-        self.module = self.phase.module if self.phase else None
         self.request.module = self.module
         return super().dispatch(*args, **kwargs)
 
 
+class IdeaListView(mixins.ProjectMixin, RequestMixin, FilteredListView):
```
xi: I do not like that you have to know the exact order of these three superclasses. Can we maybe combine `ProjectMixin` and `RequestMixin` into one?

2e2a: This did not work. Added an explanation comment for now.

PR #96, opened 2017-03-08
style category

xi:
see https://github.com/liqd/a4-meinberlin/pull/77#issuecomment-283002124


vellip:
The position is not just yet as it is in the design, but I think, we can do that later.

meinberlin/assets/scss/_variables.scss
```diff
@@ -9,6 +9,7 @@ $brand-danger: #dc3918;
 
 $body-bg: #fff;
 $bg-secondary: #f6f6f6;
+$bg-tertiary: #e6e6e6;
```
xi: We should not add new color variables all the time. For me it is not clear when to use which.

PR #98, opened 2017-03-08
responsive filter UI

xi:
this uses a grad based two-column layout for small screens and a simple inline layout on large screens (if wrapping occurs in this mode, it may look at bit broken but is still usable).

The code is a bit hacky but not too much (I hope).

### DIFF
--- a/meinberlin/assets/scss/components/_filter_bar.scss
+++ b/meinberlin/assets/scss/components/_filter_bar.scss
@@ -5,12 +5,31 @@
     .dropdown {
         display: inline-block;
     }
-}
 
-.filter-bar__filters {
-    float: left;
-}
+    @media (min-width: $breakpoint) {
+        .filter-bar__ordering {
+            float: right;
+        }
+    }
 
-.filter-bar__ordering {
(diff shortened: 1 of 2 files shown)


meinberlin/assets/scss/components/_filter_bar.scss
```diff
@@ -5,12 +5,31 @@
...
 .filter-bar__ordering {
-    float: right;
+    @media (min-width: $breakpoint) {
+        float: right;
+    }
+
+    @media (max-width: $breakpoint) {
+        // push to the right
+        // NOTE: needs to overwrite grid-tiles
+        @include grid-position(6);
```
xi: I think this needs more specificity (see comment in line 32).

vellip: It worked just fine when I tested it.

xi: On the [development server](https://meinberlin-dev.liqd.net/projects/test/) this does not work: The ordering button should be on the right, but it is left aligned.

PR #101, opened 2017-03-09
make header responsive

vellip:


2e2a:
Works nice. I would wait for @xi to take a look at the css.

meinberlin/templates/header.html
```diff
@@ -7,21 +7,26 @@
...
-            <ul>
-                <li>
-                    <a class="main-nav__item" href="{% url 'project-list' %}">{% trans 'Participation processes' %}</a>
-                </li>
-                {% get_menu "topnav" as topnav %}
-                {% for item in topnav %}
-                    <li>
-                        <a class="main-nav__item" href="{{ item.url }}">{{ item.title }}</a>
+            <button class="button button--bordered main-nav__toggle" data-toggle="collapse" data-target="#menu-collapse" aria-expanded="false" aria-controls="menu-collapse">
+                <i class="fa fa-bars"></i>
```
xi: icon should be hidden and there should be a text alternative (e.g. `title` or `aria-label`)

xi:
LGTM. There are few issues that can be fixed separately:

- The meinberlin logo should be bigger, probably around `5em`.
- This adds yet another button variant, `.button--bordered`. We could merge it with `.button--light`.

PR #105, opened 2017-03-13
Project type filter

2e2a:


xi:
Works fine. It might  be nice to also show the project type somewhere in the future.

PR #107, opened 2017-03-14
initial embedding

xi:
This adds the basic structure for embedding. It is mostly based on 2017-01-tb-embed.

Not that the JavaScript is only a first version demonstrating the general mechanism. Many edge-cases are not yet handled (see FIXMEs).

### DIFF
--- a/apps/embed/templates/meinberlin_embed/embed.html
+++ b/apps/embed/templates/meinberlin_embed/embed.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+{% load static %}
+<html>
+<head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <title>Embedded meinBerlin</title>
+    <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico" />
+    <link rel="stylesheet" type="text/css" href="{% static 'vendor.css' %}" />
+    <link rel="stylesheet" type="text/css" href="{% static 'meinberlin.css' %}" />
+    <script src="{% url 'javascript-catalog' %}"></script>
+    <script src="{% static 'vendor.js' %}"></script>
+    <script src="{% static 'meinberlin.js' %}"></script>
+    <meta name="viewport" content="width=device-width" />
+    {% block header %}{% endblock %}
+</head>
+
+<body>
+    embed
+</body>
+</html>
--- a/apps/embed/templates/meinberlin_embed/embed.html
+++ b/apps/embed/templates/meinberlin_embed/embed.html
@@ -21,9 +21,7 @@
     <header class="embed-header">
         <a>Login</a>
     </header>
-    <main class="embed-main">
-        Embed
-    </main>
+    <main class="embed-main"></main>
     <footer class="embed-footer">
         <a>Zur Plattform</a>
     </footer>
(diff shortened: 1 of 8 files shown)


meinberlin/static/embed.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  var onReady = function () {
+    // adhocracy4.onReady($target)
+
+    $main.find('form[action]').submit(function (event) {
+      event.preventDefault()
+      var form = this
+
+      // timeout required for use with CKEditor
```
vellip: What exactly is the problem with the CKEditor?

meinberlin/static/embed.js
```diff
@@ -0,0 +1,50 @@
...
+    var $root = $(html).filter('main')
+    $main.empty()
+    $main.append($root.children())
+    onReady()
+  }
+
+  var onReady = function () {
+    // adhocracy4.onReady($target)
+
+    $main.find('form[action]').submit(function (event) {
```
vellip: I think a delegated event is more suitable here.

meinberlin/static/embed.js
xi: It seems to do bind to the submit event, too. So we have to ensure to get the form data *after* ckeditor is done.

meinberlin/static/embed.js
xi: I thought that submit did not bubble. Thanks for the hint!

apps/embed/templates/meinberlin_embed/embed.html
```diff
@@ -0,0 +1,31 @@
...
+    <meta name="viewport" content="width=device-width" />
+    {% block header %}{% endblock %}
+</head>
+
+<body>
+    <header class="embed-header">
+        <a>Login</a>
+    </header>
+    <main class="embed-main">
+        Embed
```
vellip: I'd prefer to have this textnode removed, it flashes otherwise when the content got loaded.

apps/embed/urls.py
```diff
@@ -0,0 +1,8 @@
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    url(r'^$',
+        views.EmbedView.as_view(), name='idea-embed'),
```
vellip: I don't think we should call it idea-embed. For simplicity we can just call it embed and change afterwards.

meinberlin/static/embed.js
```diff
@@ -0,0 +1,50 @@
...
+      })
+    }
+  })
+
+  $(document).on('submit', 'form[action]', function (event) {
+    event.preventDefault()
+    var form = event.target
+
+    // timeout required for use with CKEditor
+    setTimeout(function () {
```
vellip: In my test this wasn't necessary [anymore?], maybe you can check that out again?

meinberlin/static/embed.js
```diff
@@ -28,7 +28,7 @@ $(document).ready(function () {
     }
   })
 
-  $(document).on('submit', function (event) {
+  $(document).on('submit', 'form[action]', function (event) {
```
xi: Out of interest (because I forgot): What are the cases where a submit event is fired from something that does not match `form[action]`? Why do we not want to handle these cases?

vellip: There was a problem with submitting a comment. After submitting it, the embed body was empty because the form had no action attribute. Most likely because it's done with React.

meinberlin/static/embed.js
xi: I guess this now works because of the delegation.

PR #108, opened 2017-03-15
Do not show rating elements if feature is not enabled

2e2a:
Depends on https://github.com/liqd/adhocracy4/pull/46.

The brainstorming blueprint does not allow rating in its phases. In this case the rating element are not shown.

### DIFF
--- a/apps/ideas/templates/meinberlin_ideas/includes/idea_list_item.html
+++ b/apps/ideas/templates/meinberlin_ideas/includes/idea_list_item.html
@@ -3,8 +3,7 @@
 <div class="list-item list-item--squashed">
     <div class="list-item__stats">
         {% spaceless %}
-        {% itemHasFeature idea "rate" as is_rating_enabled %}
-        {%  if is_rating_enabled %}
+        {% if idea|has_feature:"rate" %}}
             <span class="rating">
                 <span class="rating-button rating-up is-read-only" title="Positive Ratings">
                     {{ idea.positive_rating_count }}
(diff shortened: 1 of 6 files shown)


apps/ideas/views.py
```diff
@@ -24,6 +26,16 @@ class OrderingWidget(DropdownLinkWidget):
     right = True
 
 
+class IdeaOrdering(ChoicesRequestMixin, django_filters.OrderingFilter):
+
+    def get_choices(self, request):
+        choices = (('-created', _('Most recent')),)
+        if self.parent.request.module.has_feature('rate', idea_models.Idea):
```
vellip: What exactly is `self.parent` here?

apps/ideas/templates/meinberlin_ideas/includes/idea_list_item.html
```diff
@@ -1,16 +1,20 @@
+{% load module_tags %}
+
 <div class="list-item list-item--squashed">
     <div class="list-item__stats">
         {% spaceless %}
-        <span class="rating">
-            <span class="rating-button rating-up is-read-only" title="Positive Ratings">
-                {{ idea.positive_rating_count }}
-                <i class="fa fa-chevron-up" aria-hidden="true"></i>
+        {% if idea|has_feature:"rate" %}}
```
vellip: There's a brace too much at the end of the line.

apps/ideas/views.py
2e2a: parent is the `FilterSet`

PR #111, opened 2017-03-21
different embed navigation behaviors

xi:
This improves several aspects of navigating in the embed shell:

- ignore jump links (`href="#…"`, e.g. tabs)
- ignore query links (`href="?…"`, e.g. filter/ordering)
- ignore links in user generated content (requires the use of the richtext filter)
- ignore links/forms marked with `data-embed-target="platform"` (or "external")

If a link/form is not ignored, the navigation will happen inside of the iframe.


apps/contrib/middleware.py
```diff
@@ -0,0 +1,11 @@
...
+    """Append request path as a header.
+
+    In an ajax request, redirects are handled implicitly, so it it not possible
+    to know the path of the page where you end up. This middleware adds that
+    information in a header.
+    """
+
+    def process_response(self, request, response):
+        response['x-ajax-path'] = request.path
+        return response
```
vellip: Since we already have an `embed` app, I guess we should also define the middleware in that app.

meinberlin/static/embed.js
```diff
@@ -13,31 +15,48 @@ $(document).ready(function () {
...
+    var $link = $(event.target)
+    var embedTarget = $link.data('embedTarget')
+
+    if (
+      url &&
+      url[0] !== '#' &&
+      !event.target.target &&
+      embedTarget !== 'platform' &&
+      embedTarget !== 'external' &&
+      !$link.is('.rich-text *')
```
vellip: I am not really happy with that. Maybe we should make a function out of that that returns a boolean. That would in the end just move the problem, but at least it looks better in this part of the code?

vellip:
lookin' good otherwise!

apps/embed/templates/meinberlin_embed/embed.html
```diff
@@ -14,6 +14,7 @@
     <script src="{% static 'meinberlin.js' %}"></script>
     <script src="{% static 'embed.js' %}"></script>
     <meta name="viewport" content="width=device-width" />
+    <base target="_top" />
```
vellip: I'm still not sure this is a good idea? Not only does it navigate away from the page containing the iframe (which I'd find annoying as a user) but navigating away from that page also resets the state of the iframe should one want to go back, removing any input (i.e. a comment) the user might've prepared but not sent yet.

PR #113, opened 2017-03-22
Document embedding

xi:
This adds documentation for embedding. It describes the current approach and its impact on the overal application as well as the background knowledge we aquired through time. Feedback is very welcome, especially from @slomo and @MagdaN.

### DIFF
--- a/docs/embed.md
+++ b/docs/embed.md
@@ -1,0 +1,196 @@
+# Embedding
+
+Embedding allows to include parts of adhocracy on other websites. A common use
+case would be the website of a municipal administration that embeds an idea
+collection.  This allows users to access (and eventually contribute to) the
+participation right where they are instead of going to a third party platform.
+
+## TL;DR
+
+You can embed any project with the following code:
+
+    <iframe src="http://example.com/embed/projects/{slug}/" />
+
+## Terms
+
+-   iframe: The HTML element that enables embedding. It is basically a virtual
+    screen that is completely isolated from the embedding website for security
+    reasons.
+-   platform: The regular, non-embedded version of adhocracy. Some
+    functionality is only available on the platform.
+
+## Inspiration and Approaches
+
+While working on this concept we looked at several existing embedding systems
+and found three distinct categories:
+
+-   *Widgets* with limited interaction (e.g. [twitter timeline](https://dev.twitter.com/web/embedded-timelines))
+-   *Applications* (e.g. [soundcloud player](https://help.soundcloud.com/hc/en-us/articles/115003449627-The-HTML5-embedded-player))
+-   Functionality that *integrates* into the website (e.g. [disqus](https://disqus.com/))
+
+## Our chosen approach
+
+It was hard to decide between application and integration, but we finally
+landed on application.
+
+We embed fully functional projects. Anything outside of a project (e.g.
+managing projects, changing account settings) happens on the platform.
+
+In the iframe, we load a JavaScript application called the *embed shell* that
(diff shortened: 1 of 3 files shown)


docs/embed.md
```diff
@@ -0,0 +1,196 @@
...
+
+-   Add `data-embed-target="external"` to all links and forms that should not
+    open inside of the iframe.
+-   Do not trigger navigation from JavaScript
+-   Do not add submit handlers to forms that do not have
+    `data-embed-target="ignore"`.
+-   Do not use relative-path URLs:
+
+        // good
+        http://example.com/foo/bar
```
slomo: Since we are talking about internal links here, I would never consider this a good practice. If it needs to be with hostname we should also advocate for https or protocol relative links.

docs/embed.md
```diff
@@ -0,0 +1,196 @@
...
+
+The embed shell uses the regular views. This way, we do not need to duplicate
+all views. However, this also means that you need to be aware of some general
+restrictions:
+
+-   Add `data-embed-target="external"` to all links and forms that should not
+    open inside of the iframe.
+-   Do not trigger navigation from JavaScript
+-   Do not add submit handlers to forms that do not have
+    `data-embed-target="ignore"`.
```
slomo: Couldn't that be solved by stopping event propagation in those handlers? Or is that not reliable enough?

slomo:
Looks great so far. Just some random thoughts:

I think the `EmbeddedView` dispatch method probably needs to have `@xframe_options_exempt`, otherwise the `django.middleware.clickjacking.XFrameOptionsMiddleware` should prohibit embedding. Or am I missing something.

A super neat (NTH) feature of the `embed` app would be second view, that contains the embed code ready to copy and maybe also an example how it would look embedded in our favourite size. 

As far as I know the best practice naming convention for django apps is to use plurals of the base object (eg. `ideas`, `documents`). So in that case it should probably be `embeds`? Apply also to other apps in this repo.

docs/embed.md
MagdaN: I agree

docs/embed.md
```diff
@@ -0,0 +1,196 @@
...
+The whole point of embedding is lowering the entry barrier for new users. This
+works even better if the account system is integrated with the embedding
+website so that no additional login is required for participation.
+
+## Possible extensions
+
+-   Currently, it is only possible to embed full-functioning projects. Other
+    options may be added under the `/embed/` prefix, optionally using the embed
+    shell.
+-   The embed shell sends an additional `X-Embed` header with every request.
```
MagdaN: We agreed on implementing this, why is it a possible extension now?

MagdaN:
LGTM

docs/embed.md
xi: The header is implemented. There are simply no views yet that react to it. Also note that we are still working on this story and more features will be implemented during this sprint.

docs/embed.md
xi: Sometimes you do not want to stop event propagation. Or the event listener uses event delegation itself, in which case it is not clear which one will be executed first.

In my first draft the formulation was more like "be careful when …". I changed it to be more specific. This does not mean that you must not ever do this, it is just a recommendation to avoid trouble.

vellip:
Fine with me.

2e2a:
> I think the EmbeddedView dispatch method probably needs to have @xframe_options_exempt, otherwise the django.middleware.clickjacking.XFrameOptionsMiddleware should prohibit embedding. Or am I missing something.

Added the decorator to the embed view.

PR #114, opened 2017-03-23
Wagtail formbuilder for CSV email forms

2e2a:
Depends on https://github.com/liqd/adhocracy4/pull/48

The base email template is just a copy from a4-opin and might need adjustment.


apps/cms/emails.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def get_receivers(self):
+        return [x.strip() for x in self.object.to_address.split(',')]
+
+
+class CsvFormEmail(FormEmail):
+
+    def _generate_csv(self):
+        stream = io.StringIO()
+        fw = csv.writer(stream, delimiter=';', quoting=csv.QUOTE_MINIMAL)
```
xi: Why did you overwrite the defaults here? Is this more compatible?

apps/cms/emails.py
```diff
@@ -0,0 +1,39 @@
...
+        fw.writerow([_('Form Field'), _('Response')])
+        for field, value in self.object.field_values.items():
+            fw.writerow([field, value])
```
xi: Why did you not use `csv.DictWriter`?

apps/cms/models.py
```diff
@@ -116,3 +120,57 @@ def __str__(self):
...
+        help_text='Email content message',
+    )
+    attach_as = models.CharField(
+        max_length=3,
+        choices=(
+            ('csv', 'CSV Document'),
+        ),
+        default='csv',
+        help_text='Form results are send in this document format',
+    )
```
xi: It should not be hard to add this once we have more than one option, so why add this now?

apps/cms/models.py
```diff
@@ -116,3 +120,57 @@ def __str__(self):
...
+
+    @property
+    def field_values(self):
+        fields = {}
+        for field in self.form:
+            value = field.value()
+            if isinstance(value, list):
+                value = ', '.join(value)
+            fields[field.label] = value
+        return fields
```
xi: Why did you make this a property?

.polylintrc
```diff
@@ -1,5 +1,5 @@
 py:flake8 --exclude */migrations/*,*/settings/* %s
 py:isort -sl -rc -c %s
 scss:node_modules/.bin/stylelint --syntax scss %s
-html:node_modules/.bin/htmlhint -c .htmlhintrc %s
+html:node_modules/.bin/htmlhint -c .htmlhintrc -i **/email*.html %s
```
xi: As far as I can see, the only issue is the style tag in `email_base.html`. I would prefer not to exclude all email templates for that. Maybe just exclude `email_base.html`?

xi:
Looks good and works fine.

apps/cms/emails.py
2e2a: We need to change the delimiter, as its comma the default.
The default for quoting is already MINIMAL, so we might remove this.

apps/cms/emails.py
2e2a: I found this simple and readable enough, as we have only two columns.

apps/cms/models.py
2e2a: I had initially an additional option to put the responses in the mail, but removed it as it would require extending the email template system.

I still decided to keep this option to have the CSV generation more explicit.

apps/cms/models.py
2e2a: To spare the brackets when calling it.

.polylintrc
2e2a: Will change it.

PR #117, opened 2017-03-28
embed login

vellip:


apps/embed/templates/meinberlin_embed/embed.html
```diff
@@ -8,19 +8,32 @@
...
     <script src="{% static 'embed.js' %}"></script>
+    <script type="text/template" id="embed-confirm">
+    <div class="embed-confirm">
+        <p>{% trans 'You need to login first. By clicking "Login" a popup will open so you can sign in safely. Do you want to login now?' %}</p>
+        <div class="embed-confirm__actions">
+            <button class="embed-confirm__cancel js-embed-cancel button button--light">{% trans 'Cancel' %}</button>
+            <a href="{% url 'account_login' %}" target="_blank" class="embed-confirm__login js-embed-login button button--primary">{% trans 'Login' %}</a>
+        </div>
+    </div>
+    </script>
```
xi: I first missed the script tags because the div is not indented.

meinberlin/templates/base.html
```diff
@@ -13,6 +13,16 @@
...
     <script src="{% static 'meinberlin.js' %}"></script>
     <meta name="viewport" content="width=device-width" />
+    {% if request.user.is_authenticated %}
+    <script>
+        if (window.opener) {
+            window.opener.postMessage(
+                JSON.stringify({name: "popup-close",}),
+                location.origin
+            )
+        }
```
xi: This behavior should be document as a general restriction: Opening meinberlin in a popup for anything other than login is not supported.

xi:
The login in the header works very well. However, the embed-confirm dialog still has some issues. I can still navigate while it is shown which is strange and after logging in I end up on the initial view instead of the requested one.

I also think that `PlatformPopup` is a bit too complicated and I think I found some memory leaks. I will merge this now to avoid conflicts and will do some refactoring in a separate branch.

Issue #82, opened 2017-02-28
meinberlin header visually broken on alerts


xi:
The meinberlin header (#78) is supposed to party cover the hero image on the frontpage. this is broken if there are alerts:

![screenshot from 2017-02-28 15 25 59](https://cloud.githubusercontent.com/assets/202576/23409042/765c412c-fdca-11e6-9828-4dffb60be4eb.png)

xi:
#106 only fixed one very specific case. I have found at least to cases that are still broken:

- a resource header directly below the header should be overlapped, just like the hero image.
- an alert not followed by a resource header or a hero image should have a bottom margin.

xi:
Another issue: On mobile, the header has a bottom padding of `0.5em`, so it does not properly overlap the hero unit/alert/resource header.

xi:
> an alert not followed by a resource header or a hero image should have a bottom margin.

Example: Success alert after editing a project

PR #128, opened 2017-03-28
improve resource header

### LINKED ISSUE #118: resource image for project detail on mobile too big
just so we don't forget about it: the min-height should be reduced for mobile, it's too high.

xi:
fixes #118 and part of #82.


vellip:
Looks good and works for me!

PR #129, opened 2017-03-28
RFC: embed: keep url on login

xi:
This change would allow users to stay on the same view after login/logout. This is nice for users, but it requires an extension to the embed API that would allow attackers to embed anything, not just project detail pages. What do we prefer?

### DIFF
--- a/apps/embed/templates/meinberlin_embed/embed.html
+++ b/apps/embed/templates/meinberlin_embed/embed.html
@@ -17,7 +17,7 @@
     {% block header %}{% endblock %}
 </head>
 
-<body data-url="{% url 'project-detail' slug %}">
+<body data-url="{{ view.initial_url }}">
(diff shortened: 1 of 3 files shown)


vellip:
As said I'd prefer to just observe it as is right now. If we notice the need for it later on, this is something that still can be added.

2e2a:
What would be the thread scenario for such an attack?

xi:
Maybe "attacker" is a big word here.

- People could try to embed unsuported views, resulting in poor user experience. This would reflect badly on the project as a whole.
- To avoid phising, we have a strict rule to not embed any security related views. This pull request would stop enforcing that rule.

2e2a:
Then I would suggest to keep it as is.
There is anyway not that much navigation inside a project, so its should not be too bad from the user perspective to loose state upon login.

PR #130, opened 2017-03-29
some additional embed base styling

vellip:


apps/embed/templates/meinberlin_embed/embed.html
```diff
@@ -20,10 +20,10 @@
 <body data-url="{% url 'project-detail' slug %}">
     <header class="embed-header">
         {% if request.user.is_anonymous %}
-            <a href="{% url 'account_login' %}">{% trans 'Login' %}</a>
+            <a href="{% url 'account_login' %}" class="js-embed-login button button--light">{% trans 'Login' %}</a>
```
xi: I think the `js-*` class sneaked back in in the merge

vellip: ah i was afraid something like that would happen.

vellip: Now it should be the same as it is in master.

xi:
Much better

PR #131, opened 2017-03-29
add error handling for embed

vellip:


meinberlin/assets/scss/style.scss
```diff
@@ -26,7 +26,9 @@
 @import 'components/commenting';
 @import 'components/comments';
 @import 'components/dropdown';
+@import "components/embed-confirm";
```
xi: merge error?

meinberlin/static/embed.js
```diff
@@ -114,6 +114,35 @@ $(document).ready(function () {
...
+      default:
+        text = django.gettext('Something went wrong!')
+        break
+    }
+
+    $error.text(text)
+    $error.append($close)
+    $error.prependTo($('#embed-status'))
+
+    $close.on('click', removeMessage)
```
xi: I would prefer if `$error` was completely clickable.

meinberlin/assets/scss/components/_embed-status.scss
```diff
@@ -0,0 +1,28 @@
+.embed-status {
+    margin-top: 1em;
```
xi: IMHO this looks strange

meinberlin/assets/scss/components/_embed-status.scss
```diff
@@ -0,0 +1,28 @@
...
+
+.embed-status__message--error {
+    background-color: $brand-danger;
+    color: contrast-color($brand-danger);
+}
+
+.embed-status__close {
+    float: right;
+    color: contrast-color($brand-danger);
+}
```
xi: I would prefer to reuse `alert` skin and only use different layout.

meinberlin/static/embed.js
```diff
@@ -114,6 +114,35 @@ $(document).ready(function () {
     }
   }, false)
 
+  $(document).ajaxError(function (event, jqxhr) {
+    var text
+    var $error = $('<p class="embed-status__message embed-status__message--error"></p>')
+    var $close = $('<button class="embed-status__close"><i class="fa fa-times"></i></button>')
```
xi: `<i>` should have `aria-hidden` and there needs to be a text alternative.

xi:
I noticed that django rarely sends non-200 response codes. Invalid form inputs return with 200 as does a redirect to login due to missing permissions. This code would break if the backend behavior would change.

xi:
The ajax error handler is registered globally, so it is also triggered for requests that are triggered from the content, not just the shell. I am not sure if this is a good idea, but we can leave it for now and see how it works in practice.

meinberlin/static/embed.js
```diff
@@ -138,6 +138,7 @@ $(document).ready(function () {
     $error.text(text)
     $error.append($close)
     $error.prependTo($('#embed-status'))
+    $close.attr('title', django.gettext('Close'))
```
vellip: Access to title attribute information is not supported uniformly by screen readers. See: https://www.w3.org/TR/html/dom.html#the-title-attribute

meinberlin/static/embed.js
```diff
@@ -114,6 +114,38 @@ $(document).ready(function () {
...
+        break
+    }
+
+    $error.text(text)
+    $error.append($close)
+    $error.prependTo($('#embed-status'))
+    $close.attr('title', django.gettext('Close'))
+
+    $error.on('click', removeMessage)
+    $close.on('click', removeMessage)
```
vellip: This wouldn't be necessary anymore then.

meinberlin/assets/scss/components/_alert.scss
```diff
@@ -8,15 +8,32 @@
...
+    color: inherit;
+}
+
+.alert--small {
+    padding: 0.5em;
+}
+
+.alert--dialog-box {
+    &:first-child {
+        padding-top: calc(16px + #{$padding * 1.2});
```
vellip: I feel like this deserves a comment.

meinberlin/static/embed.js
xi: It is for keyboard interaction.

meinberlin/static/embed.js
xi: Both `title` and `aria-label` have pros and cons. I would like to discuss this separately.

meinberlin/static/embed.js
vellip: I tested it, worked for me without as well.

meinberlin/assets/scss/components/_alert.scss
vellip: Sorry, I meant the calculation. :X

PR #132, opened 2017-03-29
show project embed code in dashboard

xi:


xi:
The width and height attributes will be refined later.

PR #136, opened 2017-03-29
implement warning for 3rd party cookies disabled

vellip:
first commit was supposed to be named "add test", duh.

### DIFF
--- a/meinberlin/assets/scss/components/_alert.scss
+++ b/meinberlin/assets/scss/components/_alert.scss
@@ -21,6 +21,11 @@
         background-color: $brand-danger;
         color: contrast-color($brand-danger);
     }
+
+    &.info {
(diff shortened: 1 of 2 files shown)


meinberlin/assets/scss/components/_alert.scss
```diff
@@ -21,6 +21,11 @@
         background-color: $brand-danger;
         color: contrast-color($brand-danger);
     }
+
+    &.info {
+        background-color: $brand-primary;
+        color: contrast-color($brand-primary);
+    }
```
xi: IMHO `.info` should not have special styling but just use the default.

xi:
I am not convinced about the styling of the info alert, but I guess we will have to reconsider the alert styling in general.

PR #140, opened 2017-03-30
2017 03 pv embed header footer

vellip:


apps/embed/templates/meinberlin_embed/embed.html
```diff
@@ -29,6 +29,9 @@
     </header>
     <main class="embed-main"></main>
     <footer class="embed-footer">
+        <a href="/" target="_blank">
+            <img src="{% static 'images/logo_horizontal.svg' %}" alt="" class="embed-footer__logo">
```
xi: the alt text should be non-empty

PR #142, opened 2017-03-30
update platform link with currentPath

vellip:
I wasn't sure anymore why we didn't do this. It's just a proposal at this point.


xi:
AFAIR, we decided not to do this.

PR #144, opened 2017-03-31
2017 03 pv ideas styling

vellip:
adapt design for idea list


meinberlin/assets/scss/components/_list_item.scss
```diff
@@ -8,31 +8,34 @@
...
-
 .list-item__title {
+    margin: 0 0 0.4rem;
     font-size: $font-size-lg;
+
+    a {
+        color: inherit;
+        text-decoration: none;
+
+        &:hover {
```
xi: `:focus` missing

apps/ideas/templates/meinberlin_ideas/idea_list.html
```diff
@@ -11,19 +11,20 @@
...
+                {% endfor %}
 
-            {% if idea_list.count == 0 %}
-            {% trans "Nothing to show" %}
-            {% endif %}
+                {% if idea_list.count == 0 %}
+                    {% trans "Nothing to show" %}
+                {% endif %}
+                {% include "meinberlin_contrib/includes/pagination.html" %}
+            </div>
```
xi: I prefer having the list items in their own div

meinberlin/assets/scss/components/_list.scss
```diff
@@ -0,0 +1,10 @@
+.list--highlight {
+    padding-bottom: 2em;
+
+    background-color: $bg-tertiary;
+
+    .filter-bar:first-child {
+        position: relative;
+        top: -0.8em;
+    }
```
xi: Can we convert this to a filter bar modifier?

Also: How does this work with several lines?

vellip: Good idea, just had to figure out how to not have it always since it's in its own include. 
screen shot 2017-04-04 at 13 05 12
Looks okay to me. I mean it's not too desirable, but then again, I doubt having *that* many filters is desirable either.

xi:
@vellip I talked to @damusp and we did some hack-ish changes in 5c02c45. What do you think of this?

vellip:
Well, it looks alright, though I am not overly convinced of the element containing the block.

PR #147, opened 2017-03-31
adapt rating button styling

vellip:


meinberlin/assets/scss/components/_rating.scss
```diff
@@ -39,12 +41,20 @@
         color: contrast-color($color);
     }
 
+    .fa {
+        margin-right: 0.3em;
+    }
```
xi: In other places we use the `i` element for styling.

meinberlin/assets/scss/components/_rating.scss
```diff
@@ -39,12 +41,20 @@
...
     &.is-read-only,
     &:disabled {
         color: $color;
         background: none;
         border: 0;
     }
+
+    .comment & {
+        padding: 0.3em 0.6em;
+    }
```
xi: Why is this necessary. Can we use a modifier instead?

vellip: Ratings are render via react and ratings for comments are supposed to be smaller by design. So even if we'd use a modifier we'd have to alter the core.

xi: Ok, can we move this code to `comment.scss` or `shame.scss` then?

xi:
I was wondering whether `.rating-button` should share code with `.button`. @damusp said no.

PR #148, opened 2017-04-03
Brainstorming with a map

2e2a:
Depends on: https://github.com/liqd/adhocracy4/pull/57

Notes:
- The CSS change seems to break the categories widget.


xi:
5eb3b3a is incomplete. The data still needs to be JSON encoded. It may be good to define a template filter for that which could also be used in core.

xi:
> The CSS change seems to break the categories widget.

@vellip this is about removing the global `max-width 100%`. Can you please look into that? I also noticed that the image upload is broken.

vellip:
the max-width regressions should be fixed now.

apps/mapideas/templates/meinberlin_mapideas/mapidea_detail.html
```diff
@@ -69,8 +69,8 @@ <h1 class="resource-header__title">{{ object.name }}</h1>
                     style="height: 300px;"
                     data-map="display_point"
                     data-baseurl="{{ map_url }}"
-                    data-point="{{ object.point }}"
-                    data-polygon="{{ object.module.settings_instance.polygon }}"
+                    data-point="{{ point }}"
+                    data-polygon="{{ polygon }}"
```
xi: These values still need JSON encoding to work reliably.

vellip: That's happening in core. https://github.com/liqd/adhocracy4/pull/57/commits/05cb62a7cd40cf4a540b1319a41ba1b16473220b

PR #152, opened 2017-04-04
Fix webpack development dependency resolving

kleingeist:
Webpack does not find dependencies for `npm link`ed modules
that are available in the currently build module.
This is documented at
http://webpack.github.io/docs/troubleshooting.html#npm-linked-modules-
don-t-find-their-dependencies

If modules are available in the linked module, as well as the
currently build module, dependencies may be included twice if
the proposed `fallback` option is used.
Setting resolve[Loader].root forces webpack to resolve all
dependencies only once against the local directory.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -74,6 +74,9 @@ module.exports = {
     // against the local directory.
     root: path.resolve('./node_modules')
   },
+  resolveLoader: {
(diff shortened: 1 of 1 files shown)


xi:
Nice! This issue was open for far too long.

PR #153, opened 2017-04-04
Add watch target that rebuilds assets on change

kleingeist:
The `watch` target starts a dev server and rebuilds js and css files
with webpack on changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ PO_FILES := $(shell find . -name '*.po')
 help:
 	@echo mein.berlin development tools
 	@echo
-	@echo It will either use a exisiting virtualenv if it was entered
+	@echo It will either use an exisiting virtualenv if it was entered
 	@echo before or create a new one in the .env subdirectory.
 	@echo
 	@echo usage:
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,11 @@ help:
 	@echo
 	@echo usage:
 	@echo
-	@echo   make install      -- install dev setup
-	@echo   make build        -- build js and css and create new po and mo files
-	@echo   make lint         -- lint javascript and python
-	@echo   make server       -- start a dev server
-	@echo   make watch        -- start a dev server and rebuild js and css files\
-	                             on changes
+	@echo "  make install      -- install dev setup"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,6 +4,22 @@ SCSS_FILES := $(shell find 'meinberlin/assets/scss' -name '*.scss')
...
+	@echo
+	@echo usage:
+	@echo
+	@echo   make install      -- install dev setup
+	@echo   make build        -- build js and css and create new po and mo files
+	@echo   make lint         -- lint javascript and python
+	@echo   make server       -- start a dev server
+	@echo   make watch        -- start a dev server and rebuild js and css files\
+	                             on changes
+	@echo
```
xi: The formatting is lost when I run `make`. I think wrapping it in quotes would preserve spaces.

Makefile
```diff
@@ -4,6 +4,22 @@ SCSS_FILES := $(shell find 'meinberlin/assets/scss' -name '*.scss')
...
+	@echo mein.berlin development tools
+	@echo
+	@echo It will either use a exisiting virtualenv if it was entered
+	@echo before or create a new one in the .env subdirectory.
+	@echo
+	@echo usage:
+	@echo
+	@echo   make install      -- install dev setup
+	@echo   make build        -- build js and css and create new po and mo files
+	@echo   make lint         -- lint javascript and python
```
xi: This is misleading as there are also linters for scss, markdown, and html.

Makefile
```diff
@@ -12,12 +12,11 @@ help:
...
-	@echo   make install      -- install dev setup
-	@echo   make build        -- build js and css and create new po and mo files
-	@echo   make lint         -- lint javascript and python
-	@echo   make server       -- start a dev server
-	@echo   make watch        -- start a dev server and rebuild js and css files\
-	                             on changes
+	@echo "  make install      -- install dev setup"
+	@echo "  make build        -- build js and css and create new po and mo files"
+	@echo "  make lint         -- lint all project files"
+	@echo "  make server       -- start a dev server"
```
kleingeist: "make server" is also available in a4-opin, although it is not documented there

PR #155, opened 2017-04-05
Remove htmlhint linter

kleingeist:


xi:
Just to be sure: Would it be sufficient to just disable certain rules instead of the complete linter?

kleingeist:
it would be possible, but i would advise against it, because i expect many failures to occur and thus an evergrowing exclude list.

xi:
Ok, thanks for the research!

PR #156, opened 2017-04-05
Add moderator management to dashboard

xi:
This is still very incomplete. I do not have time to continue working on this at the moment, so it is up to anyone who is interested.

Note that I am not too experienced with django yet, so you may need to make significant adjustments.


apps/dashboard/forms.py
```diff
@@ -273,3 +276,42 @@ class ProfileForm(forms.ModelForm):
...
+                missing.append(email)
+
+        if missing:
+            messages.error(
+                self.request,
+                _('Following e-mails are not registered: ') + ', '.join(
+                    missing)
+            )
+
+        return users
```
xi: Where is this function called?

vellip: That's built-in, see 5th bullet point: https://docs.djangoproject.com/en/1.10/ref/forms/validation/

apps/users/fields.py
```diff
@@ -0,0 +1,23 @@
...
+    default_validators = [RegexValidator(
+        # a list of emails, separated by commas with optional space after
+        regex=r'^([^@]+@[^@\s]+\.[^@\s,]+((,\s?)|$))+$',
+        message=_('Please enter correct e-mail addresses, separated by '
+                  'commas.')
+    )]
+
+    widget = Input(attrs={
+        'placeholder': 'jana@beispiel.de, tobias@beispiel.de, '
+                       'swetlana@beispiel.de,…'
```
xi: We should change that placeholder.

apps/dashboard/templates/meinberlin_dashboard/project_moderators.html
```diff
@@ -0,0 +1,47 @@
...
+    <form method="POST">
+        {% csrf_token %}
+        <div class="form--single-line">
+            <div class="l-col-2">
+                {% include 'meinberlin_contrib/includes/form_field.html' with field=form.add_moderators %}
+            </div>
+            <div class="l-col-2">
+                <input type="submit" class="button button--secondary" value="{% trans 'Add' %}"/>
+            </div>
+        </div>
```
xi: The layout here feels very hacky to me.

vellip: And how so exactly?

xi:
- I have an error when I delete a moderator:

      Reverse for 'dashboard-project-moderators' with arguments '()' and keyword arguments '{'organisation_slug': 'organisation'}' not found. 1 pattern(s) tried: ['dashboard/(?P<organisation_slug>[-\\w_]+)/projects/(?P<slug>[-\\w_]+)/moderators/$']

- I expected a success alert after adding a moderator which did not appear.
- The navigation is a bit unclear. I guess we will refactor the whole dashboard UI at some point.
- I feel that the `CommaSeparatedEmailField` is uncessary, but I also do not have a problem with it.

apps/dashboard/templates/meinberlin_dashboard/project_moderators.html
xi: - It does not feel very reusable to me.
- Wrapping the button in `.l-col-2` looks strange IMHO.
- At the very least, the `margin-top` should be removed for mobile.

This is really more of a feeling at this point. I do not really have a better idea.

vellip:
The error was fixed in one of my latter commits. 

The problem with the success message is that you might run into the case where you add 5 moderators and 2 of them are not registered. So you'd have the "Following E-Mail addresses are not registered: blablabla" and right underneath a "Moderators successfully updated". So I left it at that, you can see the newly added moderators right on the page then anyway. But if you have a proposal that works for mentioned case I am all ears.

I think it's gonna be useful later, especially in regards to private projects.

apps/users/fields.py
```diff
@@ -14,8 +14,8 @@ class CommaSeparatedEmailField(Field):
     )]
 
     widget = Input(attrs={
-        'placeholder': 'jana@beispiel.de, tobias@beispiel.de, '
-                       'swetlana@beispiel.de,…'
+        'placeholder': 'maria@beispiel.de, peter@beispiel.de, '
+                       'nicola@beispiel.de,…'
```
xi: We might wanto to use a [reserved domain](https://tools.ietf.org/html/rfc2606) like `example.com`.

xi:
> The problem with the success message is that you might run into the case where you add 5 moderators and 2 of them are not registered.

Maybe something like "3 Moderators were added".

xi:
LGTM. @2e2a what do you think of the backend implementation?

kleingeist:
the backend looks fine to me

apps/dashboard/mixins.py
```diff
@@ -56,3 +57,24 @@ def post(self, request, *args, **kwargs):
...
+
+
+class DashboardModRemovalMixin:
+    def post(self, request, *args, **kwargs):
+        if 'submit_action' in request.POST:
+            pk = int(request.POST['moderator_pk'])
+            user = get_object_or_404(User, pk=pk)
+            project = self.get_object()
+            can_edit = request.user.has_perm('a4projects.edit_project',
+                                             project)
```
2e2a: Better check for `meinberlin_organisations.initiate_project`, as moderator should not be able to add moderators.

PR #157, opened 2017-04-05
draft and publish flow

vellip:
styling's still missing.


apps/dashboard/mixins.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class DashboardProjectPublishMixin:
+    def post(self, request, *args, **kwargs):
+        pk = int(request.POST['project_pk'])
+        project = get_object_or_404(Project, pk=pk)
+        can_edit = request.user.has_perm('a4projects.edit_project', project)
+
+        if not can_edit:
+            raise PermissionDenied
```
2e2a: There is no need for a permission check here. It is handled by the `PermissionRequiredMixin`.

apps/dashboard/templates/meinberlin_dashboard/includes/project_list_item.html
```diff
@@ -1,30 +1,32 @@
...
+            <a href="{% url 'dashboard-project-edit' organisation_slug=view.organisation.slug slug=project.slug %}"
+               class="button button--light">
+                <i class="fa fa-pencil" aria-hidden="true"></i>
+                {% trans 'Edit'%}
+            </a>
+        </span>
+        <form method="post" class="list-item__action">
+            {% csrf_token %}
+            <input type="hidden" name="project_pk" value="{{ project.pk }}">
+            <button type="submit" name="{{ project.is_draft|yesno:'publish,unpublish' }}"
```
2e2a: I would prefer to have a fixed form field name (e.g. publish) and set the value (e.g. true/false), or is this a common pattern in HTML forms to just use the field name?

vellip: I changed it, makes sense.

apps/dashboard/mixins.py
vellip: Without that check a user that is initiator from organisation A can publish a project from organisation B. The PermissionRequiredMixin only checks if you have access to the particular view, so you can just go to the Dashboard from organisation A which succeeds the PermissionRequiredMixin and then change the project_pk (hidden field) to another pk.

PR #158, opened 2017-04-05
pass parameters to javascript by data attributes

xi:
this fixes an XSS vulnaraibility

See also https://github.com/liqd/adhocracy4/pull/61

### DIFF
--- a/apps/documents/templatetags/react_paragraphs.py
+++ b/apps/documents/templatetags/react_paragraphs.py
@@ -1,3 +1,5 @@
+import json
+
 from ckeditor_uploader.widgets import CKEditorUploadingWidget
 from django import template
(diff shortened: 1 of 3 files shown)


apps/documents/templatetags/react_paragraphs.py
```diff
@@ -25,7 +27,7 @@ def react_paragraphs(context, doc, module):
     context = {
         'document': document,
```
slomo: We discussed this feature earlier and you were quire eager to also use `json.dumps` for the document value. Did you have any second thoughts on that?

xi: No, it was just due to a lack of time. I think there is plenty of room for further refactoring.

vellip:
Works!

PR #159, opened 2017-04-05
Add form for organisations to the dashboard

kleingeist:
Allows for participants with the modify_organisation permission to change the organisation name


apps/dashboard/templates/meinberlin_dashboard/organisation_form.html
```diff
@@ -0,0 +1,13 @@
+{% extends "meinberlin_dashboard/base_dashboard.html" %}
+{% load i18n %}
+
+{% block dashboard_content %}
+    <h3>{% trans "Edit your organisation information" %}</h3>
```
vellip: That'd better be an h1. We'll have to take care of the additional margin that comes with it separately.

PR #164, opened 2017-04-10
Add moderator feedback to budgeting

kleingeist:


apps/moderatorfeedback/fields.py
```diff
@@ -0,0 +1,117 @@
...
+from django.core.exceptions import FieldError
+from django.db import models
+from django.utils.encoding import force_text
+from django.utils.functional import curry
+from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext_lazy as _
+
+
+class ModeratorFeedbackField(models.CharField):
+    description = _("")
```
2e2a: Add some basic description here

apps/moderatorfeedback/fields.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def __init__(self, *args, **kwargs):
+        kwargs['max_length'] = 254
+        kwargs['null'] = True
+        kwargs['default'] = None
+        kwargs['blank'] = True
+        super(ModeratorFeedbackField, self).__init__(*args, **kwargs)
+
+    def deconstruct(self):
```
2e2a: No need for a custom deconstuct() here, as we use just the default arguments that are removed by the deconstruct of the super class.

xi:
Can we use color variables for the feedback colors? (e.g. `$brand-success`)

apps/moderatorfeedback/fields.py
```diff
@@ -0,0 +1,117 @@
...
+            if len(choices) == 0:
+                raise ValueError()
+        except TypeError:
+            return [
+                checks.Error(
+                    "ModeratorFeedbackField must define a 'choices' "
+                    "attribute.",
+                    hint=None,
+                    obj=self,
+                    id='fields.E???',  # how to choose the error id?
```
2e2a: `hint` and `id` are optional

apps/moderatorfeedback/fields.py
```diff
@@ -0,0 +1,117 @@
...
+    def contribute_to_class(self, cls, name, **kwargs):
+        """
+        Add field specific attributes to the enclosing model.
+
+        This method is called in the context of the enclosing
+        model. It is used to set field specific attributes on
+        the model. For example it is used to set the generic
+        model.get_XXX_display methods for the choice fields.
+        In this case it is used to make the moderator feedback
+        context available.
```
2e2a: I would skip the generic part of explanation and document just the functionality, i.e. the new attribute.

apps/moderatorfeedback/templates/meinberlin_moderatorfeedback/moderatorfeedback.html
```diff
@@ -0,0 +1,7 @@
+{% if item.get_moderator_feedback %}
+    {% with item.get_moderator_feedback as moderator_feedback %}
+    <span class="moderator-feedback moderator-feedback--{{ moderator_feedback.value_class }}">
+        {% if moderator_feedback.value %}{{ moderator_feedback.display }}{% endif %}
+    </span>
+    {% endwith %}
+{% endif %}
```
2e2a: Better add an `include` subfolder to indicate that this is not a standalone template.

PR #169, opened 2017-04-10
Adapt to core changes in maps

xi:
See https://github.com/liqd/adhocracy4/pull/66


apps/mapideas/templates/meinberlin_mapideas/mapidea_detail.html
```diff
@@ -12,12 +12,6 @@
...
 {% block additional_content %}
     {% if object.point %}
-        <div
-                style="height: 300px;"
-                data-map="display_point"
-                data-baseurl="{{ map_url }}"
-                data-point="{{ point }}"
-                data-polygon="{{ polygon }}"
-        ></div>
+        {% map_display_point object.point object.module.settings_instance.polygon %}
```
slomo: You probably want to access object.module.area_settings.polygon here instead.

xi: Feel free to change that.

apps/mapideas/templates/meinberlin_mapideas/mapidea_detail.html
```diff
@@ -12,6 +12,6 @@
 
 {% block additional_content %}
     {% if object.point %}
-        {% map_display_point object.point object.module.settings_instance.polygon %}
+        {% map_display_point object.point object.module.area_settings.polygon %}
```
slomo: This shouldn't work because the realname of that reverse relation is `areasettings_settings`, I updated my other comment a few minutes later. Sorry.

xi: Fixed in 35378b0

PR #172, opened 2017-04-11
Pin django lts for pyup

2e2a:


xi:
I think according to pep8 there should be two spaces before the comment. But not so important.

PR #194, opened 2017-04-11
include leaflet in webpack build process

vellip:


xi:
LGTM. I created #195 for the performance issues this creates.

Issue #166, opened 2017-04-10
Fix filter bar overlap


xi:
The filter bar is supposed to overlap its box in some cases:

![screenshot from 2017-04-10 14 37 24](https://cloud.githubusercontent.com/assets/202576/24863366/d0ce4bc2-1e00-11e7-972c-dc597afb7d89.png)

As you can see, the current implementation displaces the stats of the first list item.

A lot of work and thought was already put into this in #144, but with limited success. A workaround which disables the overlap was pushed in #165.

vellip:
Hint: This only occurs if there's not left aligned filter but only the ordering widget.

PR #200, opened 2017-04-11
Fix map list UI

xi:
Most importantly, this allows to switch between list and map mode. All filters, ordering, and pagination are preserved. The pagination is ignored in map mode. The default mode is list.

Note that as part of this pull request, I refactored the button styling, notably the `button--select` modifier.

Some possible next steps:

-   define filters
-   when changing view, jump directly back to the list/map
-   zoom buttons on the map are missing
-   the leaftlet popups need styling
-   the filterbar should overlap the map (see #166)

### DIFF
--- a/apps/contrib/templates/meinberlin_contrib/includes/map_filter_and_sort.html
+++ b/apps/contrib/templates/meinberlin_contrib/includes/map_filter_and_sort.html
@@ -4,13 +4,13 @@
     {% if mode == 'map' %}
         {% combined_url_parameter request.GET mode='list' as url_par %}
         <a class="button button--light" href="{{ url_par }}">
-            <i class="fa fa-map" aria-hidden="true"></i>
+            <i class="fa fa-list" aria-hidden="true"></i>
             {% trans 'view' %}
         </a>
     {% else %}
         {% combined_url_parameter request.GET mode='map' as url_par %}
         <a class="button button--light" href="{{ url_par }}">
-            <i class="fa fa-list" aria-hidden="true"></i>
+            <i class="fa fa-map" aria-hidden="true"></i>
             {% trans 'view' %}
         </a>
     {% endif %}
--- a/apps/contrib/templates/meinberlin_contrib/includes/map_filter_and_sort.html
+++ b/apps/contrib/templates/meinberlin_contrib/includes/map_filter_and_sort.html
@@ -3,13 +3,13 @@
 <div class="filter-bar">
     {% if mode == 'map' %}
         {% combined_url_parameter request.GET mode='list' as url_par %}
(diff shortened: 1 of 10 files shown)


apps/mapideas/views.py
```diff
@@ -43,6 +43,12 @@ class MapIdeaListView(map_mixins.MapItemListMixin, module_views.ItemListView):
     model = models.MapIdea
     filter_set = MapIdeaFilterSet
 
+    def dispatch(self, request, **kwargs):
+        self.mode = request.GET.get('mode', 'list')
```
vellip: I'd rather see the map as the default view for **Map**Ideas.

apps/contrib/templates/meinberlin_contrib/includes/map_filter_and_sort.html
```diff
@@ -0,0 +1,29 @@
+{% load i18n contrib_tags %}
+
+<div class="filter-bar">
+    {% if mode == 'map' %}
+        {% combined_url_parameter request.GET mode='list' as url_par %}
+        <a class="button button--light" href="{{ url_par }}">
+            <i class="fa fa-map" aria-hidden="true"></i>
+            {% trans 'view' %}
```
vellip: This is not accessible. We should overwrite the button's link text here with `aria-label="view on map"`

apps/contrib/templates/meinberlin_contrib/includes/map_filter_and_sort.html
```diff
@@ -0,0 +1,29 @@
...
+        </a>
+    {% else %}
+        {% combined_url_parameter request.GET mode='map' as url_par %}
+        <a class="button button--light" href="{{ url_par }}">
+            <i class="fa fa-list" aria-hidden="true"></i>
+            {% trans 'view' %}
```
vellip: This is not accessible. We should overwrite the button's link text here with `aria-label="view as list"`

vellip:
The fontsize for the view toggle is too small. But I suppose this can be done later as well. Also we need to be careful with icons in buttons:
Design wise there are two buttons; small ones with regular font weight and big ones with bold font weight. 
In your toggle (and any other buttons where we have icons) the icon ends up with `font-weight: bold` which actually doesn't work for icons. For the map icon it's pretty obvious. But that can be taken care of later.

xi:
> In your toggle (and any other buttons where we have icons) the icon ends up with font-weight: bold which actually doesn't work for icons.

The `.fa` class sets `font: normal normal normal 14px/1 FontAwesome;` which should take care of that.

> The fontsize for the view toggle is too small.

The font size in the filter bar is actually smaller than regular in the designs.

PR #201, opened 2017-04-12
use berlin tileserver

xi:


vellip:
breaks bounding box, see #208

PR #202, opened 2017-04-12
map popup styling

xi:


vellip:
That's alright for now.

PR #204, opened 2017-04-12
refactor labels

xi:
This merges the concepts of labels, categories, and moderator feedback on a CSS level.

Note that `.label` is used in core (comments), so we are somewhat limited in the default styling.

Also note that list items may contain an empty `.list-item__labels`. This is because it is currently rather complicated to know whether there is moderator feedback. We might think about refactoring that.


apps/moderatorfeedback/templates/meinberlin_moderatorfeedback/includes/moderatorfeedback.html
```diff
@@ -1,9 +1,7 @@
...
-        {% endif %}
-    {% endwith %}
-{% endif %}
+{% with item.get_moderator_feedback as moderator_feedback %}
+    {% if moderator_feedback.value %}
+        <span class="label label--big label--{{ moderator_feedback.value_class }} {{ cls }}">
+            {{ moderator_feedback.display }}
+        </span>
+    {% endif %}
+{% endwith %}
```
xi: The if was not actually redundant: It checks whether the function `item.get_moderator_feedback()` exists. @kleingeist what do you think?

Issue #197, opened 2017-04-11
Consistent URL design


xi:
Most of our URLs are not hierarchical. For example, the path to a project is `/projects/{project_slug}/` rather than `/organisations/{organisation_slug}/projects/{project_slug}`. However, the paths in the dashboard *are* hierarchical.

The hierarchical style lead to some security issues (e.g. #162). The issue was that the URL contains redundant information (i.e. the organisation is given both explicitly and implicitly from the project). We failed to check that the two match up which allowed initiators from one organisation to edit another organisation's projects.

I personally prefer the hierarchical style, e.g. on github. But note that github issue ids are scoped to the project, so there is no redundant information.

I do not think that one style is more secure than the other. But I mixing them leads to confusion. We should pick one of these options, ideally before going live:

1. Use hierarchical paths and check for consistency.
2. Use hierarchical paths with scoped slugs.
3. Use non-hierarchical paths.

Given the current code base, I think it third option is the easiest to implement. I also think that hierarchical paths are mostly useful for people with a technical background, so not really that important for our target group.

PR #205, opened 2017-04-12
Remove organisation slug from dashboard urls if not needed

2e2a:
Fixes https://github.com/liqd/a4-meinberlin/issues/197


vellip:
Works alright.

PR #206, opened 2017-04-12
refactor item detail templates

vellip:
* add template tag for permissions
* add template tag for item urls
* unify/update permissions for item subclasses (propose -> add, modify -> change)
* change url namespaces from item apps to reflect the app_label
* add basic item_detail and let mapidea, idea and budgeting extend it

Next steps would be to do the same with the create/update forms and possibly the delete view.

Right now, the `get_item_add_permission` tag is useless as there is no item to get the app_label/model_name. I felt it was out of scope of this pull request to get a solution for that one because this permission is only used in the list-view which ain't topic of this PR.


kleingeist:
looks good to me. 
only thing to note is that assignment tags will be deprecated with version 1.9 (simple_tag can now store results in a template variable and should be used instead). 
but for now it is fine to use them

apps/contrib/templatetags/item_tags.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@register.assignment_tag
+def get_item_update_url(item):
+    return get_item_url(item, 'update')
+
+
+@register.assignment_tag
+def get_item_delete_url(item):
+    return get_item_url(item, 'delete')
```
xi: Why are there all these additional tags. We can just use `{% get_item_url object 'delete' as delete_url %}`, right?

xi:
LGTM, too. I feel that some documentation (how does this work, what conventions should I follow, why was this approach chosen over others) would be a good addition, but that can be done separately.

PR #227, opened 2017-04-19
Update fake-factory to Faker

kleingeist:
fake-factory has been renamed to Faker. Adapt to the name change and add the correct dependency


2e2a:
lgtm

PR #230, opened 2017-04-19
polygon presets

xi:
As the [core pull request](https://github.com/liqd/adhocracy4/pull/74) was rejected, we duplicate the complete widget here instead.

Still to do:

- ask the user for confirmation before replacing the content (required?)
- add a label/help text to the select


apps/maps/widgets.py
```diff
@@ -21,14 +24,29 @@ class Media:
             staticfiles_storage.url('leaflet.draw.css'),
         ]}
 
+    def get_presets(self, category):
+        presets = MapPreset.objects.filter(category=category).all()
```
vellip: Why's there an `.all()` after the filter?

xi: I thought it was necessary to actually execute the query. If it is not, feel free to remove it.

PR #231, opened 2017-04-19
Allow only one document per project

2e2a:
- add validatos as in (https://github.com/liqd/a4-opin/pull/788/commits/7e2ddee720ffcc03c28931a8e6ac865500ca1c2d)
- copy document tests from a4-opin

### DIFF
--- a/tests/documents/test_document_models.py
+++ b/tests/documents/test_document_models.py
@@ -1,0 +1,73 @@
+import pytest
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
+
+from adhocracy4.comments import models as comments_models
+from apps.documents.models import Document
+
+
+@pytest.mark.django_db
+def test_paragraph_save(paragraph):
+    assert '<script>' not in paragraph.text
+
+
+@pytest.mark.django_db
+def test_document_clean(document):
+    another_document = Document(
+        module=document.module,
+        creator=document.creator
+    )
+
+    with pytest.raises(ValidationError):
+        another_document.clean()
+
+
+@pytest.mark.django_db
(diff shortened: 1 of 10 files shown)


tests/documents/test_document_models.py
```diff
@@ -0,0 +1,73 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
+
+from adhocracy4.comments import models as comments_models
+from apps.documents.models import Document
+
+
+@pytest.mark.django_db
+def test_paragraph_save(paragraph):
+    assert '<script>' not in paragraph.text
```
kleingeist: what does this test do?
i would make this more explicit by creating a paragraph instance with a script tag in the context of this test instead of using the more decoupled value set in the default paragraph factory

tests/documents/test_document_models.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@pytest.mark.django_db
+def test_document_clean(document):
+    another_document = Document(
+        module=document.module,
+        creator=document.creator
+    )
+
+    with pytest.raises(ValidationError):
```
kleingeist: where does the exception come from?
i'd suggest to construct valid object in the default factory and either use an explicit InvalidDocumentFactory or construct it directly in the tests context.
i suggest to add this as a new issue, while merging this PR

tests/documents/test_document_models.py
2e2a: fixed in latest commit

tests/documents/test_document_models.py
2e2a: its raised in `Document.clean` via the `single_document_per_module` validator.

PR #232, opened 2017-04-20
Script to import berlin geo data

2e2a:
based on https://github.com/liqd/adhocracy3/blob/master/src/adhocracy_meinberlin/adhocracy_meinberlin/scripts/ad_import_geodata.py

### DIFF
--- a/apps/maps/management/commands/import_geodata.py
+++ b/apps/maps/management/commands/import_geodata.py
@@ -26,20 +26,19 @@ class Command(BaseCommand):
         self._import_regions()
 
     def _import_districts(self):
-        category = self._preset_category('Bezirk')
+        category = self._preset_category('Berlin')
         tmpfile = '/tmp/bezirke.json'
         url = 'http://fbinter.stadt-berlin.de/fb/' \
               'wfs/geometry/senstadt/re_bezirke/'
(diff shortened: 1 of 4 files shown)


apps/maps/management/commands/import_geodata.py
```diff
@@ -0,0 +1,93 @@
...
+            if not map_models.MapPreset.objects.filter(name=name).exists():
+                self._create_map_preset(name, feature, category)
+        os.remove(tmpfile)
+
+    def _preset_category(self, name):
+        try:
+            category = map_models.MapPresetCategory.objects.get(name=name)
+        except map_models.MapPresetCategory.DoesNotExist:
+            category = map_models.MapPresetCategory(name=name)
+            category.save()
```
vellip: What about using `get_or_create` instead here? https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create

vellip:
LGTM

apps/maps/management/commands/import_geodata.py
2e2a: Thanks, using this.

2e2a:
Updated ordering and category structure.

PR #234, opened 2017-04-20
Fix inclusion of the django debug toolbar

kleingeist:
The required steps to include the toolbar changes with the version bumped by pyup


2e2a:
lgtm

PR #236, opened 2017-04-20
Prevent projects containing phases without dates from publishing

kleingeist:
Unfortunately this does not prevent from removing start/end dates from already published projects. Is this required?

### DIFF
--- a/apps/dashboard/mixins.py
+++ b/apps/dashboard/mixins.py
@@ -54,13 +54,30 @@ class DashboardProjectPublishMixin:
                 raise PermissionDenied
 
             if request.POST['submit_action'] == 'publish':
-                project.is_draft = False
-                messages.success(request, _('Project successfully published.'))
+                phases = project.phases
+
+                # Assure that every phase has a start and an end date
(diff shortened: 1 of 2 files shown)


2e2a:
> Unfortunately this does not prevent from removing start/end dates from already published projects. Is this required?

I'd say yes. But we can do it separately.

apps/dashboard/mixins.py
```diff
@@ -54,13 +54,30 @@ def post(self, request, *args, **kwargs):
...
             if request.POST['submit_action'] == 'publish':
-                project.is_draft = False
-                messages.success(request, _('Project successfully published.'))
+                phases = project.phases
+
+                # Assure that every phase has a start and an end date
+                missing_date = False
+                for phase in phases:
+                    missing_date = missing_date \
+                        or not phase.start_date or not phase.end_date
```
2e2a: I would prefer to have it more explicit, e.g. `break` on first missing date.

2e2a:
lgtm

PR #237, opened 2017-04-21
refactor item delete

vellip:


apps/budgeting/templates/meinberlin_budgeting/proposal_confirm_delete.html
```diff
@@ -1,17 +1 @@
...
-
-    <form enctype="multipart/form-data" action="{{ request.path }}" method="post">
-        {% csrf_token %}
-        <input type="submit" class="button button--primary" value="{% trans 'Yes'%}" />
-    </form>
-    {% get_item_detail_url object as detail_url %}
-    <a href="{{ detail_url }}" class="button button--light">{% trans 'Cancel'%}</a>
-</div>
-{% endblock %}
+{% extends "meinberlin_contrib/item_confirm_delete.html" %}
```
xi: This should use different wording: "proposal" instead of "idea".

apps/contrib/templates/meinberlin_contrib/item_confirm_delete.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base.html" %}
+{% load i18n item_tags %}
+
+{% block title %}{% trans 'Confirm Idea Deletion' %}{% endblock %}
+
+{% block content %}
+<div class="l-wrapper resource-header">
+    <h1 class="resource-header__title">{% trans 'Are you sure you want to delete the idea?' %}</h1>
```
xi: The resource header component does not work here because "Are you sure you want to delete the idea?" is not the title of a resource.

xi:
See #276

PR #238, opened 2017-04-21
add map min-height

vellip:
To make it look like in the design.


xi:
Entries in `_shame.scss` should have extensive comments. I think the generic "used by a4" does not apply here.

PR #239, opened 2017-04-21
2017 04 pv refactor item forms

vellip:


apps/contrib/templates/meinberlin_contrib/item_form.html
```diff
@@ -0,0 +1,35 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+<div class="l-wrapper">
+    <nav class="breadcrumbs">
+        <ul>
+            {% block breadcrumbs %}
+            <li>You need to overwrite the breadcrumbs block</li>
```
xi: I would prefer to have a meaningful default here, e.g. a link to the homepage.

apps/contrib/templates/meinberlin_contrib/item_form.html
```diff
@@ -0,0 +1,35 @@
...
+            {% endblock %}
+        </ul>
+    </nav>
+
+    {% comment %} reuse title as heading {% endcomment %}
+    <h1>{% block title %}You need to overwrite the title block.{% endblock %}</h1>
```
xi: I would prefer to have a meaningful default here, e.g. "edit".

vellip: It isn't always edit though, it's either edit or create.

vellip:
I wanted to have one system for all the templates. In my opinion it would've made the template system even more complex if we have extends for some templates but includes for others. Having it like this doesn't introduce more complexity than the refactoring of the detail pages introduces. 
If it's about reducing as much code as possible: Thinking about it, what might still reduce code would be to introduce an item_edit_form and an item_create_form template

xi:
> In my opinion it would've made the template system even more complex if we have extends for some templates but include for others. 

I don't competely agree. We should be very specific about when to use which (and "always use extend" is very specific). But I think there is a reason most templating systems provide both options.

xi:
See #276

PR #241, opened 2017-04-23
draft date/timepicker

vellip:
pretty drafty approach to a split date and time picker. uses http://amsul.ca/pickadate.js/


xi:
The datepicker itself looks ok to me. However, I had issues with other things like localization (time choices start with 12:00 AM) and date format (the old value "2017-01-01 12:00:00" was parsed to 1916). I think syncing the date format between django and datepicker is crucial.

vellip:
I deliberately didn't spend any time on these things because I wasn't sure it was worth following that approach. But I can work on that then.

xi:
The github page says that the project is looking got maintainers. That might be an issue.

xi:
After the recent changes I get the following error in submit:

```
AttributeError at /dashboard/projects/map/
'list' object has no attribute 'date'
```

vellip:
This should be ready to test. Two things:
1. I am not overly convinced of the widget's `render` method. It feels a bit messy. I am not sure how to refactor it though.
2. Travis fails because the datepicker is not on npm and therefore has to be included in our repository. I don't know if we should just disable it via comments or there is a cleverer way.

xi:
If I saw correctly, the datepicker is this one: https://github.com/freqdec/datePicker

I think we could install it with npm from github. I also saw that the last commit was in 2014.

The code on our side looks ok (I did not spend a lot of time reading though). Still, I am not so convinced of the datepicker itself. It is kind of sad that this is still so hard to find a good one. But maybe my expectations are just too high.

This feature is not central to our application and I do not see major security issues, so I am fine with merging this.

meinberlin/assets/js/init-picker.js
```diff
@@ -0,0 +1,12 @@
+const $ = require('jquery')
+const datePickerController = require('datepicker')
+
+$(function () {
```
xi: Not to self: this needs to be adapted when after merging #279.

package.json
```diff
@@ -8,22 +8,23 @@
...
-    "bootstrap": "4.0.0-alpha.6",
     "sass-planifolia": "xi/sass-planifolia#0.3.1",
-    "react-flip-move": "2.9.2",
-    "tether": "1.4.0",
     "shariff": "1.24.1",
-    "leaflet": "0.7.7",
-    "leaflet-draw": "0.4.9"
+    "style-loader": "0.16.1",
+    "tether": "1.4.0",
+    "webpack": "1.15.0"
```
xi: at some point this was sorted into two categories: Build toolchain and actual runtime dependencies. `devDependencies` contains stuff like linters. I am not sure that this sorting is better.

xi:
The datepicker code is still in the final diff. It should probably be removed.

PR #248, opened 2017-04-24
Combined fixes for minor bugs occurred during the review

kleingeist:


xi:
The templates look good to me. @2e2a can you check the 629b4a2?

apps/budgeting/views.py
```diff
@@ -90,7 +90,10 @@ class ProposalModerateView(rules_mixins.PermissionRequiredMixin,
     template_name = 'meinberlin_budgeting/proposal_moderate_form.html'
 
     def get_success_url(self):
-        return self.request.path
+        # Multiforms are not working nicely with the UpdateView redirection,
+        # so we have to use internal knowledge of the instantiated multiforms
+        # names for the correct redirection
+        return self.object['feedback'].get_absolute_url()
```
2e2a: Could use `self.get_object().get_absolute_url()`

kleingeist: hm yes that should work and helps to hide the internal logic in that case. in general i am not happy with multiforms way of rewriting the object

2e2a:
lgtm

PR #253, opened 2017-04-24
Fix pydocstyle for pyup until flake8-docstrings is fixed

kleingeist:
Currently flake8-docstrings is broken with pydocstyle 2.0.0
This prevents pyup from updating pydocstyle. This is only temporary
required until the bug is fixed in flake8-docstrings

[0] https://gitlab.com/pycqa/flake8-docstrings/issues/19


xi:
LGTM

PR #257, opened 2017-04-25
Fix project templates and active phase info

### LINKED ISSUE #211: Project-Detail view is broken if the initial phase is in the future
If a projects initial phase starts in the future, the default `a4projects/project_detail.html` is rendered. 
This template is by default empty and should be overridden. Currently this is tried by `apps/projects/a4projects/project_detail.html` which is not working. Django is still using the a4 template. 
Furthermore the project_detauil template is missing logic and information if no phase has been started yet

2e2a:
Fixes #211


xi:
LGTM.

PR #260, opened 2017-04-25
Priorization of topics project

2e2a:
Mostly a copy from ideas, with the following changes:
- Custom phases
- Dedicated `CreateListView` that allows item creation only for moderators and shows an info to other users
- Add/update permissions limited to project moderators
- Image upload allowed in description


2e2a:
After a discussion we decided to remove the CreatePhase and create a moderation view accessible from the dashboard instead.

apps/topicprio/views.py
```diff
@@ -0,0 +1,147 @@
...
+    category = filters.CategoryFilter()
+
+    class Meta:
+        model = models.Topic
+        fields = ['category']
+
+
+class TopicMgmtView(DashboardBaseMixin,
+                    rules_mixins.PermissionRequiredMixin,
+                    module_views.ItemListView):
```
kleingeist: If no phase is active (first phase starts in the future) the management view is not rendered, as ItemlistView inherits from ProjectMixin which in turn tries to set `self.phase = self.project.active_phase or self.project.past_phases[0]` which has to fail.
it seems we can't reuse the user space logic for the management views as we assume there is always an active (or past) phase

kleingeist:
i agree to @vellip suggestion to skip the item/topic detail view in favor the edit link. as an admin brosing the management view i would expect to have the form shown immediately

apps/topicprio/views.py
kleingeist: we could share code between managment views, as i had the same problem here: https://github.com/liqd/a4-meinberlin/blob/master/apps/polls/views.py#L46

kleingeist:
as i am currently working on a react management form for polls, i'd like to ask if it would be an option to use react for prios, too?

2e2a:
Simplified the management views in the last commit

PR #261, opened 2017-04-25
poll detail component

xi:
this implements a react component and styling for the poll detail view. It currently uses hardcoded sample data as the API and models are not ready yet.

### DIFF
--- a/apps/polls/assets/Polls.jsx
+++ b/apps/polls/assets/Polls.jsx
@@ -51,6 +51,29 @@ var Poll = React.createClass({
     let total = counts.reduce((sum, c) => sum + c, 0)
     let max = Math.max.apply(null, counts)
 
+    let footer
+    if (!this.state.active) {
+      footer = '' + total + ' ' + django.ngettext('vote', 'votes', total)
+    } else if (this.state.showResult) {
+      footer = (
+        <button type="button" className="button button--light" onClick={this.toggleShowResult}>
+          { django.gettext('To poll') }
+        </button>
+      )
+    } else {
+      footer = (
+        <div>
+          <button type="submit" className="button button--secondary">
+            { django.gettext('Vote') }
+          </button>
+          &nbsp;
+          <button type="button" className="button button--light" onClick={this.toggleShowResult}>
+            { django.gettext('Show preliminary results') }
+          </button>
+        </div>
+      )
+    }
+
     return (
       <form onSubmit={this.vote}>
         <h2>{ this.state.title }</h2>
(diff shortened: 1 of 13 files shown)


apps/polls/assets/Polls.jsx
```diff
@@ -0,0 +1,118 @@
...
+              let percent = Math.round(option.count / total * 100)
+              let highlight = option.count === max
+
+              if (this.state.showResult || !this.state.active) {
+                return (
+                  <div className="poll-row" key={i}>
+                    <div className={'poll-row__bar' + (highlight ? ' poll-row__bar--highlight' : '')} style={{width: percent + '%'}} />
+                    <div className="poll-row__number">{ percent }%</div>
+                    <div className="poll-row__label">{ option.label }</div>
+                    &nbsp;
```
vellip: I don't think it's necessary here, looks fine without.

apps/polls/assets/Polls.jsx
```diff
@@ -0,0 +1,118 @@
...
+              } else {
+                return (
+                  <label className="poll-row" key={i}>
+                    <input
+                      className="poll-row__radio"
+                      type="radio"
+                      name="poll"
+                      value={i}
+                      defaultChecked={checked} />
+                    &nbsp;
```
vellip: I'd prefer a css solution here. You have the `poll-row__radio` class already there.

apps/polls/assets/Polls.jsx
```diff
@@ -0,0 +1,118 @@
...
+        label: 'Ja',
+        count: 22434
+      }, {
+        label: 'Nein',
+        count: 40062
+      }, {
+        label: 'Vielleicht',
+        count: 17627
+      }],
+      own: null,
```
vellip: Can we name this more explicitly like `ownVote` or something like that?

apps/polls/assets/Polls.jsx
```diff
@@ -0,0 +1,118 @@
...
+                    &nbsp;
+                    { option.label }
+                  </label>
+                )
+              }
+            })
+          }
+        </div>
+
+        { !this.state.active ? (
```
xi: I spend alot of time fighting with JSX here. Can you tell me how to do it?

xi:
> However, this does not yet work and wont work as implemented for multiple questions in one poll. Can you still add that?

I was not sure whether we want multiple polls in one react component or just multiple react components on one page. This decision might be influenced by the backend API. Whichever we land on, both should be easy to add later on.

apps/polls/assets/Polls.jsx
vellip: Maybe check this for reference: https://facebook.github.io/react/docs/conditional-rendering.html I personally prefer using what they call "Element variables" on that page.

PR #263, opened 2017-04-26
Initial polls project structure 

kleingeist:
Adds models and blueprints for polls.
It has been decided to add questions and choices in a moderator phase
before the public phases

This conflicts with PR #261


apps/polls/models.py
```diff
@@ -0,0 +1,37 @@
...
+        'Poll',
+        on_delete=models.CASCADE,
+    )
+
+
+class Vote(UserGeneratedContentModel):
+    poll = models.ForeignKey(
+        'Poll',
+        on_delete=models.CASCADE,
+    )
```
xi: The poll reference is redundant.

apps/polls/models.py
```diff
@@ -0,0 +1,37 @@
...
+from adhocracy4.modules import models as module_models
+
+
+class Poll(models.Model):
+    title = models.CharField(max_length=255)
+
+    module = models.ForeignKey(
+        module_models.Module,
+        on_delete=models.CASCADE,
+    )
```
2e2a: should be an Item, then you automatically have the module reference.

kleingeist: I didn't use an item, because it would pull in the userGeneratedContentModel. 
And polls are edited by admins/moderators not regular users

Issue #259, opened 2017-04-25
Styling issues in header area at 800px width


rittermo:
The header area shows some strange styling errors when the width is set to 800px (tested on Safari 10.0.3). 

1. The menu is always shown in an open state
2. The label for the project type has almost no vertical margin to the top navigation 

The seocond problem persists even for smaller widths. The first problem I can only produce at 800px viewport width.

![bildschirmfoto 2017-04-25 um 15 25 33](https://cloud.githubusercontent.com/assets/15341015/25387656/ea86ef66-29cb-11e7-8045-07ec79138bf4.png)

xi:
Nice catch! I guess this is because we sometimes use `min-width` and sometimes `max-width` in media queries. So on the exact breakpoint, none of the media quries match. We should really stick to `min-width` (mobile first). This may not always be possible, but we should at least find a solution for this specific case.

Note that the exact width depends on browser settings because the breakpoint is defined as 50em. For me it was 750px.

xi:
#264 fixes most issues, except for the user dropdown. @vellip do you have any idea how to fix that?

vellip:
We can close that now, right?

PR #264, opened 2017-04-26
mobile first media queries

xi:
In media queries, we sometimes use `min-width` and sometimes `max-width`. This results in issues like #259. This pull request fixes this by using `min-width` wherever possible. i chose `min-width` rather than `max-width` according to the "mobile first" approach.

The notable exception is `.main-header__user .dropdown`. I have no clue how we can fix that case because the mobile and desktop styles are so different.

Media Queries Level 4 (working draft) adds [a new syntax](https://drafts.csswg.org/mediaqueries-4/#mq-range-context) that would fix this issue. But it is not widely available yet.

### DIFF
--- a/meinberlin/assets/scss/components/_header.scss
+++ b/meinberlin/assets/scss/components/_header.scss
@@ -12,32 +12,35 @@
     text-decoration: none;
 
     img {
-        width: 6em;
-        margin-top: -2em;
-        margin-bottom: -2.85em;
+        margin: 0;
+        width: 3em;
+        vertical-align: middle;
 
-        @media screen and (max-width: $breakpoint) {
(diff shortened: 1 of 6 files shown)


meinberlin/assets/scss/components/_header.scss
```diff
@@ -116,8 +116,10 @@
         right: $padding;
     }
 
-    // overwrite dropdown style on mobile
-    @media screen and (max-width: $breakpoint) {
+    // min-width and max-width can't have the same breakpoint
+    @media screen and (max-width: $breakpoint - 0.1em) {
```
xi: A bit hacky, but ok.

PR #265, opened 2017-04-26
allow to translate core strings

xi:
This copies the approach for translating core strings from a4-opin.

It works fine, but I still have some questions. @slomo maybe you can help with that?

- why was fuzzy matching disabled?
- why is a4 included twice (as python module and as npm module)
- what happens if we do not generate an english translation (with msgen)

Apart from that, I noticed that we have some very similar strings (e.g. "cancel" and "Cancel"). It might be a good idea to eliminate some variations.


slomo:
If I remember correctly:

 - Fuzzy matching was disabled because it merge a lot of strings that simply weren't the same. This happend especially on msgmerge after rescanning the project.
 -  A4 has Strings in the node and python package. The python package only includes certain static folders and templates (eg. not the react stuff). The node package includes all javascript. As a result there are some strings that are in both packages, but now all.
 - You have to fill them manually, if you want to use transifex. As far as I remember transifex requires the source language to be fully translated.

Maybe those points would be a good extension to the existing documentation.

vellip:
> You have to fill them manually, if you want to use transifex. As far as I remember transifex requires the source language to be fully translated.

Yep, I think that was the issue.

PR #267, opened 2017-04-26
update npm dependencies

xi:
I did not ready every changelog, but manual testing did not yield any issues.

Note that I did not update leaflet and webpack to new major versions.


vellip:
Updating Webpack might be something worth evaluating.

PR #269, opened 2017-04-26
poll: pass real data to react

xi:
this replaces the sample data with real data from the model. Note that `object` in the poll detail view is currently not defined, so it does not yet work.

### DIFF
--- a/apps/polls/assets/Polls.jsx
+++ b/apps/polls/assets/Polls.jsx
@@ -7,17 +7,20 @@ var Poll = React.createClass({
     // FIXME: example data
     return {
       title: 'Getrennte Eltern: Ist das Wechselmodell die beste Lösung für alle?',
-      options: [{
+      choices: [{
         label: 'Ja',
-        count: 22434
+        count: 22434,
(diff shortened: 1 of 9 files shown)


apps/polls/assets/Polls.jsx
```diff
@@ -6,18 +6,9 @@ var Poll = React.createClass({
...
-        count: 22434
-      }, {
-        label: 'Nein',
-        count: 40062
-      }, {
-        label: 'Vielleicht',
-        count: 17627
-      }],
-      ownVote: null,
+      title: this.props.poll.title,
```
kleingeist: Disclaimer: this question is more to understand react then a review.

Why are title and choices (at least the labels) state values? from the tutorials i would assume they should stay in the props. Unfortunately the choice.count and choice.ownChoice properties are state as they will change on  a vote :(

vellip: As we discussed, see https://shripadk.github.io/react/tips/initial-ajax.html. We will keep this pattern for consistency now though.

PR #270, opened 2017-04-27
Rename Poll to Question and use Poll as a container

kleingeist:


apps/polls/views.py
```diff
@@ -55,8 +35,18 @@ def get_success_url(self):
             'dashboard-project-list',
             kwargs={'organisation_slug': self.organisation.slug, })
 
+    def get_or_create_poll(self):
```
vellip: Why not use https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create ?

kleingeist: i can't assure the creator will be the same, as there can exist multiple admins. different admins can access the poll and it may only be created once for each model.

PR #277, opened 2017-05-02
update blueprint icons

xi:
-   rename agenda-setting to priorization
-   add new icon for agenda-setting

The priorization blueprint is beeing added in #260.


2e2a:
lgtm

PR #279, opened 2017-05-02
re-initialize javascript on embed navigation

xi:
*depends on https://github.com/liqd/adhocracy4/pull/84*

In the embed shell, we do not do regular navigation. This also means that JavaScript code is not automatically run when we enter a new page.

Usually, these three things can happen on page load:

-   register libraries as global variables
-   register global event handlers
-   initialize custom components

In the embed case, the first two should happen once (1), but the third should run after every pageload (2).

This pull requests suggests to do (1) by scanning each loaded page for script references and adding them to `<head>`, but only if they do not exist yet.  (2) is implemented by triggering the custom `a4.embed.ready` event that should be used whenever jQuery's `$(…)` (formerly `$(document).ready(…)`) is used.

It is important to note that dynamically inserted scripts are loaded asynchronously. Also note that `$(…)` is guaranteed to execute exactly once. So when a new page is loaded things happen in the following order:

1.  new script elements are inserted into the DOM
2.  new content is inserted into the DOM
3.  `a4.embed.ready` is triggered
4.  the new scripts are executed
5.  `$(…)` fires

So `a4.embed.ready` is triggered before the new scripts are loaded, only triggering code from old scripts. Conversely, `$(…)` has already fired for older scripts when they were first loaded, so it only triggers code from new scripts. In summary, each function registered to both events should execute exactly once per page load.

I realize that this is not exactly simple. Unfortunately I have not found a better way to do this.


vellip:
Can we make it one function instead of having to call two? Something like `adhocracy.ready([function])`.

xi:
> Can we make it one function instead of having to call two? Something like `adhocracy.ready([function])`.

That would introduce new dependencies between js file. I see the point, but I don't think it would be worth it.

vellip:
Not necessarily new dependencies, adhocracy4 is in global scope anyway.

xi:
The global `adhocracy4` object is not defined in core, so it would introduce dependencies between projects.

slomo:
But if we follow @vellip principal idea and there is only one point for registering those handlers? Like a single event "a4.ready" that can be used to initialise widgets. (This event of course needs to be triggered on document ready somehow/somewhere).

xi:
@vellip, @slomo what do you think about the change in 587033c? Should we also use that in core?

slomo:
:+1:

PR #280, opened 2017-05-02
change embed shell styling

### LINKED ISSUE #271: embed: header & footer styling
The dimensions here need to be adjusted. See status quo (left) and design (right):

embed_dimensions

Design on Zeplin: https://zpl.io/26x4Mz

xi:
fixes #271

### DIFF
--- a/meinberlin/assets/scss/components/_embed-layout.scss
+++ b/meinberlin/assets/scss/components/_embed-layout.scss
@@ -21,8 +21,9 @@ html.embed body {
 .embed-footer {
     position: relative;
     background: $bg-tertiary;
-    padding: 1em;
+    padding: 5px 0;
     text-align: right;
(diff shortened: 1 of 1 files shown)


meinberlin/assets/scss/components/_embed-layout.scss
```diff
@@ -21,8 +21,9 @@ html.embed body {
 .embed-footer {
     position: relative;
     background: $bg-tertiary;
-    padding: 1em;
+    padding: 5px 0;
```
vellip: Just to understand it: Why use 5px here instead of em?

xi: 5px is already used as border on body.

vellip: Ah, right. Okay.

PR #282, opened 2017-05-03
tackle header margin problem

### LINKED ISSUE #82: meinberlin header visually broken on alerts

vellip:
I gave the header a generic `margin-bottom` now, which added some necessary space underneath the header. This allowed me to remove the `margin-top` for the dashboard layout, following the logic that we prefer `margin-bottom` over `margin-top`. 

This fixes #82.

### DIFF
--- a/meinberlin/assets/scss/components/_header.scss
+++ b/meinberlin/assets/scss/components/_header.scss
@@ -201,9 +201,3 @@ $header-overlap: -2.5em;
         margin-top: $header-overlap - $header-margin-bottom;
     }
 }
-
-// compensate for margin-bottom on messages
-.messages + .resource-header,
-.messages + .hero-unit {
-    margin-top: $header-overlap;
-}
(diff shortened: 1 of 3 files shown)


meinberlin/assets/scss/components/_header.scss
```diff
@@ -187,10 +190,14 @@
     }
 }
 
+// compensate for generic margin-bottom on header and move those elements
+// underneath the "dialograhmen"
+$header-overlap: -2.5em;
```
xi: This value is predetermined by the dialograhmen image, so it would be better to define it in `px`. We could either use `calc` to combine it with `$messages-margin-bottom` or define `$messages-margin-bottom` in `px` as well. Or not. I guess this is not too important.

PR #273, opened 2017-04-27
Add react based poll management

kleingeist:
this PR is still missing
* Layout for the management form

which will be worked on in a different task/pr

PR #283, opened 2017-05-03
Allow votes from the user frontend

kleingeist:
This PR includes the changes from #273 
API changes depend on liqd/adhocracy4#89

It is missing
* Changes to the api.js in core
* logic for the AJAX request
  * what to show during the request
  * how to handle/show errors/success
* Layout for the management form


2e2a:
api and serializer lgtm

apps/polls/assets/QuestionForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+      <section>
+        <div className="commenting">
+          <div className="commenting__content">
+            <div className="form-group">
+              <label
+                htmlFor={'id_questions-' + this.props.key + '-name'}>
+                {django.gettext('Question:')}
+              </label>
+              <input
+                className="form-control"
```
vellip: `.form-control` is of no importance anymore, we removed it. I know, theoretically the same goes for documents.

apps/polls/assets/PollManagement.jsx
```diff
@@ -0,0 +1,224 @@
...
+
+    let data = {
+      questions: this.state.questions
+    }
+
+    let promise = api.poll.change(data, this.props.poll.id)
+    promise
+      .done(function (data) {
+        this.setState({
+          successMessage: django.gettext('The poll has been updated.')
```
vellip: You'll want to reset previous errors here.

apps/polls/assets/PollManagement.jsx
```diff
@@ -0,0 +1,224 @@
...
+
+        setTimeout(function () {
+          this.setState({
+            successMessage: ''
+          })
+        }.bind(this), 1500)
+      }.bind(this))
+      .fail(function (xhr, status, err) {
+        this.setState({
+          errors: xhr.responseJSON.questions || []
```
vellip: Why a fallback? If there are no errors to set (which `errors: []` would mean) you don't want to call `setState`.

apps/polls/assets/PollManagement.jsx
```diff
@@ -0,0 +1,224 @@
...
+        this.setState({
+          successMessage: django.gettext('The poll has been updated.')
+        })
+
+        setTimeout(function () {
+          this.setState({
+            successMessage: ''
+          })
+        }.bind(this), 1500)
+      }.bind(this))
```
vellip: Consider using [fat arrow functions](https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Functions#Arrow_functions) to get rid of all the `.bind`s

apps/polls/assets/Question.jsx
```diff
@@ -0,0 +1,169 @@
...
+  handleSubmit: function (event) {
+    event.preventDefault()
+
+    let newChoice = this.state.selectedChoice
+
+    let submitData = {
+      choice: this.props.question.choices[newChoice].id
+    }
+
+    return api.poll.vote(submitData, this.props.question.id)
```
vellip: No need to return the promise.

apps/polls/migrations/0002_auto_20170503_1524.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
kleingeist: i'll leave the generated title this time but will name it manually next time. 
maybe we could also squash them when the feature is ready

apps/polls/assets/PollManagement.jsx
```diff
@@ -0,0 +1,224 @@
...
+  */
+
+  getNextQuestionKey: function () {
+    /** Get an artifical key for non-commited questions.
+     *
+     *  Prefix to prevent collisions with real database keys;
+     */
+    var questionKey = 'local_' + (this.state.maxQuestionKey + 1)
+    this.setState({maxQuestionKey: this.state.maxQuestionKey + 1})
+    return questionKey
```
kleingeist: good point, i changed it to be a class variable. the changes should be reflected in documents, too

PR #284, opened 2017-05-03
integrate follows (wip)

xi:
This is a companion PR to https://github.com/liqd/adhocracy4/pull/87. It may be useful for reference or even to actaully integrate the feature in meinberlin.

Styling is missing. In addition, the template tag may be moved to a different template.


kleingeist:
lgtm. styling will be added as a  new task

PR #286, opened 2017-05-04
Enable commenting on polls

kleingeist:


vellip:
Something I noticed only after the initial merge, but I still want to voice: Do we need a separate phase for commenting polls? I think it'd make more sense to allow comments all the time and make it just one phase.

kleingeist:
the phases had been defined in the user story, but we just discussed this with the PM and decided to drop the second, comment only, phase. i will prepare another PR for this

apps/polls/templates/meinberlin_polls/poll_detail.html
```diff
@@ -10,4 +10,8 @@
             {% endfor %}
         </div>
     </div>
+
+    <div class="l-wrapper">
+        {% react_comments object %}
+    </div>
```
xi: This can be in the same `l-wrapper`.

PR #291, opened 2017-05-05
show image in project tiles

vellip:
screen shot 2017-05-05 at 17 23 38

### DIFF
--- a/apps/projects/templates/meinberlin_projects/includes/project_list_tile.html
+++ b/apps/projects/templates/meinberlin_projects/includes/project_list_tile.html
@@ -3,8 +3,9 @@
 <div class="tile">
     <div class="tile__head">
         <a tabindex="-1" class="tile__image tile__image--shadowed"
+           {% if project.image %}style="background-image: url({{ project.image|thumbnail_url:'project_thumbnail' }})"{% endif %}
            href="{% extproject_url project 'project-detail' %}" target="{{ project|is_external|yesno:"_blank,_self" }}">
-            <img src="http://placehold.it/285x170" alt="">
+            <span class="sr-only">{{ project.name }}</span>
         </a>
         {% if project|is_external %}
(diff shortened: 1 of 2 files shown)


meinberlin/assets/scss/components/_tile.scss
```diff
@@ -7,16 +7,14 @@
 
 .tile__head {
     position: relative;
-    min-height: 6em;
 }
 
 .tile__image {
     display: block;
+    height: 11.8em;
```
xi: Can we define this as `min-height`? It should do the same by default, but is more robust.

apps/projects/templates/meinberlin_projects/includes/project_list_tile.html
```diff
@@ -3,8 +3,9 @@
 <div class="tile">
     <div class="tile__head">
         <a tabindex="-1" class="tile__image tile__image--shadowed"
+           {% if project.image %}style="background-image: url({{ project.image|thumbnail_url:'project_thumbnail' }})"{% endif %}
            href="{% extproject_url project 'project-detail' %}" target="{{ project|is_external|yesno:"_blank,_self" }}">
-            <img src="http://placehold.it/285x170" alt="">
+            <span class="sr-only">{{ project.name }}</span>
```
xi: `project.name` is already mentioned in `tile__title`. I would add [`role="presentation"`](https://www.w3.org/TR/wai-aria/roles#presentation) to the whole link instead.

vellip: I can do that, however W3C says: 
> If an element with a role of presentation is focusable, or otherwise interactive, user agents must ignore the normal effect of the role and expose the element with implicit native semantics, in order to ensure that the element is both understandable and operable.

That basically means the role will get ignored, right?

xi: You are right. In that case I would opt for `aria-hidden="true"`. The image does not add anything other than the image itself, so it can just as well be ignored.

PR #293, opened 2017-05-08
add transifex config

xi:
it is not completely clear yet whether we qualify for the free plan, so this is on hold.


xi:
We will not use transifex for now

PR #299, opened 2017-05-08
improve priotization management styling

xi:


meinberlin/assets/scss/components/_lr_bar.scss
```diff
@@ -7,8 +7,11 @@
 
 .lr-bar__left {
     display: inline-block;
+    margin: 0;
```
vellip: This seems to solve a specific problem (headings with margin-top), but I am not sure if it's generally applicable, since it also only works if you put that class directly on the heading. I'd leave it in for now but we should be aware of that.

apps/topicprio/templates/meinberlin_topicprio/includes/topic_mgmt_list_item.html
```diff
@@ -1,21 +1,20 @@
...
+            </h3>
+        </div>
+        <div class="lr-bar__right">
             <a href="{% url 'meinberlin_topicprio:topic-mgmt-delete' slug=object.slug %}"
                class="button button--light">
                 <i class="fa fa-remove" aria-hidden="true"></i>
                 {% trans 'Delete'%}
             </a>
-        </span>
+        </div>
```
xi: I also thought about this, but decided against it because `lr-bar__right` moves down on if `lr-bar__left` has more than one line. Of course we could leave this as it is and change the `lr-bar` implementation instead.

vellip: I think it's actually fine in this case. Otherwise we'd have to give the left side a max-width, which doesn't sound like something we want.

PR #302, opened 2017-05-08
improve poll management styling

xi:
still to do:

-   choice input and delete button should be in a single line
-   choice field should not have a (visible) label

### DIFF
--- a/apps/polls/assets/QuestionForm.jsx
+++ b/apps/polls/assets/QuestionForm.jsx
@@ -48,71 +48,69 @@ let QuestionForm = React.createClass({
 
   render: function () {
     return (
-      <section>
-        <div className="commenting">
-          <div className="commenting__content">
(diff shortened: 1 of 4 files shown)


apps/polls/assets/QuestionForm.jsx
```diff
@@ -56,10 +56,9 @@ let QuestionForm = React.createClass({
                 htmlFor={'id_questions-' + this.props.key + '-name'}>
                 {django.gettext('Question:')}
               </label>
-              <input
+              <textarea
```
kleingeist: Currently the Question.label field is declared as a CharField with a max length of 255. If longer text is required we should change it in the database, too

vellip: No, the max length is more than generous. It was just changed to a textarea because of the design.

PR #304, opened 2017-05-09
Add Bplan projects to the dashboard and a first statement formular draft

kleingeist:
Mails are still missing and it is unclear if the formular can be embedded like this


2e2a:
lgtm

PR #308, opened 2017-05-09
define button--attached-top

### LINKED ISSUE #288: rounded corners of feedback button
The feedback button is not supposed to have all corners rounded, just the two on the left hand side. 

bildschirmfoto 2017-05-05 um 16 24 59

xi:
fixes #288

### DIFF
--- a/meinberlin/assets/scss/components/_button.scss
+++ b/meinberlin/assets/scss/components/_button.scss
@@ -89,6 +89,11 @@
     margin-bottom: $spacer;
 }
 
+.button--attached-top {
+    border-top-left-radius: 0;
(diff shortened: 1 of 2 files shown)


meinberlin/assets/scss/components/_button.scss
```diff
@@ -89,6 +89,11 @@
     margin-bottom: $spacer;
 }
 
+.button--attached-top {
```
vellip: why --attached-**top**?

xi: from the button's point of view, it is attached to the top. The container it is in is turned by 90deg.

PR #317, opened 2017-05-09
increase default font size

xi:


vellip:
This will make a lot of stuff bigger now, right? Did you check that?

xi:
Yes, that is the intended result.

PR #324, opened 2017-05-10
Mails: Add content_html block with a fallback that converts linebreaks

kleingeist:
This introduces a new content_html block to the mail templates which can
be used to explictly define HTML emails.
If this block is not defined a fallback is used, that captures the
output of the standard text content and converts linebreaks to html.
For this feature a new dependency to django-capture-tag is introduced.


2e2a:
nice

PR #325, opened 2017-05-10
Fix poll voting by replacing the AllowPutAsCreateMixin

kleingeist:
Since this view has a mismatch between its pk (which is
related to questions) and its serializer (which is related to votes)
the AllowPUTAsCreateMixin won't work correctly, as it sets the pk
on the serializer object which obviously results in faulty database
operations.


2e2a:
works

PR #327, opened 2017-05-10
use django date format instead of hardcoded german format

vellip:


xi:
Fixes the issue. I hope there won't be more.

PR #328, opened 2017-05-10
show total count in preliminary poll results

xi:


apps/polls/assets/Question.jsx
```diff
@@ -110,13 +110,18 @@ var Question = React.createClass({
     let max = Math.max.apply(null, counts)
 
     let footer
+    let totalString = '' + total + ' ' + django.ngettext('vote', 'votes', total)
```
vellip: `'' + [...]` Why the empty string?

vellip: You _could_ also use es2015 template strings.

PR #329, opened 2017-05-10
link to login on poll without authentication

xi:
this can probably be refactored to be more generic, but I went with a simple solution for now.


xi:
b98c9bd is unrelated, but I pushed it here to avoid merge conflicts.

PR #331, opened 2017-05-10
update body colour

vellip:
as in design


xi:
I tested some contrasts and this should be fine.

PR #337, opened 2017-05-10
style poll detail

vellip:


apps/polls/assets/Question.jsx
```diff
@@ -130,7 +130,7 @@ var Question = React.createClass({
             { django.gettext('Vote') }
           </button>
           &nbsp;
-          <button type="button" className="button button--light" onClick={this.toggleShowResult}>
+          <button type="button" className="link" onClick={this.toggleShowResult}>
```
xi: Before this change, questions had the same height no matter which view was selected. I think this was an important property.

PR #339, opened 2017-05-11
avoid voting for ended projects

vellip:
@kleingeist this still is possible via API, I presume. I wasn't too sure where to do that. Either put it in an issue or fix if you find time.


kleingeist:
i will modify the backend, too

kleingeist:
@vellip actually a regular user will not be allowed to vote after the phase. but project admins and site admis are always allowed to edit items (and thus votes)

i would merge this PR after the demonstration

PR #344, opened 2017-05-11
style follow

vellip:


kleingeist:
lgtm
using meinberlin specific templates/classes should be done in the next sprint

meinberlin/assets/scss/_shame.scss
```diff
@@ -20,6 +20,23 @@
...
+    &:first-child {
+        border-bottom-right-radius: 0;
+        border-top-right-radius: 0;
+    }
+
+    &:last-child {
+        border-bottom-left-radius: 0;
+        border-top-left-radius: 0;
+    }
+}
```
xi: We already have `.button-group`. Maybe it is better to extend that here. (And maybe we should just rename our `.button` to `.btn` to avoid this kind of compatibility issues)

PR #348, opened 2017-05-11
style radio buttons

vellip:


2e2a:
Looks nice.

But there is a problem with choices spanning over multiple lines:

radio_multiline

vellip:
Good catch!

xi:
This has some overlap with `.form-check`

xi:
There is no focus indication with these radio boxes.

xi:
b251c4c breaks the padding in the result view.

PR #349, opened 2017-05-11
Add poll 404 page

kleingeist:
If the modules Poll has not been created yet a custom 404 page with a condititonal link to the dashboard is shown

### DIFF
--- a/apps/polls/templates/meinberlin_polls/poll_404.html
+++ b/apps/polls/templates/meinberlin_polls/poll_404.html
@@ -1,0 +1,17 @@
+{% extends "base_errors.html" %}
+{% load i18n rules item_tags %}
+
+{% block error_status %}404{% endblock %}
+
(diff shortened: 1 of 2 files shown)


apps/polls/templates/meinberlin_polls/poll_404.html
```diff
@@ -4,14 +4,17 @@
 {% block error_status %}404{% endblock %}
 
 {% block error_description %}
-    {% trans 'Participation is not possible at the moment.' %}
+    <p>
```
kleingeist: unfortunately this results in invalid html, as https://github.com/liqd/a4-meinberlin/blob/dd5b43723b55d5a7c2f3b6e8398ef4968858b3d0/meinberlin/templates/base_errors.html already encapsulates the error_description in a `<p>` and paragraphs are not allowed nested in html

PR #350, opened 2017-05-11
Validate bplan image downloaded in API serializer

### LINKED ISSUE #346: Bplan: Image size restrictions are not checked when imported with the api
The project form limits image to a minimum size. Those restrictions are not checked when importing images with the api.
If a smaller image is already imported, the project can not be edited in the dashboard.

2e2a:
Fixes #346

### DIFF
--- a/apps/bplan/serializers.py
+++ b/apps/bplan/serializers.py
@@ -137,9 +137,9 @@ class BplanSerializer(serializers.ModelSerializer):
 
         try:
             self._validate_image(file_name)
-        except ValidationError:
+        except ValidationError as e:
             self._image_storage.delete(file_name)
-            raise
+            raise serializers.ValidationError(e)
 
         return file_name
 
--- a/apps/bplan/serializers.py
+++ b/apps/bplan/serializers.py
@@ -1,4 +1,5 @@
-import os
+import datetime
+import posixpath
 from urllib import request
 from urllib.parse import urlparse
 
@@ -15,6 +16,7 @@ from rest_framework import serializers
 from adhocracy4.images.validators import validate_image
 from adhocracy4.modules import models as module_models
 from adhocracy4.phases import models as phase_models
+from adhocracy4.projects import models as project_models
 
 from .models import Bplan
 from .phases import StatementPhase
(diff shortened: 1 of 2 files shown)


apps/bplan/serializers.py
```diff
@@ -104,12 +102,35 @@ def _get_absolute_url(self, bplan):
...
+            file_name = os.path.join(settings.MEDIA_ROOT, file_path)
+            file_dir = os.path.dirname(file_name)
+            os.makedirs(file_dir, exist_ok=True)
+
+            request.urlretrieve(url, file_name)
+        except:
+            self._remove_image_if_exists(file_name)
+            raise serializers.ValidationError(
+                'Failed to download image {}'.format(url))
+        self._validate_image(file_name)
```
kleingeist: why are you validating the image outside of the try block?
i suggest to remove the try/except block from the _validate_image function and just move the call in the try block

apps/bplan/serializers.py
```diff
@@ -104,12 +102,35 @@ def _get_absolute_url(self, bplan):
...
+
+    def _validate_image(self, file_name):
+        image_file = open(file_name, "rb")
+        image = ImageFile(image_file, file_name)
+        config = settings.IMAGE_ALIASES.get('*', {})
+        config.update(settings.IMAGE_ALIASES['heroimage'])
+        try:
+            validate_image(image, **config)
+        except ValidationError as e:
+            self._remove_image_if_exists(file_name)
```
kleingeist: i don't like the idea, that the validator tries to delete the validated image. see above

kleingeist:
the image is not updated on PATCH

kleingeist:
also start and enddate are not updated

apps/bplan/serializers.py
2e2a: To differentiate the error cases, without checking the exceptions. I'll change it.

2e2a:
Good point,  will fix that.

kleingeist:
@2e2a i just did commit the changes to use the django storage backend. could you please review the changes and merge them if they are okay

apps/bplan/serializers.py
```diff
@@ -1,4 +1,5 @@
-import os
+import datetime
+import posixpath
```
xi: Out of interest: Why the switch from `os.path` to `posixpath`?

kleingeist: i did copy the _generate_image_filename logic from django where posixpath was used. i am not sure why the prefered it over os.path

PR #369, opened 2017-05-12
smaller email style adjustments

vellip:
Just a little cleanup and text alignment unification


kleingeist:
lgtm.
still i think we should reconsider the overall email template. i don't get why we are limiting us to that small width

PR #379, opened 2017-05-12
input action element

vellip:
To have an input and an action next to each other.

screen shot 2017-05-12 at 18 33 10


xi:
@damusp also noted that the choice fields should not fill the full available width.

xi:
screenshot from 2017-05-15 18-17-14

I pushed a version that uses light buttons. Feel free to revert if you do not like it.

damusp:
but i like it!

vellip:
The width shouldn't be a part of this element. Your changes are fine with me.

PR #384, opened 2017-05-15
Add pagination to dashboard project list.

### LINKED ISSUE #311: Pagination in dashboard project list
While the list gets cut after x items already, there is no pagination to access the other pages.

kleingeist:
Fixes #311


kleingeist:
This just adds the pagination.
But i just realized that the breadcrumb navigation "<- back to dashboard" always links to the first page. i think it would be nice to have the correct reference back to the page you had been browsing before. but that is another issue

PR #389, opened 2017-05-15
add missing poll migration

xi:
fixup to https://github.com/liqd/a4-meinberlin/pull/382


kleingeist:
@xi thank you for adding it. i tend to forget to make migrations if i change meta fields on the database that won't reflect in changes to the actual db

PR #390, opened 2017-05-15
limit form width

xi:
edit/create form and delete confirm should have same width as detail

### DIFF
--- a/apps/topicprio/templates/meinberlin_topicprio/topic_mgmt_confirm_delete.html
+++ b/apps/topicprio/templates/meinberlin_topicprio/topic_mgmt_confirm_delete.html
@@ -5,17 +5,15 @@
 
 {% block dashboard_content %}
 <div class="l-wrapper">
-    <div class="l-center-6">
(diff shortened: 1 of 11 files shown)


apps/topicprio/templates/meinberlin_topicprio/topic_mgmt_confirm_delete.html
```diff
@@ -2,18 +2,20 @@
 {% load i18n item_tags %}
 
 {% block title %}{% trans 'Confirm Topic Deletion' %}{% endblock %}
-{% block dashboard_content %}
 
+{% block dashboard_content %}
 <div class="l-wrapper">
-    <h1>{% trans 'Are you sure you want to delete the topic?' %} </h1>
+    <div class="l-center-6">
```
kleingeist: The topicprio deletion confirmation is only shown in the dashboard, where the full width may be used.

xi: Good catch!

PR #393, opened 2017-05-16
A3 user import script

2e2a:
Copy from https://github.com/liqd/a4-opin


xi:
Do we want to merge this? We will probably only run it a single time, so I do not see much value in having it in master.

2e2a:
No, we can close this pull request, but should keep the branch.

PR #395, opened 2017-05-16
fix broken variable name in webpack

### LINKED ISSUE #334: Webpack generates broken variable name
With the setup since #278, webpack generates the following code:

```
var leaflet.draw = …
```

This breaks some mapping functionality.

Possible fix: rename bundle to leafletDraw.

xi:
fixes #334

Before this, webpack would generate the following (invalid) code:

    var leaflet.draw = …

Now, the it will be more like this:

    this['leaflet.draw'] = …

Since this code is executed in the global scope (and this therefore refers to `window`), it should result in the same thing, just with valid syntax.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -38,7 +38,7 @@ module.exports = {
   },
   devtool: 'eval',
   output: {
-    libraryTarget: 'var',
+    libraryTarget: 'this',
     library: '[name]',
     path: './meinberlin/static/',
     publicPath: '/static/',


webpack.config.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
   },
   devtool: 'eval',
   output: {
-    libraryTarget: 'var',
+    libraryTarget: 'this',
     library: '[name]',
     path: './meinberlin/static/',
```
kleingeist: just a side note: https://webpack.github.io/docs/configuration.html#output-path states this should be "The output directory as an *absolute path* (required)."

kleingeist:
i deem this a suitable solution for our problem. as the libraryTarget is only relevant for the output we are using on our site, we can be certain `this` points to the `window`

PR #396, opened 2017-05-16
tweak phase info styling

xi:


kleingeist:
looks better then the "from" ... "to" solution and is more tidy

PR #398, opened 2017-05-16
Webpack: Use absolute output path

kleingeist:
The docu states, that the output directory has to be an absolute path.
see https://webpack.github.io/docs/configuration.html#output-path


xi:
LGTM.

PR #399, opened 2017-05-16
add edit button on project detail

### LINKED ISSUE #374: Edit participation project button in UI?
Did we plan an edit participation project button in the UI? This way it would be convenient to arrive directly in the dashboard of the respective process.

xi:
fixes #374


kleingeist:
i don't like the button that much, as it is very dominant. but as this has been discussed with @damusp and there are plans to restructure the project view i will merge

PR #405, opened 2017-05-16
A3 project import

2e2a:
Imports:
- [x] polls
- [x] extern
- [x] idea collections
- [x] commenting text
- [x] kiezkasse
- [x] bürgerhaushalt
- [x] bplan

TODO:
- [x] Make markdown submissions readable
- [ ] Badges?


apps/projects/management/commands/util.py
```diff
@@ -0,0 +1,101 @@
...
+
+def a3_get_sheet_field(resource_url, headers, sheet, field):
+    res = requests.get(resource_url, headers=headers)
+    if res.status_code != requests.codes.ok:
+        raise CommandError('Request failed for URL: {}'.format(
+            resource_url)
+        )
+    data = res.json()
+    sheet_field_value = data['data'][sheet][field]
+    return sheet_field_value
```
xi: If this gets too slow, a simple optimization could be to define `a3_get_resource()` and decorate it with `functools.lru_cache()`.

apps/projects/management/commands/util.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+
+    module = Module.objects.create(
+        name=project.slug + '_module',
+        weight=1,
+        project=project,
+    )
+
+    phase_start = start_date
+    phase_duration = (end_date - start_date) / len(phase_contents)
```
xi: which type are `start_date` and `and_date`? Timestamp or something from `datetime`?

apps/projects/management/commands/util.py
```diff
@@ -0,0 +1,101 @@
...
+
+def a3_login(url, headers, username, password):
+    login_url = url + 'login_username'
+    res = requests.post(
+        login_url,
+        json={'name': username, 'password': password}
+    )
+    if res.status_code != requests.codes.ok:
+        raise CommandError('API user authentication failed.')
+    headers['X-User-Token'] = res.json()['user_token']
```
xi: I think it would improve readability a great deal if this returned the token instead of setting it directly.

apps/projects/management/commands/import_a3_polls.py
```diff
@@ -0,0 +1,92 @@
...
+                continue
+
+            orga_name = util.a3_get_sheet_field(
+                orga_path, headers,
+                'adhocracy_core.sheets.name.IName', 'name'
+            )
+            orga, created = Organisation.objects.get_or_create(name=orga_name)
+
+            self._import_polls(headers, poll_paths, orga, default_creator)
+
```
xi: Can we move everything abouve this point to `util`?

apps/projects/management/commands/import_a3_polls.py
```diff
@@ -0,0 +1,92 @@
...
+                    'adhocracy_core.sheets.title.ITitle', 'title')
+
+                project, module = util.create_project(
+                    organisation,
+                    'poll-tbd',
+                    'desc-tbd',
+                    'info-tbd',
+                    creation_date,
+                    modification_date,
+                    _('Poll'),
```
xi: The translation is done on import time, so I do not see the point. We can just hardcode the desired language.

apps/projects/management/commands/import_a3_polls.py
2e2a: see new commits

apps/projects/management/commands/util.py
xi: I didn't know you could divide timedeltas by integers. Nice!

xi:
Something is still off. If I simply compare https://mein.berlin.de/w/prozesse/sind-sie-bereit-sich-klimaneutral-zu-verhalten/ and the imported equivalent:

- votes: 50 -> 51
- comments: 37 -> 24 + 12 replies

xi:
I found that many texts do not have HTML tags. Maybe markdown conversion is missing in some places?

Example: https://meinberlin-stage.liqd.net/mapideas/cafe-evelin/

2e2a:
no need to merge this onetime thing. Using the branch to import.

PR #407, opened 2017-05-17
Add point_label field to mapideas

kleingeist:
The point label is an optional text field that can be used to describe the point selected on the map. for example "Alexanderplatz", "Rollbergstr. 15", "Liqd Büro" etc.


xi:
LGTM

PR #409, opened 2017-05-17
Add kiezkasse project

kleingeist:
was depending on liqd/adhocracy4#97 and #407 

Unfortunately this PR includes several changes. But the are encapsulated in the individual commits

* removing the category related field in accordance to  liqd/adhocracy4#97
* use an explicit related name for item pointers on abstract ideas to prevent clashes
* add kiezkasse as a new app


2e2a:
lgtm

PR #426, opened 2017-05-18
Refactor moderator feedback to act as an abstract model.

kleingeist:
A new abstract Moderateable model is introduces that add the moderator
statement and moderator feedback to an item.
The actual model may overwrite the default choices by setting
`moderator_feedback_choices`.
To generalize the forms a factory is required, as the actual item type
has to be known to build correct forms.


2e2a:
Nice

Issue #289, opened 2017-05-05
Forgot Password? // Log In (Primary and secondary brand colour)


damusp:
I accidentally click on the magenta button because my brain expects it to be the button that logs me in. I think this is because it is positioned on the very left while the log in comes in second place. Also its color is much stronger. 

Suggestion:
Log In button comes first
Password Forgotten follows below, maybe styled as a link instead of a button

<img width="510" alt="bildschirmfoto 2017-05-05 um 16 27 24" src="https://cloud.githubusercontent.com/assets/22769111/25749753/dee49ea6-31af-11e7-885d-c46b8fbf55aa.png">

vellip:
Interestingly, this might be an issue with our colour variables:
```scss
$brand-primary: #253276;
$brand-secondary: #cd0055;
```

xi:
Related to #124

PR #428, opened 2017-05-22
colour variables

vellip:
This takes care of #289 by switching brand-primary and brand-secondary. These colours and their usage should be unified with this.


xi:
LGTM.

PR #430, opened 2017-05-22
Use allauth routes and overwrite templates

kleingeist:
Followup to #429

Use the allauth default routes and overwrite the allauth templates
instead of using a new view. The templates relevant for the account-
dashboard are inheriting from the account_dashboard.html to display the
navigation.

### DIFF
--- a/meinberlin/settings/base.py
+++ b/meinberlin/settings/base.py
@@ -229,7 +229,6 @@ ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION = True
 ACCOUNT_LOGIN_ON_PASSWORD_RESET = True
 
 LOGIN_URL = 'account_login'
-LOGOUT_URL = 'account_logout'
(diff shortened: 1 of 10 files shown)


kleingeist:
@vellip do we still need the rules in shame.scss 
https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/assets/scss/_shame.scss#L46

meinberlin/settings/base.py
```diff
@@ -228,6 +228,8 @@
 ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION = True
 ACCOUNT_LOGIN_ON_PASSWORD_RESET = True
 
+LOGIN_URL = 'account_login'
+LOGOUT_URL = 'account_logout'
```
vellip: There is no LOGOUT_URL setting and it's nowhere used. https://tree.taiga.io/project/ggr-meinberlinteam/us/1009?no-milestone=1

PR #437, opened 2017-05-22
embed: only handle ajax errors from navigation

### LINKED ISSUE #420: embed shell registers too general ajax error handlers
the embed shell registers a handler to `$(document).ajaxError(…)`. This handler will be triggered on all ajax errors, not just the ones that result from embed navigation.

An example is an expected 404 response on `/api/follow` if the user has never followed a resource.

xi:
fixes #420

@vellip: Was there any reason to use the general handler?

### DIFF
--- a/meinberlin/static/embed.js
+++ b/meinberlin/static/embed.js
@@ -70,6 +70,24 @@ $(document).ready(function () {
     $top.focus()
   }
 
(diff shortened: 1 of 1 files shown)


vellip:
I don't think so. I just thought it'd be desirable to show any of these errors, but obviously not for follows.

PR #441, opened 2017-05-22
style ellipsis icons

xi:


vellip:
Following @damusp's design logic, we'd have `font-weight: bold` for normal buttons and `font-weight: normal` for small buttons.

vellip:
Also, the three dots aren't centered vertically as of now. A `vertical-align: middle` on the `i` seems to have the wanted effect.

vellip:
Looks better!

PR #442, opened 2017-05-22
Initial moderatorfeedback migration

2e2a:


kleingeist:
if you already have a db where the model has been created as a dependency use the following command to mark it as already run
`./manage.py migrate --fake meinberlin_moderatorfeedback 0001_initial`

PR #444, opened 2017-05-22
Add allauth emails and make all emails translateable

kleingeist:
using translations in mails makes them more messy but less likely to forget them before release


2e2a:
lgtm

PR #449, opened 2017-05-23
consistent list item labels

### LINKED ISSUE #439: categories shown twice on map idea list item
https://meinberlin-dev.liqd.net/projects/geographische-ideensammlung-2/?mode=list

xi:
fixes #439

### DIFF
--- a/apps/ideas/templates/meinberlin_ideas/includes/idea_list_item.html
+++ b/apps/ideas/templates/meinberlin_ideas/includes/idea_list_item.html
@@ -27,13 +27,13 @@
         {{ object.name }}
         </a>
     </h3>
-    <div class="list-item__labels">
-        {% if object.category %}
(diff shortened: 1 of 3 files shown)


apps/ideas/templates/meinberlin_ideas/includes/idea_list_item.html
```diff
@@ -27,11 +27,13 @@ <h3 class="list-item__title">
         {{ object.name }}
         </a>
     </h3>
-    {% if object.category %}
-        <div class="list-item__labels">
-            <span class="label label--big">{{ object.category }}</span>
-        </div>
-    {% endif %}
+    <div class="list-item__labels">
+        {% if object.category %}
```
vellip: Can we wrap the if around the div? Otherwise an empty div might create some unneeded margin.

PR #450, opened 2017-05-23
prevent datepicker icon from being cut

vellip:


xi:
I cannot reproduce the issue. With this change, there is the icon is not centered (because there is a `&nbsp;` atfer it).

vellip:
Alternatively, we can increase the span's width and height just enough for the image to fit in. Something like 19px x 20px.
screen shot 2017-05-23 at 14 23 25

xi:
Ah, I see the issue now. The size of the icon is defined in `rem`, the size of the containing span is defined in `px`.

PR #451, opened 2017-05-23
format placeholder of datepicker to respective language

vellip:


xi:
I cannot reproduce the issue. With this change, there is the icon is not centered (because there is a `&nbsp;` atfer it).

xi:
LGTM.

PR #452, opened 2017-05-23
use template filter to get translated blueprint name

xi:
We currently store `blueprint.title` as `project.typ` when creating a
project. Unfortunately, this will write the translation that is used for
the creator into the database.

This pull request adds a filter that maps a blueprint key to the
translated title. This way, we can store the key into the database and
get the ranslation in the template where it is displayed.

### DIFF
--- a/apps/dashboard/blueprints.py
+++ b/apps/dashboard/blueprints.py
@@ -169,4 +169,8 @@ blueprints = [
 class BlueprintMixin:
     @property
     def blueprint(self):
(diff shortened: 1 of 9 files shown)


kleingeist:
I did update the dashboard project creation logic to use the blueprint key. 
but we have to be aware, that sorting for existing projects will break when we merge this.
is this relevant for the dev server where we already have some content?

2e2a:
lgtm. I will fix this manually on the dev server.

PR #453, opened 2017-05-23
2017 05 pv filterbar overlap

### LINKED ISSUE #166: Fix filter bar overlap

vellip:
fixes #166 

This only does the top portion of it, I think the pagination is fine where it is for now. `margin-bottom` as well as `position: relative` work fine here, I don't think the additional space coming with `position: relative` is a problem here.

### DIFF
--- a/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
+++ b/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
@@ -21,12 +21,14 @@
 {% endblock %}
 
 {% block phase_content %}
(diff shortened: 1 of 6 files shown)


xi:
Nice! Just one detail: I think we can define `filter-bar__top-overlap` as a modifier on `filter-bar`. This way we could save one level of nesting.

Issue #454, opened 2017-05-23
Budget fields do not specify currency


xi:
We have some project types which require people to insert a budget. This is defined as a number without any currency information. I personally will probably only need €, but we should allow to configure this for propoer localization support.

xi:
@Janaba should this be a global setting or per project?

PR #455, opened 2017-05-23
Show currency on budget fields

### LINKED ISSUE #251: Proposal forms are missing the currency symbol
the input field for the budget is missing a currency symbol when a new proposal is entered or updated

screenshot from 2017-05-02 13-54-40

xi:
fixes #251

See https://v4-alpha.getbootstrap.com/components/input-group/ for inspiration.

See #454 for about the hardcoded `€`.


vellip:
Looks nice. I'd prefer if we only have a 1px vertical border in those cases instead of 2px. Could you still add that?

vellip:
The borders are still doubled on the poll contents view, shall we leave it like that for now or did you just not check that?

xi:
I created #469 for tracking refactoring related to `.input-addon`. I would like to get this merged and fix the double border issue later. Fixing it could require a bigger refactoring.

PR #456, opened 2017-05-23
Filter wagtail project choice options

### LINKED ISSUE #422: Filter archived and draft projects from the wagtail frontpage selection element

xi:
fixes #422

While working on this, I found a bug in wagtail: When leaving the project choice widget empty, you will get a 500 error on submit. I was not yet able to reproduce that issue in an isolated unit test though.

This issue is relevant because with this pull request a project that was once available for selection may become unavailable. Users that try to edit the frontpage after that will run into the described issue.

### DIFF
--- a/apps/cms/models.py
+++ b/apps/cms/models.py
@@ -49,6 +49,12 @@ class ProjectSelectionBlock(blocks.ChooserBlock):
             return value.pk
         return value
 
+    def value_from_form(self, value):
+        # if project became unavailable (unpublished), selection will become an
+        # empty string and cause a server error on save, so we give a fallback
+        value = value or None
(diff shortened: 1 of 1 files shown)


apps/cms/models.py
```diff
@@ -49,6 +49,12 @@ def value_for_form(self, value):
             return value.pk
         return value
 
+    def value_from_form(self, value):
+        # if project became unavailable (unpublished), selection will become an
+        # empty string and cause a server error on save, so we give a fallback
+        value = value or None
+        return super().value_from_form(value)
```
xi: This is a wagtail bug. We should add a note that this can be removed once the bug is fixed.

PR #458, opened 2017-05-24
simple user profile

xi:
This adds a simple user profile. Its main purpose is to fix the broken links on comments.

The information that is shown is username, join date and followed projects. Is this ok or should we keep some of that information private?

The implementation is based on a4-opin. I am not completely sure whether the selection of projects is useful in our context. Maybe it would also be good to ignore archived projects to not end up with hundrets o projects over time.

### DIFF
--- a/apps/users/templates/meinberlin_users/user_detail.html
+++ b/apps/users/templates/meinberlin_users/user_detail.html
@@ -10,6 +10,7 @@
                 {{ object.username }}
             </h1>
 
+            <h2>{% trans 'User participation' %}</h2>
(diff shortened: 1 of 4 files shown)


kleingeist:
lgtm
@xi do you want to review my changes and then merge this PR?

vellip:
Just a minor remark, maybe make the `h2` something like "Projects {{user}} participated in".

vellip:
Guys, Travis seems to keep failing here.

xi:
I think it fails because the PR was merged before travis could start. My bad.

PR #472, opened 2017-05-29
Enable ServiceKonto Login

kleingeist:
This enables basic servicekonto support.
There is still some functionality missing: 

- [x] show linked account in account dashboard
- [x] enable to link existing accounts from the dashboard
- [x] overwrite social templates (f.ex. the form shown if an email is already registered)
- [x] overwrite social email templates (if there are some)
- [x] tests
- [x] fix isort
- [x] hide navigation if social account is disabled


kleingeist:
Currently users coming from Service-Konto are logged in (setup) automatically. Regardless if they visited the login page on meinBerlin before. 
I am not sure if this is the intended behavior or if we should only login/setup users that clicked on "Login via Service-Konto" before

kleingeist:
"/tests/servicekonto/factories.py Imports are incorrectly sorted" but i don't know why, as i can't reproduce the isort error locally  :(

tests/servicekonto/test_servicekonto.py
```diff
@@ -0,0 +1,249 @@
...
+
+    response = client.get(reverse('servicekonto_callback'),
+                          {'Token': '1234567'})
+    assert response.status_code == 302
+
+    response_url = urlparse(response.url)
+    assert response_url.path == '/login_successful'
+
+    user = User.objects.get(email=service_konto.hhgw['email'])
+    assert user.pk
```
2e2a: This will raise before assert

meinberlin/templates/socialaccount/signup.html
```diff
@@ -0,0 +1,24 @@
+{% comment %}
+By default, if the email address is already registered with an existing account, the user will be prompted to enter
+another email address, for which a new account will be created. The new account will be linked to the social account.
+In meinBerlin, instead of showing the form to create the new account, users will be notified that they should login
+to the existing account and connect the social account from the account settings.
+Attention: Providers that won't pass the users email address won't work with this page, as in that case it would be
+mandatory for users to provide an email address to be able to create a new, valid account.
+{% endcomment %}
+
+{% extends "socialaccount/base.html" %}
```
2e2a: Must be above the comment block, otherwise I get an exception.

2e2a:
nice

PR #473, opened 2017-05-29
avoid to translate variables

xi:
This is more of a proposal.

I think that translating variables can mask missing translations in other places. It is better to translate string literals directly. A common case is verbose names on model fields (see https://github.com/liqd/adhocracy4/pull/101)


vellip:
Now that you mention it, I've never seen this method used in Opin. I think all the texts in variables can be marked for translation in python code.

kleingeist:
:+1 
i found only one other occurrence of this pattern which i fixed

PR #475, opened 2017-05-30
[filters] use ugettext_lazy in some cases

vellip:


kleingeist:
we should always use lazy translation for localized class properties like helptexts etc that will be set when the class loads. 
i'd also propose to include the `ugettext_lazy as _` as proposed in the django docs: https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#lazy-translation

PR #476, opened 2017-05-30
project list fallback texts

### LINKED ISSUE #464: project list does not show text if there are no matching projects
It should show something like "No projects could be found". Note that the text should work for two cases: "there are no projects" and "there are no projects matching the filters".

vellip:
fixes #464

I also changed the dashboard project list fallback text since I think "No projects created" is not applicable when there's a filter on.

### DIFF
--- a/apps/projects/templates/meinberlin_projects/project_list.html
+++ b/apps/projects/templates/meinberlin_projects/project_list.html
@@ -14,7 +14,8 @@
             {% if show %}
                 {% include "meinberlin_projects/includes/project_list_tile.html" with project=project %}
             {% endif %}
-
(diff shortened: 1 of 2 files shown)


apps/projects/templates/meinberlin_projects/project_list.html
```diff
@@ -14,7 +14,8 @@
             {% if show %}
```
kleingeist: if there are only projects the user is not allowed to view the alternative `{% empty %}` block won't be displayed. i am not sure how we could fix this. we would have to either filter the queryset in the backend or count the actually displayed projects

PR #478, opened 2017-05-30
Import ugettext_lazy as _ if used as a class property

kleingeist:
Class properties may be evaluated on module load time when the users
locale is not available yet. Thus it is necessary to use ugettext_lazy
to show a correct translation.

For consistency the ugettext_lazy method is imported as `_` if it is
required in a single case. This is in accordance with the django docs
[1,2].

ugettext_lazy can be used as a replacement of ugettext in most cases, as it acts like a **unicode** string object. But it can't be used if the string is handled as a _bytestring_ (**str** object). 

In templates string operations on lazily translated strings can be used as usual.

*Attention*: If a default `LANGUAGE_CODE="de-de"` is set in the `local.py` settings, the failures may not be reproducible.

[1] https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#standard-translation
[2] https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#lazy-
translation

### DIFF
--- a/apps/account/views.py
+++ b/apps/account/views.py
@@ -1,6 +1,6 @@
 from django.contrib.messages.views import SuccessMessageMixin
 from django.shortcuts import get_object_or_404
-from django.utils.translation import ugettext as _
+from django.utils.translation import ugettext_lazy as _
 from django.views import generic
 from django.views.generic.base import RedirectView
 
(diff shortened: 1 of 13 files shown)


xi:
LGTM.

xi:
@MagdaN maybe the information here is useful for you, too.

PR #481, opened 2017-05-30
Allow to filter objects lists based on permissions

### LINKED ISSUE #464: project list does not show text if there are no matching projects

kleingeist:
Fixes #464 

For testing a project can be set to non public via the django admin. Regular users are then not allowed to view that project.


apps/contrib/templatetags/list_rules.py
```diff
@@ -0,0 +1,11 @@
...
+
+register = template.Library()
+
+
+@register.assignment_tag
+def filter_has_perm(perm, user, objects):
+    if not hasattr(user, 'has_perm'):  # pragma: no cover
+        return objects  # swapped user model that doesn't support permissions
+    else:
+        return (obj for obj in objects if user.has_perm(perm, obj))
```
xi: We usually call templatetag files `*_tags.py`.

apps/contrib/templatetags/list_rules.py
```diff
@@ -0,0 +1,11 @@
+from django import template
+
+register = template.Library()
+
+
+@register.assignment_tag
+def filter_has_perm(perm, user, objects):
+    if not hasattr(user, 'has_perm'):  # pragma: no cover
+        return objects  # swapped user model that doesn't support permissions
```
xi: I assume this was copied from somewhere. Maybe a comment would be nice.

apps/projects/templates/meinberlin_projects/project_list.html
```diff
@@ -8,12 +8,10 @@
         {% include "meinberlin_contrib/includes/filter_and_sort.html" with filter=view.filter %}
 
         <div class="l-tiles-4">
-            {% for project in project_list %}
+            {% filter_has_perm 'a4projects.view_project' request.user project_list as filtered_projects  %}
```
xi: I wonder whether a syntax like `{% for project in project_list|has_perm:'a4projects.view_project' %}` would be nicer.

kleingeist: unfortunately this is not possible, as custom filters only accept a single value argument [1]. But we would need the name as the permission as a string and the user as an object. 
[1] https://docs.djangoproject.com/en/1.8/howto/custom-template-tags/#writing-custom-template-filters

kleingeist:
@xi could you please review the changes i made in accordance to your comments

PR #485, opened 2017-05-30
Remove feedback phase from budgeting

kleingeist:
This also removes every existing feedback phase from the database

ToDo:
- [x] translate


2e2a:
lgtm

PR #487, opened 2017-05-30
hide platform link and user indicator for bplan embed

xi:
note that the criterium is whether the project is external, not if it is a bplan. This should work for now, but we may have to change it later on.


kleingeist:
LGTM

PR #488, opened 2017-05-30
allow to login with username

### LINKED ISSUE #484: Login with username
It is currently not possible to login with username, only with email. This *was* possible with the old meinberlin that was based on a3. So the login flow of established users breaks even though we migrated all accounts.

xi:
fixes #484


kleingeist:
lgtm

PR #489, opened 2017-05-30
add ids to paragraphs

xi:
this allows to link to individual paragraphs

I would have liked to have semi-human-readable ids that are generated from the paragraph title, but I could not find a utility function that does the conversion just yet.


kleingeist:
looks good to me.

PR #492, opened 2017-05-30
include info page

xi:
This copies the info page from https://github.com/liqd/meinberlin-wagtail to this repository. It is ugly, but it works.


vellip:
@2e2a said, you'd make a separate repository for this? If I remember correctly, that is.

xi:
> @2e2a said, you'd make a separate repository for this? If I remember correctly, that is.

We dicussed it again. A separate repository would be cleaner, but also a lot more work for deployment. If someone vetos this solution the separate repo could still be an option.

xi:
A much better solution would be to leave the source files in the meinberlin-wagtail repo and only copy the generated files here. I will implement that separately.

PR #500, opened 2017-05-31
Fix topic creation permissions

2e2a:


2e2a:
peer programming with @kleingeist

PR #504, opened 2017-05-31
add info page

xi:
replaces #492


vellip:
Does that mean if we change something we have to change it in the original repo, then build it, take the changed build over to this repo and make a pull request? If so, this sounds tedious, and I do see changes coming, especially with the launch.

xi:
> Does that mean if we change something we have to change it in the original repo, then build it, take the changed build over to this repo and make a pull request?

Yes. However, we do not need to create a pull request here every time we change the source. It is a compromise for sure.

PR #522, opened 2017-06-06
Complex text models

kleingeist:
missing
* react components to create and sort chapters


vellip:
Just a note: _the react components to create and sort chapters_ will most likely **not** be like the component we have in polls or the previous Text Review create view. I talked about that with @damusp and it would not work. I think he's working on a different solution.

kleingeist:
@2e2a could you please review, so that @xi can start working on the frontend

xi:
When running the migrations I gut the following warning:

```
The following content types are stale and need to be deleted:

    meinberlin_documents | document

Any objects related to these content types by a foreign key will also
be deleted. Are you sure you want to delete these content types?
If you're unsure, answer 'no'.
```

I accepted and it seems to work fine, but we should make sure that this does not delete relevant information (e.g. comments). @2e2a could you please look into that?

kleingeist:
@xi @2e2a : comments on documents will be deleted when the Document model is renamed to Chapter. I will look into it and try to get a solution for migrating the comments

kleingeist:
@xi @2e2a i did add a migration to update the content type model.
fyi: this has been a known issue and has been fixed with newer django versions :( 
https://code.djangoproject.com/ticket/24067

apps/documents/urls_compatibility.py
```diff
@@ -0,0 +1,15 @@
...
+from django.views.generic.base import RedirectView
+
+
+urlpatterns = [
+    url(
+        r'^(?P<pk>\d+)/$',
+        RedirectView.as_view(
+            pattern_name='meinberlin_documents:paragraph-detail',
+            permanent=True,
+        )
```
2e2a: As discussed, I would suggest to do the redirect in nginx.

2e2a:
lgtm. migrations work.

apps/documents/urls.py
```diff
@@ -4,8 +4,19 @@
...
+        name='chapter-detail'
+    ),
+    url(
+        r'^paragraphs/(?P<pk>\d+)/$',
         views.ParagraphDetailView.as_view(),
         name='paragraph-detail'
     ),
+    # url(r'^chapters/(?P<pk>\d+)/$'
+    url(
+        r'^manage/chapter/(?P<pk>\d+)/$',
```
2e2a: Better plural for consistancy

2e2a:
permission checks work

2e2a:
lgtm

PR #523, opened 2017-06-07
Fix signup terms of use translation

kleingeist:
The msgid has to be exactly the same in all translations.
And unfortunatly it is not possible to have the origin of the translated
string to end on a newline while the actual translation does not. Thus
it is necessary to add an empty line to translation.


2e2a:
lgtm

PR #526, opened 2017-06-07
chapter styling

xi:
builds on #522 

This implements both the new document detail styling and styling for chapter navigation. Still to do: The table of content should be in an accordion.

Note that there was some styling-related refactoring, so @vellip should have a look.


vellip:
Code looks alright for me. 

Since we touch that part of the code now anyway, especially important for the coming Text Review changes for the Dashboard: We should keep in mind that the styling that resides in `_commenting.scss` also applies to the poll dashboard.

PR #527, opened 2017-06-07
update isort

xi:
The isort update in #514 caused travis to produce errors. So here I fixed it separately. The required changes seem reasonable to me.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,3 +18,4 @@ indent_style = tab
 [*.py]
 skip=migrations
 known_first_party=adhocracy4, meinberlin, apps, tests
+force_single_line = true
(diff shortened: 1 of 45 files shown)


kleingeist:
still gives different results on my local dev machine which is weird. so can't really review this PR :(

i'd also suggest to remove the isort `-sl` parameter from `.polylintrc` in favor for a explicit `force_single_line = true` in the `.editorconfig`

2e2a:
> i'd suggest to remove the isort -sl parameter from .polylintrc in favor for a explicit force_single_line = true in the .editorconfig

It's a good idea. Done.

2e2a:
lgtm

PR #529, opened 2017-06-08
Simple Action Stream 

kleingeist:
@xi: This is just a rough draft and not integrated into wagtail yet, but i hope it still allows you to start with the layout.


xi:
I feel that it is frustrating for users that there is no link to the activity object (e.g. a comment or idea). The closest think is a link to the project and it is very tedious to go from there to a specific comment.

I understand that a link to the object would be hard, because objects can be very different things. In some cases (such as comments) it also may be very hard to find a place to link to. In those cases, we should link to the most specific possible place, e.g. the page that a comment is on.

apps/activities/blocks.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class PlatformActivityBlock(blocks.StructBlock):
+    heading = blocks.CharBlock(label=_('Heading'))
+    count = blocks.IntegerBlock(label=_('Count'), default=5)
+
+    def get_context(self, value, parent_context=None):
+        context = super().get_context(value, parent_context=parent_context)
+        block = context['self']
+        context['activities'] = Activity.objects.all()[:block['count']]
```
xi: I am not sure whether additional filters are necessary here, e.g. to remove activities from private projects.

PR #532, opened 2017-06-12
rm participated projects from user profile

### LINKED ISSUE #530: Participated projects visible for all on user profile
If I write a comment on platform meinBerlin and take part in a participation process, everyone can see my profile and in which participation process I took part, even I'm not logged in. It's not anonymous enough.

xi:
fixes #530


kleingeist:
lgtm

PR #540, opened 2017-06-13
mv cms blocks to separate file

xi:


apps/cms/migrations/0004_add_projects_block.py
```diff
@@ -3,7 +3,7 @@
 
 from django.db import migrations, models
 import wagtail.wagtailcore.fields
-import apps.cms.models
+import apps.cms.blocks
```
kleingeist: i don't think existing miragrations should be changed.

xi: If I do not change this `migrate` and `makemigrations` will fail with an exception.

kleingeist:
lgtm

PR #541, opened 2017-06-13
link to object in action

xi:
it may be necessary to define `get_absolute_url()` on some more models


kleingeist:
the truncated text is still very short, but i don't see a solution while keeping the current layout

PR #542, opened 2017-06-13
Allow additional apps to be a list

2e2a:
Easier for integration with salt deployment


2e2a:
done in peer programming

PR #543, opened 2017-06-13
Add missing get_absolute_url implementations and use them in templates

kleingeist:


apps/budgeting/models.py
```diff
@@ -13,5 +12,6 @@ class Proposal(mapidea_models.AbstractMapIdea, Moderateable):
     )
 
     def get_absolute_url(self):
+        from django.core.urlresolvers import reverse
```
xi: why did you move this here? I guess it is not because of circular dependencies.

kleingeist: good question. i did follow the django docs [1] and then tried to unify the form. actually i can't see any reason on why to import it within the method. 
i will change it everywhere

https://docs.djangoproject.com/en/1.11/ref/models/instances/#get-absolute-url

PR #544, opened 2017-06-14
update leaflet to 1.0.x

xi:
depends on https://github.com/liqd/adhocracy4/pull/107


vellip:
Detailpage with maps fail for me with "Bounds are not valid" (Bürgerhaushalt)

xi:
>  Detailpage with maps fail for me with "Bounds are not valid" (Bürgerhaushalt)

Cannot reproduce. I got the error at first, but it was gone after reloading (so probably the build was not complete on first try).

vellip:
Huh. Works for me [now?].

PR #547, opened 2017-06-14
style service konto login

xi:
This is my porposal for service konto styling. The idea is to make it as simple as possible because service konto is currently our only SSO option. Note that I removed the possibility to add more account connections once the servicekonto is connected.

@damusp what do you think?

screenshot from 2017-06-14 17-56-17

screenshot from 2017-06-14 17-56-04

screenshot from 2017-06-14 17-55-42

### DIFF
--- a/meinberlin/templates/account/login.html
+++ b/meinberlin/templates/account/login.html
@@ -6,24 +6,8 @@
 {% block content %}
     <h1>{% trans "Login" %}</h1>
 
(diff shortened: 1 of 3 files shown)


damusp:
the dashboard view is alright imo
i am not so much convinced about the login/register form though. 

but I just talked to @kleingeist and we decided that it would be best to merge your this – and create a styling issue for some details that need to be changed.

PR #549, opened 2017-06-15
Connect all signals from a Proxy model to its parent

kleingeist:
There is 9 year old bug report where the problem is discussed:
https://code.djangoproject.com/ticket/9318


2e2a:
lgtm

PR #550, opened 2017-06-15
Fix email notifications

kleingeist:
Moderators did not get any emails as there was a regression bug
introduced with the default email language.

This commit also enforces the use of the meinBerlin Action proxy when
using signals.

Note, that signals for Actions are not automatically bubbled down to
proxies, as discussed in https://code.djangoproject.com/ticket/9318

### DIFF
--- a/apps/actions/models.py
+++ b/apps/actions/models.py
@@ -39,3 +39,10 @@ class Action(A4Action):
             return 'clock-o'
         else:
             return 'star'
+
+    @staticmethod
+    def from_parent(action):
+        """Cast an A4Action object to the proxied Action."""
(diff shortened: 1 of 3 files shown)


apps/actions/models.py
```diff
@@ -39,3 +39,10 @@ def icon(self):
             return 'clock-o'
         else:
             return 'star'
+
+    @staticmethod
+    def from_parent(action):
```
2e2a: Maybe we can use a name making the proxy relation more explicit. E.g. proxy_of()

2e2a:
lgtm

PR #551, opened 2017-06-15
Prevent access to items before the project has started

### LINKED ISSUE #525: Use predicates to prevent access to items if no phase has started yet
With topicprio and polls items are created before any phase has been started.
On the project_detail page a according message and no content is shown.
This is checked via the `PhaseDispatchMixin`.
As for items it is not possible to use the `PhaseDispatchMixin` a new
predicate has been introduced which tests if a phase has been started
yet. Be aware, that this does not imply an active phase.
If no phase has been started yet access to the items should be prevented. 
But if participation did finish the items should still be available.
See 362a8ce6a79888b2cc1dbdeb789107a156c5d15f for reference.

kleingeist:
fixes #525

### DIFF
--- a/apps/polls/rules.py
+++ b/apps/polls/rules.py
@@ -2,6 +2,7 @@ import rules
 from rules.predicates import is_superuser
 
 from adhocracy4.modules import predicates as module_predicates
+from apps.contrib import predicates as contrib_predicates
 
 from . import models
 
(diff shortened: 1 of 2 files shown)


2e2a:
Works for topicprio. Why do we need this for polls?

kleingeist:
currently there is no explicit view for a single poll and the PhaseDispatchMixin shows a message instead of the poll if the phase has not started yet, but if some will add an view or even in case of an Poll api endpoint the view_poll rule will be automatically checked, and i thought it would be useful to fix the rule now when i am already at it

2e2a:
Makes sense.

PR #552, opened 2017-06-16
Unify item ordering and filter

### LINKED ISSUE #524: When I switch from map to list view the list is not sorted by latest entries as the sort dropdown suggests

2e2a:
- Default ordering for items
- Default filter ordering

Fixes #524


kleingeist:
@2e2a i did some changes in this PR
could you please re-review and if it looks good to you merge it?

PR #554, opened 2017-06-16
Basic offline event documentation

2e2a:
This pull request starts the offline event documentation implementation by:
- add app
- add model
- add detail view
- exemplify inclusion in phase indicator
- management views

As the designs are not yet clear, further steps need to be discussed.


apps/offlineevents/views.py
```diff
@@ -10,3 +19,102 @@ class OfflineEventDetailView(module_views.ItemDetailView):
     @property
     def project(self):
         return self.get_object().project
+
+
+class OfflineEventMgmtView(DashboardBaseMixin,
```
kleingeist: I am not a big fan of naming the methods Mgmt/managment, as those are very special views that have to be initially routed via the quirky lookup found due to a PhaseContent management_view property. 
Offline events are different as they will be available for every project regardless of the blueprint/phases.
Also every  view of the offline events settings are routed via it's own urls.py.

I suggest to either remove the Mgmt/management and the manage route altogether or replace it by something called dashboard.

For the routes i would prefer `dashboard/projects/{slug}/offlineevents` analogous to 
 `dashboard/projects/{slug}/moderators` and the management  start point `dashboard/projects/{slug}/manage`-
Unfortunately this would be a little messay, as this route would have to be defined in the context of the dashboard app :(

apps/offlineevents/views.py
```diff
@@ -10,3 +19,102 @@ class OfflineEventDetailView(module_views.ItemDetailView):
...
+                           rules_mixins.PermissionRequiredMixin,
+                           generic.ListView):
+    model = models.OfflineEvent
+    template_name = 'meinberlin_offlineevents/offlineevent_mgmt_list.html'
+    permission_required = 'a4projects.add_project'
```
kleingeist: maybe we should think about introducing a predicate `a4projects.moderate_project` or `a4projects.edit_project` for all the project dashboard related views.
we could make that predicate intelligent, so that it may accept an `Item` (or anything with a `.project` property), or a `Project` itself, or an `Organisation` and checks if the user is an initiator or in the projects moderator list

apps/offlineevents/views.py
```diff
@@ -10,3 +19,102 @@ class OfflineEventDetailView(module_views.ItemDetailView):
...
+    template_name = \
+        'meinberlin_offlineevents/offlineevent_mgmt_create_form.html'
+    menu_item = 'project'
+
+    @property
+    def organisation(self):
+        return self.project.organisation
+
+    def get_permission_object(self):
+        return self.organisation
```
kleingeist: the rule 'meinberlin_offlineevents.add_offlineevent' expects an object with a `.project` property (f.ex. a module or an item)
With an `Organisation` the rule `a4projects.add_project` may be used

apps/offlineevents/urls.py
```diff
@@ -5,4 +5,15 @@
...
+        views.OfflineEventMgmtView.as_view(), name='offlineevent-mgmt'),
+    url(r'^manage/create/project/(?P<slug>[-\w_]+)/$',
+        views.OfflineEventMgmtCreateView.as_view(),
+        name='offlineevent-mgmt-create'),
+    url(r'^manage/(?P<slug>[-\w_]+)/update/$',
+        views.OfflineEventMgmtUpdateView.as_view(),
+        name='offlineevent-mgmt-update'),
+    url(r'^manage/(?P<slug>[-\w_]+)/delete/$',
+        views.OfflineEventMgmtDeleteView.as_view(),
+        name='offlineevent-mgmt-delete'),
```
kleingeist: see the comments below

> For the routes i would prefer dashboard/projects/{slug}/offlineevents analogous to
dashboard/projects/{slug}/moderators and the management start point dashboard/projects/{slug}/manage-
Unfortunately this would be a little messay, as this route would have to be defined in the context of the dashboard app :(

apps/offlineevents/views.py
```diff
@@ -10,3 +19,102 @@ class OfflineEventDetailView(module_views.ItemDetailView):
...
+    template_name = 'meinberlin_offlineevents/offlineevent_mgmt_list.html'
+    permission_required = 'a4projects.add_project'
+    menu_item = 'project'
+
+    def dispatch(self, *args, **kwargs):
+        self.project = Project.objects.get(slug=kwargs['slug'])
+        return super(OfflineEventMgmtView, self).dispatch(*args, **kwargs)
+
+    def get_queryset(self):
+        return super().get_queryset().filter(project__slug=self.project)
```
kleingeist: self.project is not a slug. should be `.filter(project=self.project)`

apps/offlineevents/views.py
```diff
@@ -0,0 +1,120 @@
...
+class OfflineEventDetailView(module_views.ItemDetailView):
+    model = models.OfflineEvent
+    permission_required = 'meinberlin_offlineevents.view_offlineevent'
+
+    @property
+    def project(self):
+        return self.get_object().project
+
+
+class OfflineEventMgmtView(DashboardBaseMixin,
```
kleingeist: not sure if we should use DashboardBaseMixin here or write a more general mixin that does not already overwrite `get_permission_object` and `get_success_url` as those are very specific to the dashboard create/edit project views

also i think the `slug` kwarg is to lax for a general purpose mixin. i'd prefer project_slug

apps/offlineevents/views.py
kleingeist: btw. the same problem occurs for `meinberlin_topicprio.add_topic`

apps/offlineevents/views.py
kleingeist: note: in topicprio a equivalent rule `meinberlin_topicprio.add_topic` is checked. there it does work, as `TopicMgmtCreateView` inherits from `ItemCreateView` which does lookup a module via the `slug` kwarg (which i deem also bad choice, as slug is just to gneric)

apps/offlineevents/urls.py
```diff
@@ -0,0 +1,19 @@
+from django.conf.urls import url
+
+from . import views
+
+urlpatterns = [
+    url(r'^(?P<slug>[-\w_]+)/$',
+        views.OfflineEventDetailView.as_view(), name='offlineevent-detail'),
+    url(r'^/manage/projects/(?P<slug>[-\w_]+)/$',
```
kleingeist: remove leading `/`

apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
```diff
@@ -1,32 +1,40 @@
...
+                            href="{% url 'project-detail' project.slug %}"
+                        class="phase-list__button {% if obj == active_phase %} is-active{% endif %}"
+                    >
+                        {{ obj.name }}
+                    </a>
+                </li>
+            {% else %}
+                <li>
+                    <a
+                        href="{% url 'meinberlin_offlineevents:offlineevent-detail' obj.slug %}"
```
kleingeist: i'd prefer to set `get_absolute_url` on `OfflineEvent` in `modely.py` and here just `href="{ obj.get_absolute_url }"`
(see #543)

apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
```diff
@@ -1,32 +1,40 @@
...
-                    data-html="true"
-                >
-                    {{ phase.name }}
-                </button>
-            </li>
+        {% for obj in phases_and_events %}
+            {% if obj|is_phase %}
+                <li>
+                    <a
+                            href="{% url 'project-detail' project.slug %}"
```
kleingeist: i'd prefer to use `{ project.get_absolute_url }`

apps/offlineevents/views.py
2e2a: Removed the mgmt naming.

With regards to the URLs. I agree with your suggestion, but would fix it with the dashboard  refactoring that anyway will have to define an URL schema.

apps/offlineevents/views.py
2e2a: As this is a core issue, we can discuss it with the others.

2e2a:
Updated according to comments

kleingeist:
lgtm.

PR #555, opened 2017-06-19
Complex text management admin

kleingeist:
missing
- [x] tests
- [x] server site weight 
- [x] updating of chapters/paragraphs
- [x] error handling in javascript


apps/documents/assets/ChapterForm.jsx
```diff
@@ -0,0 +1,61 @@
...
+                id={paragraph.id || paragraph.key}
+                key={paragraph.id || paragraph.key}
+                index={index} // why is this needed for ck?
+                paragraph={paragraph}
+                errors={{}}
+                config={this.props.config}
+                onDelete={() => { this.props.onParagraphDelete(index) }}
+                onMoveUp={index !== 0 ? () => { this.props.onParagraphMoveUp(index) } : null}
+                onMoveDown={index < arr.length - 1 ? () => { this.props.onParagraphMoveDown(index) } : null}
+                onParagraphAddBefore={() => { this.props.onParagraphAddBefore(index) }}
```
xi: Is this still needed?

apps/documents/assets/ChapterListItem.jsx
```diff
@@ -0,0 +1,81 @@
...
+//         className="button button--light"
+//         onClick={handleDelete}
+//         title={django.gettext('Delete')}
+//         type="button">
+//         <i className="fa fa-trash"
+//            aria-label={django.gettext('Delete')}/>
+//       </button>
+//     </div>
+//   </div>
+// )}
```
xi: should be removed

apps/documents/assets/DocumentManagement.jsx
```diff
@@ -0,0 +1,315 @@
...
+      key: this.getNextLocalKey(),
+      weight: 0,
+      paragraphs: []
+    }
+  },
+
+  handleChapterMoveUp: function (index) {
+    const value = this.state.chapters[index]
+    const diff = {$splice: [[index, 1], [index - 1, 0, value]]}
+    let editChapterIndex = this.state.editChapterIndex
```
xi: Tracking the current chapter by index seems tedious to me. Is it not possible to track it by something that does not change all the time (e.g. key/id)?

meinberlin/assets/scss/components/_commenting.scss
```diff
@@ -1,18 +1,20 @@
...
     padding: $padding;
     border: 1px solid $border-color;
     border-radius: ($padding / 2);
 }
 
+.commenting--toc__button {
+    display: block;
+    width: 100%;
+    text-align: left;
+}
```
xi: This class should either not add styling or not be combined with `.button`

xi:
Did you remove FlipMove on purpose?

apps/documents/assets/ChapterList.jsx
```diff
@@ -0,0 +1,37 @@
+var React = require('react')
+var django = require('django')
+var ChapterListItem = require('./ChapterListItem')
+
+const ChapterList = React.createClass({
```
xi: I feel like the name `ChapterList` is a bit confusing. Maybe `ChapterNav` would be better. I am not convniced though.

apps/documents/assets/ParagraphForm.jsx
```diff
@@ -1,80 +1,74 @@
...
+        this.handleTextChange(text)
       }.bind(this))
       editor.setData(this.props.paragraph.text)
     }
   },
+
   componentWillUpdate: function (nextProps) {
     if (nextProps.index > this.props.index) {
-      this.ckEditorDestroy()
+      this.ckEditorDestroy() // why is this needed?
```
xi: see a4fc734de131105d2f194bac680410d0ad79586e: We had the issue that ckeditor became unresponsive after moving it around. I do not remember whether this was because of react or only in combination with FlipMove.

apps/documents/assets/ParagraphForm.jsx
```diff
@@ -1,80 +1,74 @@
...
+  handleNameChange: function (e) {
+    const name = e.target.value
+    this.props.onNameChange(name)
   },
-  down: function () {
-    this.props.moveParagraphDown(this.props.index)
+
+  handleTextChange: function (text) {
+    this.props.onTextChange(text)
   },
```
xi: I do not see the point of this indirection.

apps/documents/assets/DocumentManagement.jsx
kleingeist: yes i thought about that too, but as errors are returned from django index based it would double the work

apps/documents/assets/ChapterList.jsx
kleingeist: how about ChapterToc to be in line with the css?

apps/documents/serializers.py
```diff
@@ -6,55 +6,105 @@
...
+                         for item in paragraphs_data
+                         if 'id' in item]
+        chapter.paragraphs.exclude(id__in=paragraph_ids).delete()
+
+        # Update or create the chapters
+        for weight, paragraph_data in enumerate(paragraphs_data):
+            paragraph_data['weight'] = weight
+            if 'id' in paragraph_data:
+                paragraph = Paragraph.objects.get(id=paragraph_data['id'])
+                assert paragraph.chapter == chapter
```
2e2a: Maybe move to a validator?

apps/documents/serializers.py
```diff
@@ -6,55 +6,105 @@
...
+            .filter(module_id=self.context['module_pk']) \
+            .exclude(id__in=chapter_ids) \
+            .delete()
+
+        for weight, chapter_data in enumerate(validated_data):
+            chapter_data['weight'] = weight
+            if 'id' in chapter_data:
+                instance = Chapter.objects.get(id=chapter_data['id'])
+                assert instance.module_id == self.context['module_pk']
+                yield chapter_serializer.update(instance, chapter_data)
```
2e2a: Why using `yield`? It is directly transformed to list().

kleingeist: i was to lazy to build the resulting arrays with "result = [] ... result.append(instance) ... return result"
but i can change it

apps/documents/serializers.py
kleingeist: when validating i don't have the any instances and would have to duplicate the code. it would probably be better, but DRF does not really want you to check on models, but rather the json itself :(

2e2a:
Api, serializer and test lgtm

meinberlin/assets/scss/_form.scss
```diff
@@ -153,3 +153,7 @@ select {
         border-bottom-right-radius: 0.3em;
     }
 }
+
+.errorcount {
+    color: $brand-danger;
+}
```
xi: This should replace `.tab-errorcount`

xi:
Nice!

PR #557, opened 2017-06-20
collapse chapter TOC

xi:
This makes the chapter TOC collapsible and collapses it on all chapter detail
pages, except for the first one.

It uses bootstrap's [collapse component](https://v4-alpha.getbootstrap.com/components/collapse/).


kleingeist:
lgtm

PR #558, opened 2017-06-20
rm truncated comment from activity stream

xi:
this change was requested by @damusp:

> This looks rather odd. Also, the quote is too short to ever get any meaning from it.


kleingeist:
lgtm

PR #562, opened 2017-06-21
Filter public actions

kleingeist:
Add a custom QuerySet that allows to filter public actions.
An action is public if it is either not connected to a project or the
project is published and public.


2e2a:
lgtm

PR #563, opened 2017-06-21
Only inlcude js strings from a4 apps

### LINKED ISSUE #560: Making messages for translation has system dependent output
When calling `make makemessages` the output of the generated "djangojs.po" is dependent on the system configuration. 
For example in my case it inserts messages from:
"adhocracy4/node_modules/ajv/dist/regenerator.min.js"  and
"adhocracy4/node_modules/js-yaml/dist/js-yaml.min.js"

xi:
Fixes #560

Most importantly, this avoids importing strings from adhocracy4/node_modules.


kleingeist:
lgtm, results in correct messages

PR #565, opened 2017-06-21
Use controlled components

kleingeist:
If the value label is undefined (calling `getNewChoice()` without a parameter) the input field is initially uncontrolled.
For controlled components see:
https://facebook.github.io/react/docs/forms.html#controlled-components


xi:
Looks fine to me.

Just one detail to make this more robust: We could use `''` as default value inside of `getNewChoise()` like this:

    label: label || ''

kleingeist:
> Just one detail to make this more robust: We could use '' as default value inside of getNewChoise() like this:
> `label: label || ''`

i preferred the new syntax: `getNewQuestion: function (label = '')` and changed the functions method in that way

PR #566, opened 2017-06-21
Filter UPDATE actions from the activity stream

kleingeist:


xi:
Now there are mostly comment add actions left. I am not completely sure if this is preferable. We will have to see how this works in practice.

PR #567, opened 2017-06-21
Re-Style login form

### LINKED ISSUE #548: Re-Style login/register form
as noted by @damusp in #547 the login and register are in need of some re-styling

xi:
fixes #548

@damusp what do you think?

screenshot from 2017-06-21 17-05-10


damusp:
cool tobi! it's a much more delightful form now i think. 

there's just one more detail that I've been wishing for, which is reducing the width of the form fields. Ideally they should not be so wide here because the inserted data will most likely be relatively short.

PR #573, opened 2017-06-22
Use raw_id_field for user selection

kleingeist:
With a large user base (>20000) the multi choice widget from django
admin gets unresponsive.


2e2a:
lgtm

PR #577, opened 2017-06-26
Allow blank moderator statements

### LINKED ISSUE #575: Can't edit proposals with moderator_statement in django admin
No proposal can be edited/created with django admin because it's missing the moderator_statement which is a required field for some reason.

kleingeist:
While moderator statements had been optional (null=True) they had been
marked as required in forms, because the (blank=True) option was not
set.

Fixes #575


2e2a:
lgtm

PR #578, opened 2017-06-26
Mark the currently edited Chapter in the TOC

### LINKED ISSUE #576: Text Review: Highlight current chapter in dashboard

kleingeist:
Fixes #576

### DIFF
--- a/apps/documents/assets/ChapterNav.jsx
+++ b/apps/documents/assets/ChapterNav.jsx
@@ -4,22 +4,27 @@ var ChapterNavItem = require('./ChapterNavItem')
 var FlipMove = require('react-flip-move')
 
 const ChapterNav = (props) => {
+  const activeKey = props.activeChapter.id || props.activeChapter.key
   return (
     <nav aria-label={django.gettext('Chapter navigation')}>
       <FlipMove easing="cubic-bezier(0.25, 0.5, 0.75, 1)" typeName="ol" className="u-list-reset">
         {
-          props.chapters.map((chapter, index, arr) =>
-            <li key={chapter.id || chapter.key}>
-              <ChapterNavItem
-                name={chapter.name}
-                onMoveUp={index !== 0 ? () => { props.onMoveUp(index) } : null}
-                onMoveDown={index < arr.length - 1 ? () => { props.onMoveDown(index) } : null}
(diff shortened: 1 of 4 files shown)


apps/documents/assets/ChapterNav.jsx
```diff
@@ -4,22 +4,27 @@ var ChapterNavItem = require('./ChapterNavItem')
...
-                onMoveUp={index !== 0 ? () => { props.onMoveUp(index) } : null}
-                onMoveDown={index < arr.length - 1 ? () => { props.onMoveDown(index) } : null}
-                onDelete={() => { props.onDelete(index) }}
-                onClick={() => { props.onClick(index) }}
-                errors={props.errors ? props.errors[index] : {}}
-                />
-            </li>
-          )
+          props.chapters.map((chapter, index, arr) => {
+            const key = chapter.id || chapter.key
```
vellip: Does this fallback work? When is this applied? Because `component.key`should be undefined, I suppose: https://github.com/facebook/react/issues/2429
Same goes for where ever else this is used.

apps/documents/assets/ChapterNavItem.jsx
```diff
@@ -14,7 +14,10 @@ function getErrorCount (props) {
 const ChapterNavItem = (props) => {
   return (
     <div className="commenting">
-      <button className="commenting__content commenting--toc__button button button--light button--small" type="button" onClick={props.onClick}>
+      <button
+        type="button"
+        className={`commenting__content commenting--toc__button button button--light button--small ${props.active ? 'active' : ''}`}
```
vellip: You could use `classnames` here. I think we have it anyway: https://www.npmjs.com/package/classnames

apps/documents/assets/ChapterNav.jsx
kleingeist: actually this is just bad naming. chapter.key does access the attribute "key" from a chapter object in props, not the components key (even though the objects key is used as the components key is necessary)

apps/documents/assets/ChapterNavItem.jsx
kleingeist: thank, nice hint

PR #579, opened 2017-06-26
Exemplary tests for idea item rules

kleingeist:


tests/ideas/rules/test_rules_add.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.mark.django_db
+def test_rules_pre_phase(phase_factory, user):
+    phase, module, project, _ = setup_phase(phase_factory, None,
+                                            phases.CollectPhase)
+    anonymous, moderator, initiator = setup_users(project)
+
+    assert project.is_public
+    with freeze_time(phase.start_date - timedelta(days=1)):
```
2e2a: Would be nice to define helper for all the freeze functions. E.g. freeze_phase/pre_phase/after_phase.

2e2a:
Looks great.

I would suggest more explicit test names:
- `test_pre_phase` (assuming project being public per default)
- `test_rules_phase_active`  (assuming project being public per default)
- `test_rules_phase_active_project_draft`
- `...`

What do you think?

2e2a:
lgtm

PR #583, opened 2017-06-28
Refactor ProjectMixin

kleingeist:
depends on liqd/adhocracy4#123

did depend on liqd/adhocracy4#115 previously, but the functionality was moved to the contrib app

Note: https://github.com/liqd/adhocracy4/blob/master/adhocracy4/modules/views.py is no longer in use and may be deleted


2e2a:
nice

PR #584, opened 2017-06-28
Do not show user and date of topics

2e2a:
As topics are created by the same initiator at one time, having this information shown looks weird and is not interesting for the users.


kleingeist:
lgtm.
i just ensured the horiziontal line below the text is shown even if item-detail__meta is not available. 
there is also a minor problem with spaces in the list view. but i opened issue #590 for that

PR #588, opened 2017-06-29
Use project tile image

2e2a:
Depends on: https://github.com/liqd/adhocracy4/pull/118


apps/projects/templates/meinberlin_projects/includes/project_list_tile.html
```diff
@@ -4,7 +4,11 @@
     <div class="tile__head">
         <a tabindex="-1" class="tile__image tile__image--shadowed"
            aria-labelledby="project-title-{{ project.pk }}"
-           {% if project.image %}style="background-image: url({{ project.image|thumbnail_url:'project_thumbnail' }})"{% endif %}
+           {% if project.tile_image %}
+               style="background-image: url({{ project.tile_image|thumbnail_url:'project_thumbnail' }})"
+           {% else %}
```
kleingeist: should be an else if, as it is not guaranteed, that the project.image exists

kleingeist:
fixed checking if project.image exists in templates and merged

PR #589, opened 2017-06-29
Add an export framework and item export functionality

kleingeist:
The export framework is designed with extensibility in mind, while the current implementation only allows for exporting "anything which looks like a item list"

Furthermore until now it is only allowed to have on export view per project. In the future it may be possible to select different exports for a project. For example for different modules.

### DIFF
--- a/apps/exports/views.py
+++ b/apps/exports/views.py
@@ -8,6 +8,7 @@ from django.utils.translation import ugettext as _
 from django.views import generic
 
 from adhocracy4.comments.models import Comment
+from adhocracy4.modules import models as module_models
 from adhocracy4.ratings.models import Rating
 
 
@@ -16,10 +17,26 @@ class VirtualFieldMixin:
         return OrderedDict()
 
 
-class ItemExportView(generic.ListView, VirtualFieldMixin):
+class AbstractCSVExportView(generic.View):
+    def get(self, request, *args, **kwargs):
+        response = HttpResponse(content_type='text/csv; charset=utf-8')
+        response['Content-Disposition'] = (
+            'attachment; filename="ideas.csv"'
+        )
+
+        writer = csv.writer(response, lineterminator='\n',
+                            quotechar='"', quoting=csv.QUOTE_ALL)
+        writer.writerow(self.get_header())
+        writer.writerows(self.export_rows())
+
+        return response
+
+
+class ItemExportView(AbstractCSVExportView,
+                     VirtualFieldMixin,
+                     generic.list.MultipleObjectMixin):
     fields = None
     exclude = None
-    virtual = None
     model = None
 
     def __init__(self):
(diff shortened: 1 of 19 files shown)


apps/exports/views.py
```diff
@@ -0,0 +1,324 @@
...
+    def get_redirect_url(self, *args, **kwargs):
+        project = self.get_object()
+
+        assert project.module_set.count() == 1
+        module = project.module_set.first()
+
+        exports = get_exports(project)
+        assert len(exports) <= 1
+
+        # Show error page if no exports are available
```
2e2a: Is this a TODO?

apps/exports/views.py
```diff
@@ -0,0 +1,324 @@
...
+        meta = self.model._meta
+        exclude = self.exclude if self.exclude else []
+
+        if self.fields:
+            fields = [meta.get_field(name) for name in self.fields]
+        else:
+            fields = meta.get_fields()
+
+        names = ['link']
+        header = [_('Link')]
```
2e2a: This is set in `get_virtual_fields`

apps/exports/views.py
```diff
@@ -0,0 +1,324 @@
...
+            if hasattr(get_field_attr, '__call__'):
+                return get_field_attr(item)
+            return get_field_attr
+
+        # If item is a dict, return the fields data by key
+        try:
+            if name in item:
+                return item['name']
+        except TypeError:
+            pass
```
2e2a: When does this TypeError occur?

apps/exports/views.py
```diff
@@ -0,0 +1,324 @@
...
+        if 'ratings_positive' not in virtual:
+            virtual['ratings_positive'] = _('Positive ratings')
+        if 'ratings_negative' not in virtual:
+            virtual['ratings_negative'] = _('Negative ratings')
+
+        return virtual
+
+    def _count_ratings(self, item, value):
+        ct = ContentType.objects.get_for_model(item)
+        return Rating.objects.filter(content_type=ct, value=value).count()
```
2e2a: Could use the annotation value here.

apps/exports/views.py
```diff
@@ -0,0 +1,324 @@
...
+        else:
+            self.module = \
+                module_models.Module.objects.get(slug=kwargs['module_slug'])
+
+        return super().dispatch(request, *args, **kwargs)
+
+    def get(self, request, *args, **kwargs):
+        response = HttpResponse(content_type='text/csv; charset=utf-8')
+        response['Content-Disposition'] = (
+            'attachment; filename="ideas.csv"'
```
2e2a: File name should be project/module specific with a timestamp.

apps/exports/views.py
kleingeist: it the item is no dict. for example if item is an django model object.

apps/exports/views.py
kleingeist: im trying to get the annotated rating in the get_ratings_x_data methods. the _count_ratings is only called if it is not available

tests/exports/factories.py
```diff
@@ -0,0 +1,47 @@
...
+    content_object = factory.SubFactory(idea_factories.IdeaFactory)
+
+
+class ModeratorStatementFactory(factory.django.DjangoModelFactory):
+
+    class Meta:
+        model = moderatorfeedback_models.ModeratorStatement
+
+    statement = factory.Faker('text')
+    creator = factory.SubFactory(factories.UserFactory)
```
2e2a: Maybe we should move the general factories (Rarting, ModeratorStatement) to test/factories.py?
Otherwise we might end up duplicating them in the next app.

2e2a:
lgtm

PR #593, opened 2017-06-29
Make creators readonly in django admin

### LINKED ISSUE #574: Make creator a readonly field in django admin
With a large user base the dropdown used for the user selection in django admin becomes unresponsive. As there is no apparent reason to change the creator of an object (comment, item, rate, poll, ...) the creator field should be made read_only.

The same problem occurs for every model where a user is set as a foreign key (Action: actor, Projects: member/moderator, Organisation: initiator)

The readonly property can either be set on every Admin class by setting ` readonly_fields = ('creator',)` individually or by using a custom `A4Admin(admin.ModelAdmin)` which has to be set as the parent for every Admin class used.

kleingeist:
Fixes #574

### DIFF
--- a/apps/documents/admin.py
+++ b/apps/documents/admin.py
@@ -3,9 +3,12 @@ from django.contrib import admin
 from . import models
 
 
+@admin.register(models.Paragraph)
 class ParagraphAdmin(admin.ModelAdmin):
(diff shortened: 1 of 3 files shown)


2e2a:
lgtm

PR #600, opened 2017-07-05
Use xlsx format for data exports

kleingeist:
This is using xlsx instead of csv for data exports. This should allow for Microsoft Excel to import the data on all platforms.

Unfortunately the exported data is not perfect, as it is still tailored to the flat CSV structure. So comments are remaining in a single field and dates are emitted as strings. 
The main purpose of this PR is to allow for easier opening of exports. Reformatting the export for example to multiple spreadsheets etc. would be a new user story


2e2a:
lgtm

PR #602, opened 2017-07-06
Test rules

2e2a:
Apply #579 to all apps


tests/polls/rules/test_rules_add_chapter.py
```diff
@@ -0,0 +1,92 @@
...
+import rules
+
+from apps.documents import phases
+from tests.helpers import freeze_phase
+from tests.helpers import freeze_post_phase
+from tests.helpers import freeze_pre_phase
+from tests.helpers import setup_phase
+from tests.helpers import setup_users
+
+perm_name = 'meinberlin_documents.add_chapter'
```
kleingeist: this rule ended up in the wrong path: "tests/polls"

kleingeist:
i am not sure about using phases on tests of rules we are using only in the dashboard, where phases don't matter.
but maybe this is a more general problem, as it currently seems more logic to me to distinguish "user space rules" from "dashboard rules" with a prefix by their name

kleingeist:
@2e2a i merged this for now, even though i am not sure about how to check rules that are not connected to phases. maybe we can discuss that next week

PR #603, opened 2017-07-07
add private project form field and override label

vellip:


xi:
Before merging this, the following features need to be finished (or at least close):

- [ ] add users to private projects
- [ ] make sure private projects are not broken

PR #605, opened 2017-07-07
streamline static and assets directories

### LINKED ISSUE #252: Move non-generated files from static to assets
Currently we are mixing generated and custom content in /static
I propose to move all custom content (images, js, scss) to assets and copy them with webpack to static. 
This would enable us to completely ignore the /static/ folder ing git

### LINKED ISSUE #295: Static/Assets directories in apps
Django says to include app specific static files in a `/static` directory in the root of the particular app (https://docs.djangoproject.com/en/1.11/howto/static-files/). On `runserver` Django collects static files from the project wide static directory (mentioned in base.py) and the `/static` directories within all apps, grouping them into one global `/static` directory.
Now, we don't need `.jsx` files in that global `/static` directory since they are required and put together into one file by webpack. Still, those files need a directory to be put into on app level – therefore sometimes there might be an asset directory instead of a static directory or the other way around.
It'd be great to have one directory and in the best case that would include not having `.jsx` files in that django-generated, global `/static` directory.

vellip:
fixes #252 
fixes #295


.gitignore
```diff
@@ -18,8 +18,8 @@ media/
...
 /meinberlin/static/embed.js
 /meinberlin/static/fonts/
-/meinberlin/static/images/layers-2x.png
-/meinberlin/static/images/layers.png
-/meinberlin/static/images/spritesheet-2x.png
-/meinberlin/static/images/spritesheet.*
+/images/layers-2x.png
+/images/layers.png
+/images/spritesheet-2x.png
+/images/spritesheet.*
```
xi: Why not remove these?

xi:
LGTM.

webpack.config.js
```diff
@@ -101,7 +101,7 @@ module.exports = {
       {
         from: './meinberlin/assets/images/**/*',
         to: 'images/',
-        flatten: false
+        flatten: true
```
xi: What is the benefit?

.gitignore
vellip: I am not sure what those apply to, because there is no `images`directory in the root, I suppose?

webpack.config.js
vellip: It puts all files from all sub directories into the target without any additional directories below it.

xi: Yes, but why?

.gitignore
kleingeist: we could now ignore the whole static directory, couldn't we?

xi: we do.

kleingeist: ah sorry, i have been on an unmerged branch and wondered why git was showing the files

Issue #503, opened 2017-05-31
"I accept the terms of use" on signup page is not translated


xi:
A translation exists, but it is not used.

kleingeist:
Fixed with #523

Janaba:
problems still exist in dashboard:
Bebauungsplanverfahren
start date
end date
office worker email

Janaba:
on profile page there is a spelling mistake in the translation:

to follow or unfollow you should set a 'Haken' not a 'Hacken' :)

xi:
should all be fixed now.

PR #606, opened 2017-07-10
fix some i18n issues

xi:
see #503


kleingeist:
lgtm, would merge if @xi doesn't want to add other fixes to this branch

PR #612, opened 2017-07-11
2017 07 pv improve activity stream timestamps

vellip:


meinberlin/assets/scss/components/_action.scss
```diff
@@ -7,10 +7,27 @@
...
+    border: 1px solid $border-color;
+
+    @media (min-width: $breakpoint) {
+        padding: 1em;
+    }
+
+    h2 {
+        margin-top: 0;
+    }
+}
```
xi: I don't like having two components on a single DOM element. Can we change this to `.block--bordered`?

meinberlin/assets/scss/components/_action.scss
```diff
@@ -7,10 +7,27 @@
     font-size: $font-size * 100%;
 }
 
+.actions {
+    padding: 0;
```
xi: Having no padding looks strange IMHO.

apps/actions/assets/timestamps.js
```diff
@@ -0,0 +1,14 @@
...
+
+$(function () {
+  const $times = $('time.relative')
+
+  $times.each((i, e) => {
+    let datetime = moment($(e).attr('datetime'))
+
+    if (datetime.isAfter(moment().subtract(7, 'days'))) {
+      e.textContent = datetime.fromNow()
+    }
```
xi: Nice solution!

meinberlin/assets/scss/components/_action.scss
vellip: how so?

meinberlin/assets/scss/components/_action.scss
vellip: What two components?

xi: `block l-center-6 actions` -> `block` and `actions`

meinberlin/assets/scss/components/_action.scss
vellip: You were right, don't know what I was thinking there.

PR #613, opened 2017-07-11
prevent doubly id's from datetimepicker

### LINKED ISSUE #512: datetimepicker unique ids
screen shot 2017-06-01 at 15 59 21

vellip:
fixes #512


xi:
This fixes the issue. However, there is still a related issue: The `<label>` element has a `for` attribute that points to a non-existing id.

PR #614, opened 2017-07-11
Enable private projects and participants management

kleingeist:
missing:
- [x] view tests
- [x] tests for autofollow functionality
- [ ] activity stream view tests

Currently there are not that many places where private projects could be shown:
- [x] project list
- [x] user profile (profiles are not implemented)
- [ ] activity stream
- [x] notifications

Regarding notifications:
Currently the Follow object of a user will be deleted if she is removed as a participant of a project. 
But if a project which already has some users following it is set to private, the followers are kept.
Furthermore i have not tested if the API is safe in regard to private projects


apps/dashboard/templates/meinberlin_dashboard/includes/project_list_item.html
```diff
@@ -53,12 +53,22 @@ <h3 class="list-item__title">
...
             <a href="{% url 'dashboard-project-moderators' slug=project.slug %}"
                class="button button--light button--small">
-                <i class="fa fa-user" aria-hidden="true"></i>
+                <i class="fa fa-user-md" aria-hidden="true"></i>
                 {% trans 'Moderators' %}
             </a>
         </span>
         {% endif %}
 
+        {% if not project.is_public %}
```
vellip: I am not convinced of putting **another** button there.

vellip: Did you talk to @damusp about this?

apps/dashboard/mixins.py
```diff
@@ -123,6 +123,27 @@ def post(self, request, *args, **kwargs):
                         organisation_slug=self.organisation.slug)
 
 
+class DashboardUserRemovalMixin:
```
vellip: Can you not merge this one and the `ModRemovalMixin` into one? Maybe determine the action via the `submit_action` value, set in the submit button.

apps/dashboard/mixins.py
```diff
@@ -123,6 +123,27 @@ def post(self, request, *args, **kwargs):
...
+    related_users_field = 'participants'
+    success_message = _('User successfully removed.')
+
+    def post(self, request, *args, **kwargs):
+        if 'submit_action' in request.POST and (
+                request.POST['submit_action'] == 'remove_user'):
+            pk = int(request.POST['user_pk'])
+            user = get_object_or_404(User, pk=pk)
+            project = self.get_object()
+
```
vellip: Are you certain there is no check for permissions needed? Because for removing moderators it definitely was (at least at one point in time).

apps/dashboard/templates/meinberlin_dashboard/includes/project_list_item.html
```diff
@@ -53,12 +53,22 @@ <h3 class="list-item__title">
...
+        <span class="list-item__action">
+            <a href="{% url 'dashboard-project-participants' slug=project.slug %}"
+               class="button button--light button--small">
+                <i class="fa fa-user" aria-hidden="true"></i>
```
vellip: I'd suggest to make this `fa-users`, not user.

apps/dashboard/templates/meinberlin_dashboard/includes/project_list_item.html
kleingeist: yes :( 
but he didn't come up with another place and was hoping for dashboard 2.0

apps/dashboard/mixins.py
kleingeist: oh actually i just forgot to delete the mixins, they have both been moved to views.py as 
DashboardProjectUserListView and DashboardProjectModeratorsView

apps/dashboard/mixins.py
kleingeist: see above, the mixins are to be deleted.
but as the mixin had been used solely as a parent to the DashboardProjectModeratorsView, which did already check permission it was not required. but i am not sure where the moderator mixin had been used before.

kleingeist:
view tests are missing, but i am not sure, if we should do them :(

apps/dashboard/templates/meinberlin_dashboard/includes/project_list_item.html
vellip: I'd rather see it on the project detail below/in the list. Because that'll be the stuff you're gonna export anyway.

vellip: Merging Moderator and Participants into one view is not an option?

kleingeist: you would have to check back with @damusp and @Janaba but i'd propose to do it separately in the first iterator and redesign it with the new dashboard

apps/projects/views.py
```diff
@@ -105,4 +106,9 @@ class ProjectListView(filter_views.FilteredListView):
     filter_set = ProjectFilterSet
 
     def get_queryset(self):
-        return super().get_queryset().filter(is_draft=False)
+        return super().get_queryset().filter(
+            Q(is_public=False, participants__pk=self.request.user.pk) |
```
kleingeist: this will also show private projects, that are still in draft mode. is this intended?

vellip: Nope.

PR #617, opened 2017-07-11
Adapt offlineevents to ProjectContextDispatcher

kleingeist:
offline events had been forgotten to be adapted to the ProjectContextDispatcher due to parallel mergings.


apps/dashboard/mixins.py
```diff
@@ -23,6 +23,8 @@ def organisation(self):
         if 'organisation_slug' in self.kwargs:
             slug = self.kwargs['organisation_slug']
             return get_object_or_404(org_models.Organisation, slug=slug)
+        if hasattr(self, 'project'):
+            return self.project.organisation
         if 'slug' in self.kwargs:
```
xi: can we remove this case?

xi:
The general concept looks good to me. I did not check whether something was missed. @2e2a maybe you should have a look.

Marging now because this fixes a real-world bug.

PR #620, opened 2017-07-12
Update phase-indicator

Navigator28:
Instead of showing phases and offlineEvents on the timeline "above" moduls and OfflineEvents are shown. All phases of the moduls are shown below with the description.

There are problems with the links for the OfflineEvents which lead to an error. See #617 for the fix.


apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
```diff
@@ -32,28 +32,30 @@
...
 {% endif %}
+
 {% endwith %}
 
 {% if not object|is_offlineevent %}
-<div class="phase-info">
-    <div class="l-center-6">
-        {% if active_phase %}
+<div class="l-center-6">
+    {% for phase in project.phases %}
```
xi: This loops over all phases in a project. It should be limited to the current module.

xi:
Some next steps (@vellip, can you please look into this?)

- The phase popovers are gone now, so we can remove some related code.
- Styling needs to be improved, especially considering edge-cases (single phase, no offline events, …).
- `phase_indicator.html` and `.phase-list` should be renamed.

kleingeist:
also we have to remember to commit the new property `first_phase_start` on `Module` to the a4 core

kleingeist:
this PR depends on liqd/adhocracy4#131

Issue #610, opened 2017-07-10
Get genders straight (in de translation)


kleingeist:
Currently we are using the asterisk `*` on some of the translated texts. For example in
* "E-Mail Adresse der Sachbearbeiter*in"
* "Die Beschreibung verdeutlicht Nutzer*innen was das Ziel des Projekts ist und "
* ...

But are missing it on some other places:
* "Nutzer können Vorschläge mit Budgetangabe machen und erhalten durch "
* "Jeder Nutzer darf pro Frage nur eine Stimme abgeben."
* ... 

I think we should use the asterisk everywhere

xi:
Since meinBerlin is operated by the administration of Berlin, the guidelines ["Geschlechtergerechte Sprache in der Berliner Verwaltung"](https://www.berlin.de/sen/frauen/_assets/flyer_geschlechtergerechte_sprache.pdf) might be interesting.

vellip:
We will use "-innen und er".

xi:
There are quite some problematic constructs:

- compound nouns ("Nutzerkonto")
- articles ("Der Nutzer", similar: "Jeder Nutzer")
- large difference in size between german and english for short strings ("user" -> "Nutzerin oder Nutzer")

I tried to address these in 10f2abf2

fuzzylogic2000:
These are the rules, that we try to follow: 

-   use male form with compound nouns and articles ("Nutzerkonto", "Der Nutzer")
-   use * for short strings ("Nutzer*in")
-   use two genders for long sentences ("Nutzerinnen und Nutzer können …")

@CarolingerSeilchenspringer @hmkay @phillimorland @MagdaN Should we close this issue for now or keep it open as a reminder?

hmkay:
hmm...probably @CarolingerSeilchenspringer should have the last word here, but I´d say close it for now...

PR #623, opened 2017-07-17
update translations

xi:


vellip:
Not sure about the gendering.

xi:
> Not sure about the gendering.

Me neither. I see that as a separate issue (see #610). For now I tried to be consistent with existing translations.

2e2a:
lgtm

PR #626, opened 2017-07-18
change timeline styling

xi:
We git the feedback that with the current design users do neither understand that the timeline is a timeline nor that the entries can be clicked. This new design is supposed to improve that.

Note that the conditions for `is-active` and `is-selected` do not yet work as intended.

### DIFF
--- a/apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
+++ b/apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
@@ -13,7 +13,7 @@
                 <li>
                     <a
                         href="{{ project.get_absolute_url }}"
-                        class="button button--light button--small timeline__button {% if obj == active_phase.module %}is-active{% endif %} {% if obj == object %}is-selected{% endif %}"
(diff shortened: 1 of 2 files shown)


apps/projects/templates/meinberlin_projects/includes/phase_indicator.html
```diff
@@ -13,7 +13,7 @@
                 <li>
                     <a
                         href="{{ project.get_absolute_url }}"
-                        class="button button--light button--small timeline__button {% if obj == active_phase.module %}is-active{% endif %} {% if obj == object %}is-selected{% endif %}"
+                        class="button button--light button--small timeline__button {% if obj == active_phase.module %}is-active{% endif %} {% if not object|is_offlineevent %}is-selected{% endif %}"
```
vellip: Let's hope we remember that hack when we introduce multiple modules.

xi: I just created a ticket specifically for that.

PR #627, opened 2017-07-18
Fix project dispatcher for bplan projects

2e2a:
CreateViews provide the a failing get_object method which should not be called when validating the object project.


kleingeist:
i don't like that we have to take care of that ourself, but it will fix our problems so i'll merge

PR #629, opened 2017-07-19
phase-info and offline-event styling

vellip:


apps/offlineevents/templates/meinberlin_offlineevents/offlineevent_detail.html
```diff
@@ -8,6 +8,18 @@
...
+                    <h3 class="phase-info__title">
+                        {% trans 'Offline-Event' %}
+                    </h3>
+                    <div class="phase-info__subtitle">
+                        {{ object.date|date:'DATE_FORMAT' }}
+                    </div>
+                    <div class="phase-info__description">
+                        {% trans 'Here you can read the documentation on what happened during this Offline-Event.' %}
+                    </div>
+                </div>
```
xi: The equivalent of this code is in `timeline.html`. We should either move this there or move the other code to an equivalent module file.

apps/offlineevents/templates/meinberlin_offlineevents/offlineevent_detail.html
```diff
@@ -8,6 +8,18 @@
 
             <div class="item-detail">
 
+                <div class="phase-info">
+                    <h3 class="phase-info__title">
+                        {% trans 'Offline-Event' %}
+                    </h3>
+                    <div class="phase-info__subtitle">
+                        {{ object.date|date:'DATE_FORMAT' }}
```
xi: Now that the date is shown here, it should be removed below.

apps/offlineevents/templates/meinberlin_offlineevents/offlineevent_detail.html
```diff
@@ -8,6 +8,18 @@
...
+                    </div>
+                    <div class="phase-info__description">
+                        {% trans 'Here you can read the documentation on what happened during this Offline-Event.' %}
```
xi: I guess we have to distinguish whether the event is in the past or future and display different texts.

apps/projects/templates/meinberlin_projects/includes/timeline.html
```diff
@@ -35,25 +35,48 @@
...
+                        {% trans "Participation is not possible at the moment." %}
+                    </h3>
+                {% endfor %}
+            </div>
+            <button class="button button--link" type="button" data-toggle="collapse" data-target="#phase-overview" aria-expanded="false" aria-controls="phase-overview">
+                <span class="u-if-collapsed">{% trans 'Show all phases' %}</span>
+                <span class="u-if-not-collapsed">{% trans 'Hide all phases' %}</span>
+            </button>
+        {% endif %}
+    </div>
```
xi: The case where there is not `active_phase` is missing. We need to be careful with wording in that one because it can be before, after, or even in between phases.

@damusp suggested to automatically generate a text that summarizes the process so far. But I think this is out of scope for now.

meinberlin/assets/scss/components/_timeline.scss
```diff
@@ -118,22 +119,49 @@ $circle-size: 0.6em;
 }
 
 .phase-info {
+    box-shadow: 0 0 1.25em 0 rgba(0, 0, 0, 0.2);
```
xi: The button looks much smaller than in the design. I think we can increase it to `2em`.

meinberlin/assets/scss/components/_timeline.scss
```diff
@@ -118,22 +119,49 @@ $circle-size: 0.6em;
...
+    border-radius: $border-radius;
     margin-bottom: $spacer * 2;
```
xi: The margin looks much smaller than in the design. I think we can increase it to `$spacer * 3` and also apply it to the top (in order to leave enough space for the shadow).

meinberlin/assets/scss/components/_timeline.scss
```diff
@@ -118,22 +119,49 @@ $circle-size: 0.6em;
...
+.phase-info__overview {
+    margin-top: $spacer * 2;
+
+    .phase-info__title {
+        font-size: 1em;
+    }
+
+    .phase-info__subtitle {
+        margin-bottom: 0;
+    }
```
xi: Can we avoid descendant selectors here?

meinberlin/assets/scss/components/_timeline.scss
```diff
@@ -118,22 +119,49 @@ $circle-size: 0.6em;
...
-    background-color: $bg-tertiary;
-    border-left-color: $brand-primary;
+.phase-info__active {
+    margin-bottom: $spacer;
 }
 
 .phase-info__title {
-    font-size: $font-size-xl;
+    font-size: $font-size-lg;
+    font-weight: bold;
```
xi: regular headings are not bold, so these should be neither in order to not look overly important

xi:
replaced by #631

PR #631, opened 2017-07-19
change phase info design

xi:
this builds on/replaces #629

@damusp and me performed a little user testing which prompted us to change some things in the phase info.

## Test results

The test setup was created from a "agenda setting" blueprint with one additional event ("assembly") that would take place at the end of the online participation. This way we could test the full timeline + phase info UI.

-   In general the test person found most information easily. However, they misinterpreted some UI components which might lead to confusion during further usage.
-   The test person was confused by the phase info on the offline event detail page (felt like duplicate information) and failed to find the date of the event.
-   They did not understand that the timeline shows a time-sequence of modules. However, they did understand that it can be used for navigation.
-   They were confused about the idea list and mistook the name of the current phase ("Collect ideas") as a heading explaining it.
-   They expected information about the process (phases, events, …) in the "information tab".
-   They were confused about why the phases are listed in the "online participation" tab, but the date of the event in the "assembly" tab. They expected this information in a single place. The module/event hierarchy was not clear.

## Changes

-   We removed the phase info from the offline event detail page.
-   We removed the active phase from the module page and show module information (for now project information) instead. The styling uses `.item-detail` like on the offline event page.

## Rationale

-   The relation between timeline button and page content is clearer now because the button text is reused in the heading.
-   The relation between heading and idea list is clearer now because this now actually *is* a heading for the content and not tangential information like phases.
    -   Phases are tangential because they do not change which content is displayed, but only what you can do with it. This is different e.g. for modules.
-   The relation between heading and phase overview is clearer now because this now actually *is* a heading for the phase overview, not a duplication of some of the overview's content.
-   The phases are less prominent, but still fully available (collapsed by default)
-   The most important phase information ("participation is not possible right now, but will be on {date}") is displayed prominently.

---

-   We do not feel like the timeline needs to be recognizable as such.
    -   Two strongly time-related components on the page (timeline + phase overview) could be confusing.
    -   The ordering is clear. Many people might implicitly assume that it is ordered chronologically anyway.
    -   The test person did not have any issues interacting with it.
-   We do think that initiators should include a process overview in the information tab.
    -   We should think about ways to incentivise this.

## Edge Cases

-   This design works well for large numbers of modules/events because the current module/event is clearly displayed
-   This design works well for any number of phases because the phase overview is displayed in a simple and constistent UI across all modules.
-   This design does not yet work well for cases without a timeline (single module + no events) because it adds another heading that is useless in that case. (The currently hard-coded heading "Online participation" is especially useless for these cases.) @damusp, @vellip can you please think about a solution for this?


meinberlin/apps/projects/templates/meinberlin_projects/includes/timeline.html
```diff
@@ -14,7 +15,7 @@
                         href="{{ project.get_absolute_url }}"
                         class="button button--light button--small timeline__button {% if obj == active_phase.module %}is-active{% endif %} {% if not object|is_offlineevent %}is-selected{% endif %}"
```
kleingeist: `{% if not object|is_offlineevent %}is-selected{% endif %}` is only working as long as there is no second module.
otherwise both modules will be marked `is-selected`
currently this is no issue, but maybe we should add a comment/fixme

xi: There are several places in the code that are not yet ready for multi-modules. This is also because there are not yet sufficient interfaces. I have a general outline of the necessary steps in my head.

kleingeist:
lgtm, i did a visual and functional test of cornes cases (long titles, many events, pre-, post-, active phase...) and they did work quite nice.

i just felt, that the phases "timeline" is a little to dense. i'd prefer a little bit more space between the title, date and description (see the screenshot) but i would still merge
dense-phases

PR #634, opened 2017-07-24
Add newsletter functionality

kleingeist:
Missing
- [x] test
- [x] decide on CK input (h1 vs bold)

Depends on liqd/adhocracy4#113


Navigator28:
We should not forget to use the sender name and the sender mail in the actual email.

kleingeist:
Users that are registered, but not activated receive newsletters

kleingeist:
line-height in newsletter_base.html should be set to something else then 2px

meinberlin/apps/newsletters/views.py
```diff
@@ -92,6 +94,10 @@ def form_valid(self, form):
             emails.NewsletterEmail.send(instance,
                                         participant_ids=list(participant_ids),
                                         **self.get_email_kwargs())
+            messages.success(_('Newsletter has been queued to be send.'))
```
vellip: I don't think the initiator has to know it's a queue, wouldn't be sure how to translate it either and would prefer something as simple as "The newsletter has been saved and will be sent within the next minutes." Or whatever time unit is appropriate.

2e2a:
lgtm

PR #640, opened 2017-07-25
Show 404 error if the document has no chapters yet

kleingeist:


meinberlin/apps/documents/views.py
```diff
@@ -45,10 +47,14 @@ def chapter_list(self):
 class DocumentDetailView(ChapterDetailView):
 
     def get_object(self):
-        return models.Chapter.objects\
-            .filter(module=self.project.last_active_module)\
+        first_chapter = models.Chapter.objects \
+            .filter(module=self.project.last_active_module) \
```
xi: Why add whitespace here?

2e2a:
lgtm

meinberlin/apps/documents/views.py
kleingeist: not sure why that happend. i think pycharm did it form me.

PR #642, opened 2017-07-25
Fix content creation before module active

2e2a:


kleingeist:
lgtm. 
every other case should be routed via PhaseDispatcher.
If a user guesses a valid slug of a not published item yet, she will get a ServerError which is okay until we rewrite multi modules

Issue #633, opened 2017-07-21
Hard to make out individual map pins


damusp:
Overlapping pins blend in with each other. Clusters of pins look like one big mess.

<img width="220" alt="bildschirmfoto 2017-07-21 um 16 10 10" src="https://user-images.githubusercontent.com/22769111/28467250-89cc4092-6e2f-11e7-8467-f3c011aad071.png">

Ideas to solve this:
- adding borders to the pins 
- giving a hover style to the pins

xi:
Just provide us with new marker images that have a border.

damusp:
I've sent these to @Navigator28 

![pin_default](https://user-images.githubusercontent.com/22769111/28514636-0c3a6068-705b-11e7-88d9-83e9d76e759a.png)

![pin_hover](https://user-images.githubusercontent.com/22769111/28514637-0c3b889e-705b-11e7-9a10-3e105b4af272.png)

xi:
The fix is not merged just yet!

PR #643, opened 2017-07-25
Change map markers

Navigator28:
Depends on https://github.com/liqd/adhocracy4/pull/142
See:  #633 
The code won't function without the changes in the core because the shadow image is deleted.
- Images have been replaced/deleted


xi:
LGTM

PR #646, opened 2017-07-25
better highlighting of active module

xi:


meinberlin/apps/projects/templates/meinberlin_projects/includes/timeline.html
```diff
@@ -18,7 +18,7 @@
                         {% trans 'Online participation' %}
                     </a>
                     {% if obj == active_phase.module %}
-                        <span class="timeline__active">{% trans 'active' %}</span>
+                        <span class="timeline__active" aria-label="">{% trans 'active' %}</span>
```
xi: This looks wrong…

meinberlin/apps/projects/templates/meinberlin_projects/includes/timeline.html
```diff
@@ -28,6 +28,7 @@
                         class="button button--light button--small timeline__button {% if obj == object %}is-selected{% endif %}"
                     >
                         {{ obj.name }}
+                        {% if obj == object %}<span class="sr-only">{% trans 'current page' %}</span>{% endif %}
```
xi: This also needs to appear in  the case that `obj|is_module`

Issue #641, opened 2017-07-25
Project tile gradient should be more subtle


damusp:
The current gradient works well for photographic images but not for graphic images. 
Please tone it done a bit (maybe opacity around 0.2 instead of 0.8) 

<img width="626" alt="bildschirmfoto 2017-07-25 um 10 00 41" src="https://user-images.githubusercontent.com/22769111/28561876-6c280e60-7120-11e7-854d-133d02315b52.png">

PR #647, opened 2017-07-25
Improve style of tile images

Navigator28:
See: #641 
- [x] opacity
- [ ] new background color (white) for transparent parts of the picture/logo


xi:
> new background color (white) for transparent parts of the picture/logo

I don't think this is necessary. This is already useful in its current form. Merging…

PR #649, opened 2017-07-25
user activity stream

xi:
This adds an activity stream to the user dashboard. The purpose is *not* to be a public record of a user's activities, but to allow users to see what happened in the projects (or other resources) they follow.

Note:

-   Someone should double check that no private information is leaking via actions (e.g. private projects).
-   If I am not logged in the page currently throws an error. I did not find out why.
-   The UI still needs some work.
-   Filtering actions by follows could be moved to core. But it will also get more complicated once it is possible to follow resources other than projects.

### DIFF
--- a/meinberlin/apps/account/templates/meinberlin_account/actions.html
+++ b/meinberlin/apps/account/templates/meinberlin_account/actions.html
@@ -1,0 +1,17 @@
+{% extends "meinberlin_account/account_dashboard.html" %}
+
+{% load i18n %}
+
+{% block title %}{% trans 'Activities' %} &mdash; {{ block.super }}{% endblock %}
+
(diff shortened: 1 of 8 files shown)


meinberlin/apps/account/templates/meinberlin_account/actions.html
```diff
@@ -6,6 +6,7 @@
 
 {% block dashboard_content %}
     <h1>{% trans 'Activities' %}</h1>
+    <p>{% trans 'Here you can find any actions that have happened in projects you are following.' %}</p>
```
xi: I personally would prefer "This is a list of actions" instead of "Here you can find any actions"

PR #650, opened 2017-07-26
Fixup 019aac2: Also add selected page text to module case

xi:
see https://github.com/liqd/a4-meinberlin/pull/646#discussion_r129511859


vellip:
Thanks

PR #652, opened 2017-07-28
pollbar animation

### LINKED ISSUE #347: Animation for polls
It would be really cool to have a build-up animation for the graph in our polls. 

A transition that builds up the bars from left to right would not just be pretty, it would also create a better sense for what's happening.

vellip:
fixes #347


xi:
looks nice indeed.

PR #660, opened 2017-08-02
initiator request

xi:
This adds a simple form at http://localhost:8000/initiators/request/ which can be used by regular users to request initiator permissions. An email will be send to `settings.CONTACT_EMAIL`.

Note that currently there is only an informal `organisation_name` field. We could change this to a select box of all existing organisations. This would provide us with enough information to allow admins to accept the request with a single click of a button. I am not sure though whether that is desirable/required.

### DIFF
--- a/meinberlin/apps/initiators/templates/meinberlin_initiators/emails/initiator_request.de.email
+++ b/meinberlin/apps/initiators/templates/meinberlin_initiators/emails/initiator_request.de.email
@@ -26,4 +26,4 @@ Datum der Nutzerregistrierung auf meinBerlin
 {% endblock %}
 
 {% block cta_label %}Zur Administration{% endblock %}
-{% block cta_url %}https://mein.berlin.de/django-admin{% endblock %}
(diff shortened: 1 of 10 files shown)


meinberlin/apps/initiators/templates/meinberlin_initiators/emails/initiator_request.de.email
```diff
@@ -0,0 +1,29 @@
...
+Telefonnummer
+{{ phone }}
+
+Datum der Nutzerregistrierung auf meinBerlin
+{{ user.date_joined }}
+
+{% endblock %}
+
+{% block cta_label %}Zur Administration{% endblock %}
+{% block cta_url %}https://mein.berlin.de/django-admin{% endblock %}
```
kleingeist: better use `{{ email.get_host }}{% url 'admin:index' %}`

kleingeist:
i added the use of dynamic/named urls and styled the form with the mein berlin templates.
i will merge to be able to test it on stage

PR #661, opened 2017-08-02
Add free text filter to topic prioritization projects

Navigator28:
Adds a free text filter to topic prioritization projects so you (as a user) are able to search for specific topics.


kleingeist:
lgtm

PR #662, opened 2017-08-02
Django 1.11 compatibility

kleingeist:
depends on liqd/adhocracy4#141

Known Problems
- [x] Moderator Feedback is not working anymore
- [x] multiform has to be pointed to a PR that adds 1.9 compatibility (bmispelon/django-multiform#13)


xi:
Merging now. This way we can do extensive testing and fix issues as we find them.

PR #664, opened 2017-08-03
add project duplicate

vellip:
This adds the feature to duplicate projects.


kleingeist:
@vellip  do you known if moderators and members should be duplicated, too?

kleingeist:
lgtm

PR #665, opened 2017-08-03
Improve print layout

Navigator28:
- Hides unnecessary information/fields in the print document.
- Adds page breaks


xi:
I am not sure whether it is a good idea to have all print related styling in a single place. We leave other media queries with the components they belong to. Both are valid approaches, I am just not sure which I like better.

vellip:
I think because the code is pretty specific to one media only and would have to be repeated over and over again (see the display: none) it's nicer to have it in one spot.

xi:
I removed some styles in f26d326. I am not generally opposed to them, but I would like to merge the rest now and think about the removed ones a bit longer.

meinberlin/assets/scss/components/_comments.scss
```diff
@@ -55,6 +63,10 @@
 .comment {
     margin-top: $spacer * 2;
     margin-bottom: $spacer * 3;
+
+    @media print {
+        margin-bottom: 1em;
+    }
```
xi: I am not completely sure why this was added.

PR #666, opened 2017-08-04
fix search styling for safari

vellip:


xi:
The styling in chrome does not seem to be affected.

PR #669, opened 2017-08-07
attempt to fix flex-box issue for IE

vellip:
This is more a basis for discussion than a proposal. Flexbox is behaving rather weird on IE, the original problem is, that IE11 translates `flex: 1` differently than other browsers (https://github.com/philipwalton/flexbugs#6-the-default-flex-value-has-changed). However, just using the longer hand syntax doesn't fix the issue because the user agent seems to have a problem with box-sizing and width calculations in general (either both the input and the button are too short or the input is longer than it should be and hides the button behind the next dropdown). We also can't use `flex-basis: auto` because we have set all form elements to 100% width which will also apply to those and create some overlapping of dropdowns (mostly on IE, though, too).

### DIFF
--- a/meinberlin/templates/a4filters/widgets/free_text_filter.html
+++ b/meinberlin/templates/a4filters/widgets/free_text_filter.html
@@ -7,7 +7,9 @@
             <input type="hidden" name="{{ par_name }}" value="{{ value }}">
         {% endif %}
     {% endfor %}
-    <button class="input-addon__after button button--light" type="submit" title="{% trans 'Search' %}">
-        <i class="fa fa-search" aria-label="{% trans 'Search' %}"></i>
-    </button>
+    <div class="input-addon__after">
+        <button class="button button--light" type="submit" title="{% trans 'Search' %}">
+            <i class="fa fa-search" aria-label="{% trans 'Search' %}"></i>
+        </button>
+    </div>
 </form>
--- a/meinberlin/assets/scss/_form.scss
+++ b/meinberlin/assets/scss/_form.scss
@@ -117,12 +117,13 @@ select {
 }
 
 .input-addon__input {
-    flex: 1;
+    flex-grow: 1;
 }
 
 .input-addon__before,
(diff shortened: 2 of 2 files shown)


meinberlin/templates/a4filters/widgets/free_text_filter.html
```diff
@@ -7,7 +7,9 @@
...
         {% endif %}
     {% endfor %}
-    <button class="input-addon__after button button--light" type="submit" title="{% trans 'Search' %}">
-        <i class="fa fa-search" aria-label="{% trans 'Search' %}"></i>
-    </button>
+    <div class="input-addon__after">
+        <button class="button button--light" type="submit" title="{% trans 'Search' %}">
+            <i class="fa fa-search" aria-label="{% trans 'Search' %}"></i>
+        </button>
+    </div>
```
xi: This breaks the border radius.

xi:
The spec recommends to use the flex shorthand:

> Authors are encouraged to control flexibility using the flex shorthand rather than with its longhand properties directly, as the shorthand correctly resets any unspecified components to accommodate common uses.
> -- https://drafts.csswg.org/css-flexbox-1/#flex-components

Furthermore, in order to avoid browser incompatibilities, I would always specify all 3 values.

I tested this issue in IE11, edge, and chrome. I found three versions that somehow work. All of them use   `flex: 1 1 auto` on `.input-addon__input`.

1. Have `flex: 0 1 auto` on `.input-addon__after` and have that on a separate div (not the button). Issue: The button still has round borders on the left (in all browsers).
2. Have `flex: 0 1 0%` on  `.input-addon__after`. Issue: The button is a bit squashed (only IE).
3. Have `flex: 0 1 auto; width: auto !important` on `.input-addon__after`. Issue: `!important`.

Most notable, I found that we do not need to have both separate elements (1) and `!important` (3) as it is done in this PR.

My order of preference is 2, 3, 1.

PR #670, opened 2017-08-07
mobile layout

vellip:
After talking to @damusp about it, we found it's nicer to have the `.l-center-6` 100% wide on mobile. Otherwise it looks more like a bug than desired.


meinberlin/apps/cms/templates/meinberlin_cms/blocks/projects_block.html
```diff
@@ -1,7 +1,7 @@
 {% load rules i18n %}
 <div class="block block--projects">
     <h2>{{ value.title }}</h2>
-    <ul class="l-tiles-4">
+    <ul class="l-tiles-4 u-list-reset">
```
xi: Shouldn't this better be fixed in `.l-tiles-4`?

PR #671, opened 2017-08-07
Send wagtail form mails synchronously

2e2a:
Temporary fix


kleingeist:
lgtm.
but we should not forget to fix it and provide serializable arguments.

PR #675, opened 2017-08-08
Send wagtail form email asynchronously

### LINKED ISSUE #672: Send wagtail form emails asynchronously
Currently wagtail form emails are sent synchronously because the `send` attributes are not serializable. see #671 for the temporary fix.
Rather then sending sync we should fix the problem and provide the required data as a serializable type

2e2a:
Fixes #672


kleingeist:
lgtm

PR #679, opened 2017-08-10
drop moment.js, replace with timeago.js

vellip:


kleingeist:
lgtm

meinberlin/apps/actions/assets/timestamps.js
```diff
@@ -1,14 +1,17 @@
-const moment = require('moment')
+const timeago = require('timeago.js')
+require('timeago.js/src/timeago.locales.js')
```
xi: We don't really need all locales. It would be sufficient to just load german as we did with moment.

PR #682, opened 2017-08-11
dashboard2 nav

vellip:
based on #681 
items still need some styling


xi:
I would take a different approach here: `.dashboard-nav` is used in the old dashboard and must therefore stay as it is for backwards compatibility. I would go with `.tab` here and some additional layout.

PR #686, opened 2017-08-14
copy newsletter and organisation settings to dashboard2

xi:


xi:
These views will be plugged into the dashboard with a different mechanism. Closing.

PR #689, opened 2017-08-15
Update to django 1.11 middlewares and disable debug toolbar

kleingeist:
* Update to django 1.11 middlewares (see https://docs.djangoproject.com/en/1.11/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware)
* Disable django debug toolbar until wagtail 1.12 is released (see jazzband/django-debug-toolbar#950 for reference)


xi:
LGTM

PR #691, opened 2017-08-16
implement progress indicator (circle)

xi:
This uses the SVG method described in https://www.smashingmagazine.com/2015/07/designing-simple-pie-charts-with-css/

Note that SVG scaling is broken in IE, so we may have to write some compatibility code for that.


xi:
i added some more features of the project indicator, including the "publish" button. Note that I just did the UI and there is no functionality behind that yet.

PR #692, opened 2017-08-16
Add basic component registry and dispatcher

kleingeist:
The registry is named "contents" for the sake of consistency with phases (and modules)

### DIFF
--- a/meinberlin/apps/dashboard2/__init__.py
+++ b/meinberlin/apps/dashboard2/__init__.py
@@ -1,0 +1,3 @@
+from .contents import content
+
+__all__ = ['content']
--- a/meinberlin/apps/dashboard2/__init__.py
+++ b/meinberlin/apps/dashboard2/__init__.py
@@ -1,3 +1,5 @@
 from .contents import content
+from .components import DashboardModuleComponent
+from .components import DashboardProjectComponent
 
-__all__ = ['content']
+__all__ = ['content', 'DashboardModuleComponent', 'DashboardProjectComponent']
--- a/meinberlin/apps/dashboard2/components.py
+++ b/meinberlin/apps/dashboard2/components.py
@@ -9,6 +9,10 @@ class DashboardComponent:
     - identifier (defaults to "app_label:label")
       - has to be unique
       - may only contain alphanumeric characters and -, _, :
+
+    Required methods:
+    - get_view(): view function
+      Return a view function which takes menu and project/module as kwargs
     """
 
     app_label = None
@@ -23,6 +27,9 @@ class DashboardComponent:
             app_label=self.app_label,
             component_label=self.label,
         )
+
+    def get_view(self):
+        pass
 
 
 class DashboardProjectComponent(DashboardComponent):
--- a/meinberlin/apps/dashboard2/components.py
+++ b/meinberlin/apps/dashboard2/components.py
@@ -1,0 +1,55 @@
+class DashboardComponent:
+    """Abstract interface for dashboard components.
+
+    Required properties:
+    - app_label
+
+    Required properties with default implementation:
+    - label (defaults to the lower class name)
+    - identifier (defaults to "app_label:label")
+      - has to be unique
+      - may only contain alphanumeric characters and -, _, :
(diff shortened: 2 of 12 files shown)


meinberlin/apps/dashboard2/contents.py
```diff
@@ -1,5 +1,27 @@
...
+
+    def __getitem__(self, identifier):
+        component = self._registry['project'].get(identifier, None)
+        if not component:
+            component = self._registry['module'].get(identifier)
+        return component
+
+    def __contains__(self, identifier):
+        return (identifier in self._registry['project'] or
+                identifier in self._registry['module'])
```
xi: I don't like to treat these two dictionaries as one. I would prefer to have `get_project_component(self, identifier)` and  `get_module_component(self, identifier)` instead.

meinberlin/apps/dashboard2/components.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+
+
+class DashboardProjectComponent(DashboardComponent):
+    """Abstract interface for project components.
+
+    Inherits from DashboardComponent.
+
+    Required methods: str | None
+    - get_menu_item(project):
```
xi: `str | None` should probably be in the line below.

meinberlin/apps/dashboard2/components.py
```diff
@@ -0,0 +1,55 @@
...
+      Return a localized string if the component should be rendered for the
+      project passed. Return None otherwise
+    """
+
+    def get_menu_item(self, project):
+        pass
```
xi: I don't really see the point of having `DashboardProjectComponent` and `DashboardModuleComponent`. I think just `DashboardComponent` is sufficient. I also think that the interfaces of the two cases will stay identical.

xi: I would like to rename `get_menu_item` to `get_menu_label` or `get_menu_text` or something.

meinberlin/apps/dashboard2/templates/meinberlin_dashboard2/base_dashboard_project.html
```diff
@@ -9,12 +9,20 @@
...
+                            </li>
+                        {% endfor %}
+                    {% else %}
                     <li class="dashboard-nav__page">
                         <a href="#" class="dashboard-nav__item dashboard-nav__item--interactive is-active">Grundeinstellungen</a>
                     </li>
                     <li class="dashboard-nav__page">
                         <a href="#" class="dashboard-nav__item dashboard-nav__item--interactive">Information</a>
                     </li>
+                    {% endif %}
```
xi: The example code should be removed.

meinberlin/apps/dashboard2/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .contents import content
+
+__all__ = ['content']
```
kleingeist: this is to be consistent with phases.content and allows us to 
`from adhocracy4.phases import content` instead of `from adhocracy4.phases.content import content`
resp 
`from meinberlin.dashboard2 import content` instead of `from meinberlin.dashboard2.content import content`

meinberlin/apps/dashboard2/contents.py
kleingeist: i wasn't sure either, but it can be so convenient to use this method as identifiers are defined to be unique anyway. it would also be possible to store the _registry as a dict of {identifier: ('project', component} 
then the cost of get_project_components would be higher but the general registry would be consistent

meinberlin/apps/dashboard2/components.py
kleingeist: it's mainly for documentation. with the two interfaces it is always clear what to expect and what variables are present. furthermore we could check in the ProjectDispatcher if the component really is a ProjectComponent (while this could also be done with separate get_project_component and get_module_components)

but mostly i don't like method signatures like:
`def get_menu_item(self, something)`

xi:
Merging now. Refactoring can be done separately.

PR #694, opened 2017-08-16
change project detail styling

xi:
this pull request changes some layout and styling. It also brings back the project short description.


vellip:
Upon comparing, I noticed there is a grey-ish background on the tabs in the design. It's just a minor thing, but I think it makes a difference in the design.

xi:
@damusp and me decided to not have the gray background for now because it does not work so well if there is no header image.

vellip:
LGTM otherwise!

PR #695, opened 2017-08-16
fix tablist after alert message

xi:
there was a margin that did not belong


kleingeist:
lgtm after a visual test

PR #696, opened 2017-08-17
Add dashboard error count and set form fields to required for published projects

kleingeist:
This PR introduces a `get_progress(project_or_module)` method on the DashboardComponent Interface. 
This get_progress is used to show error notes in the menu and display the progress bar. 
During the development specialized abstract Components for forms were introduces.

The progress / error counting has to be done on the models as it is not possible to reliably convert a model object to a POST data dict.
For example: phase.start_date may be split into two fields by the datetimepicker widget. Thus two separate values are expected in the POST data dict, which will be merged after the initial validation. But fields/widgets are not holding any information about the expected POST data fields.

### DIFF
--- a/meinberlin/apps/dashboard2/forms.py
+++ b/meinberlin/apps/dashboard2/forms.py
@@ -49,7 +49,61 @@ class ProjectUpdateForm(forms.ModelForm):
         }
 
 
-class ProjectBasicForm(forms.ModelForm):
+def _make_fields_required(fields, required):
+    if required:
+        for name, field in fields:
+            if required == '__all__' or name in required:
+                field.required = True
+
+
+class ProjectDashboardForm(forms.ModelForm):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if not self.instance.is_draft:
+            _make_fields_required(self.fields.items(),
+                                  self.get_required_fields())
+
+    @classmethod
+    def get_required_fields(cls):
+        meta = getattr(cls, 'Meta', None)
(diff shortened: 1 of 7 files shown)


meinberlin/apps/dashboard2/forms.py
```diff
@@ -49,7 +49,61 @@ class Meta:
         }
 
 
-class ProjectBasicForm(forms.ModelForm):
+def _make_fields_required(fields, required):
+    if required:
+        for name, field in fields:
+            if required == '__all__' or name in required:
+                field.required = True
```
kleingeist: >  Is it officially supported?
i'd say yes. the properties are public (non underscored) and from a OOP perspective it is definitively valid to change an objects state by changing its public properties. 

> Are we sure that the same field instance is not reused somewhere else?
yes, the method is invoked with the fields from a concrete form object. this object has it's own list of form fields which it can edit in any way it wants. see the comment at the django code:
https://github.com/django/django/blob/master/django/forms/forms.py#L91

meinberlin/apps/dashboard2/forms.py
```diff
@@ -104,26 +49,102 @@ class Meta:
...
+        return getattr(meta, 'required', [])
+
+
+class ProjectBasicForm(ProjectDashboardForm):
 
     class Meta:
         model = project_models.Project
         fields = ['name', 'description', 'image', 'tile_image', 'is_archived',
                   'is_public']
+        required = '__all__'
```
xi: `required` is a very general name. Maybe change it to something like `required_for_project_publish`.

PR #697, opened 2017-08-21
add egg for multiform

vellip:
Pip apparently needs an egg for vcs. https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support


kleingeist:
i  hop we will get rid of multiform soon :)

PR #700, opened 2017-08-21
update npm packages

xi:
most notably update to bootstrap 4-beta


vellip:
This breaks our JS because we haven't included Popper.js which replaces Tether.js I wasn't able to fix it right away, I had some strange errors with Popover still expecting Tether and later it missing some Util and Tooltip modules.

xi:
I managed to fix some of the issues with bootstrap, but in the end I did not manage to fix tabs. So I reverted that part for now. See 7cf66528 for my current state.

vellip:
We can handle bootstrap separately.

PR #701, opened 2017-08-21
Add topicprio dashboard component with multiple  urls

kleingeist:
The interface is extended to include a get_urls method which has to return a triple of (pattern, view, name).
TopicPrio is used as an example to test the functionality.


meinberlin/apps/dashboard2/mixins.py
```diff
@@ -51,19 +52,8 @@ def blueprint_key(self):
...
-        self.project = project
-        return super().dispatch(request, *args, **kwargs)
-
-    def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        context['project'] = self.project
-        context['module'] = self.module
-        return context
+class DashboardComponentMixin(ProjectContextDispatcher):
+    component = None
```
kleingeist: otherwise we would have to copy n paste the whole logic. ProjectContext is a very general function which does try to find project and module out of many possible configurations.

meinberlin/apps/dashboard2/components.py
```diff
@@ -9,9 +9,9 @@ class DashboardComponent:
 
     Required properties with default implementation:
     - label (defaults to the lower class name)
-    - identifier (defaults to "app_label:label")
+    - identifier (defaults to "app_label-label")
       - has to be unique
-      - may only contain alphanumeric characters and -, _, :
+      - may only contain alphanumeric characters, _ and -
```
xi: Should be "_, and -" (the english rule is different than the german one here).

xi:
I think I still like the idea to separate the main view for a component (the one that is linked to in the menu) from the additional URLs. Currently, the main view is defined as the first one from `get_urls()`. The default case is to implement `get_urls()` and make it return a list with a single entry. Can we switch it: By default you only implement `get_base_url()` and if you need more URLs you also implement `get_additional_urls()`?

meinberlin/apps/dashboard2/mixins.py
xi: I am not questioning whether these classes should inherit. I am just wandering about the names of these classes since the hierarchy is typically `Dashboard -> Project, Module`. Maybe the inheriting class should be called `ProjectDashboard*`.

kleingeist:
> I think I still like the idea to separate the main view for a component (the one that is linked to in the menu) from the additional URLs. This way the default case would be cleaner.

i'm not a fan of this as it brings back confusion as why there are two methods for the same thing.
this way we have a clean interface. the component has to define the urlpatterns it want's to use (as all apps in django have to) and it has to be "resolvable" to the a certain url string (as model objects in django are with get_absolute_url). we keep the interface very close to django and thus help developers.

Furthermorer adding a get_addition_urls method would mean we'd have to implement more logic to distinguish the  two slightly different cases. which would make our code more complicated.


> Currently, the main view is defined as the first one from get_urls(). The default case is to implement get_urls() and make it return a list with a single entry. Can we switch it? By default you only implement get_base_url() and if you need more URLs you also implement get_additional_urls()?

we would then change the behaviour of get_base_url which is (in accordance to model.get_absolute_url)  returning a url string. instead we would have to return a triple of url, view and name. 
changing get_base_url would prevent us from using this as expected in templates etc. as the result always would have to be unpacked and passed to a reverse function.

A third option could be to drop get_urls altogether and force the app defining the dashboard.py to register it's url via urls.py. But that way the form helpers would not work that easy and the encapsulating provided by components would be lost. Furthermore the dashboard url prefix would be lost. So in total i'd adivse against ist.

Previously the get_view method and the get_urls acted exactly as you request them now, as get_view was required while get_urls could add an arbitrary number of option urls. It was not required that the view returned by get_view was also used within the get_urls list. Making the abstract class using get_urls[0] to give a abstract get_view implementation was just for convenience. But i think this is a bad way as it introduces new concepts to understand and one would have to remember that for the main view a dispatcher is used with different url args then for the rest of the components urls.

in short. i think it's good to have a simple interface that is in accordance with djangos default. and i think it is pythonic to return a list with a single element and thus won't confuse developers. the get_base_url has a sane default implementation is won't require developers to overwrite it every time.

4. option: it would be possible to allow get_urls also return just a triple of type tuple and check that in the calling code. but i think it's to much of an effort for allowing developers to skip "[ ]"

meinberlin/apps/dashboard2/components.py
```diff
@@ -1,72 +1,135 @@
...
+    """Interface for dashboard components.
+
+    Dashboard components are globally registered by their unique identifier.
+    Every dashboard component is used to configure either a project
+    or a module.
+    They define the urlpatterns and views required for configuration
+    and provide an url which acts as the base entry point.
+    Dashboard components keep track of the projects publish progress.
+    A project may only be published if every component progress is complete.
+    If a project is published, the component has to ensure that it's progress
```
xi: type: "its"

meinberlin/apps/dashboard2/components.py
```diff
@@ -123,45 +197,77 @@ def _get_progress_for_object(object, required_fields):
...
+    a get_progress implementation based on this list and prevents from removing
+    required data if the project is published.
+
+    Properties
+    ----------
+    menu_label : str
+        This label is always returned regardless of module state
+    form_title : str
+        This title is shown on top of the rendered form
+    form_class : ModuleDashboardForm
```
xi: The type is wrong here: `form_class` should not be an instance of `ModuleDashboardForm` but a descendant

PR #707, opened 2017-08-22
Allow to (un)publish projects

kleingeist:
This PR adds a separate view that acts as the target of the form containing the (un) publish button inat the progress indicator.
It ensures each required field of a project and its modules is present. 
*Note: Currently this is done by checking the progress explicitly. But this could be replaces by some kind of signal/callback functionality that would be called prior to publishing. This callback could also check more complicated conditions like: allow publishing only if phases are non overlapping*

The view redirect to the previous page. Or the default project-update page if the previous page can not be detected.


xi:
LGTM.

PR #710, opened 2017-08-22
mark required fields for publish

vellip:


meinberlin/apps/dashboard2/forms.py
```diff
@@ -72,6 +80,9 @@ def __init__(self, *args, **kwargs):
             _make_fields_required(self.fields.items(),
                                   self.get_required_fields())
 
+        _make_fields_required_for_publish(self.fields.items(),
+                                          self.get_required_fields())
```
xi: I am not sure: Maybe we want this only if the project is still in draft.

PR #711, opened 2017-08-22
fix poll management with multi module

xi:
This is a temporary workaround. Longterm we should define the module in the URL directly.


kleingeist:
lgtm. i really look forward to do this in the new dashboard

PR #712, opened 2017-08-22
add unload warning

vellip:
- [x] debug CKEDITOR change event not firing


meinberlin/apps/contrib/static/js/unload_warning.js
```diff
@@ -0,0 +1,24 @@
...
+
+$(function () {
+  var submitted = false
+  var changeHandler = function () {
+    $(window).on('beforeunload', function () {
+      if (!submitted) {
+        return false
+      }
+    })
+  }
```
xi: The spec says that you should set [`event.returnValue`](https://www.w3.org/TR/html5/browsers.html#dom-beforeunloadevent-returnvalue) to a message. [MDN](https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload) says that some browsers ignore the message and always show a default one. I think it does not hurt to add some compatibility code here.

meinberlin/apps/contrib/static/js/unload_warning.js
```diff
@@ -0,0 +1,24 @@
...
+
+  // eslint-disable-next-line no-constant-condition
+  if (!typeof CKEDITOR === 'undefined') {
```
xi: This condition does not work for me. Why not simply `if (CKEDITOR)`?

vellip: It threw an error for me, for some weird reason.

meinberlin/apps/dashboard2/templates/meinberlin_dashboard2/base_dashboard_project.html
```diff
@@ -54,3 +54,8 @@
         </div>
     </div>
 {% endblock %}
+
+
+{% block extra_js %}
+    <script src="{% static 'js/unload_warning.js' %}"></script>
```
kleingeist: maybe we could just include the  script tag in the body/dashboard_content to ensure it is always loaded. currently templates inheriting from base_dashboard_project.html have to remember to call block.super

vellip: I really dislike having JS **somewhere** in the middle of `<body>`. I know there are possibly already occurrences where we do this and I know loading it in `<head>` isn't better, but I'd rather not make it worse.

kleingeist: okay, then i'll merge now :)

PR #713, opened 2017-08-22
Fix missing initiator request emails for the supervisor

kleingeist:
Add SUPERVISOR_EMAIL setting
Supervisors should e.g. receive notifications about initiator requests.

Note that supervisor-berlin@liqd.net is not set yet. Before deployment it should be set up and redirected to the correct supervisor addresses.

Note also, that this is an ugly hack, as we should define the role supervisor on the User model if we'd really need them and then adapt get_receivers to return the users with the supervisor role.


xi:
I think this is fine for now. We do not need to over-engineer our role model.

PR #714, opened 2017-08-22
Duplicate module settings if present

kleingeist:
Fixes the issue with error if a project with a module with settings was duplicated


xi:
LGTM

PR #715, opened 2017-08-22
fix project create view

xi:
it probably broke in #690


kleingeist:
lgtm but i don't know if the classes are correct
the blueprint list is using 'menu-layout__title' for the h1 
the project list 'lr-bar__left' 
and create project the enclosing divs with 'menu_layout' and 'menu-layout__content menu-layout__content--aside'

xi:
`menu-layout__title` is deprecated in removed in #709

PR #720, opened 2017-08-23
refactor category form

xi:
this propose several changes to the category form. The goals are

-   better usability (e.g. for deleting)
-   simpler code (e.g. no popover)
-   make the form look ok on its own (as it will be on its own page in the new dashboard)


vellip:
works good for me

PR #721, opened 2017-08-23
create project in modal

vellip:


meinberlin/apps/contrib/static/js/ajax_modal.js
```diff
@@ -0,0 +1,34 @@
+/* globals $ */
+$(function () {
+  var $ajaxModals = $('[data-toggle="ajax-modal"]')
+  var modalHTML = (
+    '<div class="modal">' +
+      '<div class="modal-dialog modal-lg" role="document">' +
+        '<div class="modal-content">' +
+          '<div class="modal-header"><h2 class="modal-title u-first-heading"></h2><button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button></div>' +
```
xi: The "Close" label should be translated.

meinberlin/apps/contrib/static/js/ajax_modal.js
```diff
@@ -0,0 +1,34 @@
...
+          '<div class="modal-body"></div>' +
+      '  </div>' +
+      '</div>' +
+    '</div>'
+  )
+  var $backdrop = $('<div class="modal-backdrop" />')
+
+  $ajaxModals.on('click', function (e) {
```
xi: I would prefer to use event delegation here `$(document).on('click', '[data-toggle="ajax-modal"]', function() {})`

vellip:
I noticed the thing with the validation errors too. But I think that's something we can live with. I'd not want to code JS Form validation.  That's out of scope.

kleingeist:
i also think we should redirect to the blueprint list on cancel instead of the project list.

PR #722, opened 2017-08-23
Sort components by weight

kleingeist:
If the weight is the same, the identifier is used to get a constant order.

If this will be required some day it would be possible to use a weight @property which determines the weight from a setting.

### DIFF
--- a/meinberlin/apps/dashboard2/contents.py
+++ b/meinberlin/apps/dashboard2/contents.py
@@ -17,14 +17,6 @@ class DashboardContents:
                              .format(component.identifier))
         self._registry[section][component.identifier] = component
 
-    def unregister_project(self, component):
-        identifier = getattr(component, 'identifier', component)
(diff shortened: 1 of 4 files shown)


meinberlin/apps/dashboard2/contents.py
```diff
@@ -1,25 +1,37 @@
...
+                             .format(component.identifier))
+        self._registry[section][component.identifier] = component
+
+    def unregister_project(self, component):
+        identifier = getattr(component, 'identifier', component)
+        self._registry['projects'].pop(identifier, None)
+
+    def unregister_module(self, component):
+        identifier = getattr(component, 'identifier', component)
+        self._registry['modules'].pop(identifier, None)
```
xi: Are we sure that unregistering actually works?

kleingeist: i removed them until we need them. reverting a single commit will be fairly easy

PR #724, opened 2017-08-23
Temporary set single module name and description

2e2a:
Set valid module name and description from project.
Will be replaced by module type and user input in the new dashboard.


kleingeist:
lgtm

PR #728, opened 2017-08-24
style select item

vellip:
screen shot 2017-08-24 at 13 54 41

depends on https://github.com/liqd/adhocracy4/pull/164 but merging won't break anything.


2e2a:
lgtm

PR #729, opened 2017-08-25
add datepicker to dashboard2

vellip:


kleingeist:
lgtm

vellip:
That's a good question. I guess the best would be to let each team have their own version of the datepicker.

PR #734, opened 2017-08-28
Dashboard2 fixes

kleingeist:
- [x] Fix project creation without module settings
- [x] Set project tab to active on projects list and create
- [x] menu order and required fields def


2e2a:
lgtm

PR #738, opened 2017-08-28
Dashboard2.0 for external and bplan projects

kleingeist:
This PR has become bigger then expected.
For one thing it splits the components to different packages and renames the registry from contents to components
Furthermore it adds a configurable ProjectDashboardAbstraction which provides information about the dashboard "state" of a certain project.

And for another thing it adds (based on the former) support for external and bplan projects.


meinberlin/apps/dashboard2/utils.py
```diff
@@ -0,0 +1,134 @@
+from functools import lru_cache
+
+from django.conf import settings
+from django.utils.module_loading import import_string
+
+from . import components
+
+
+def get_project_dashboard(project):
```
2e2a: Not sure that utils is a good place for this dashboard interface. Maybe just move them to `__init__.py`?

2e2a:
lgtm

PR #739, opened 2017-08-28
add nav to project list

vellip:
This merely adds a nav to the project list in dashboard. We could still do the same for blueprint list and highlight the currently active one (which doesn't work for blueprint list because it's a button now).


2e2a:
lgtm

PR #740, opened 2017-08-28
 Add Area Settings component 

kleingeist:
As FormSets won't work with OneToOne relations AreaSettings now have a custom form which is replacing the forms instance with the modules settings_instance. 

Note: The settings concept is still broken as OneToOne makes no sense at all. But this brings back the current functionality. Settings should be discussed at SoS


2e2a:
lgtm

PR #743, opened 2017-08-29
Export module component 

kleingeist:
Currently the component always returns the first export found. In the
future different exports for a single module may be possible, but this
would require a intermediate page to select the desired export.

The  export component is only shown for published projects


2e2a:
lgtm

PR #744, opened 2017-08-29
Document dashboard component

2e2a:
- [x] reload dashboard view on document save


kleingeist:
i did add the functionality to reload the page on success but i am not convinced it is a good option anmyore:

It helps to get rid of the the red exclamation marks, but it  is
really stupid.
Alternatives would be to
- reload the entire page with javascript and replace the contents like
it is done in the embed
- remove the icon explicitely on success instead of reloading

@vellip could you please take a look at the commit: https://github.com/liqd/a4-meinberlin/pull/744/commits/97035853cfacb38723e14fe4dd856368c5a45696  and  give some feedback or even find a less weird solution? that'd be great :sparkling_heart:

vellip:
https://github.com/liqd/a4-meinberlin/pull/744/commits/980eace1cf4bf26293f1587990a972ef0ed6b155 should theoretically work, I wasn't able to test since the module's menu items have no clue which one is active yet.
We just have to call the particular function in `DocumentManagement` when there is at least one chapter or there is none.

kleingeist:
thank you, that looks better. but i just realized we are also missing the progress circle.
it seems like a really bad idea to have state in different places :( (django/html and react)

vellip:
That should work now! :+1:

kleingeist:
yipie. thank you @vellip  :)

PR #747, opened 2017-08-30
style focus for image upload input

vellip:


vellip:
ugh nevermind, doesn't work if there's an image uploaded already because then it'd highlight the delete button

PR #748, opened 2017-08-30
Allow for optional datetimepicker fields

kleingeist:
This allows for optional datetimepicker fields. 
Furthermore it sets the default time if only a date is submitted.


vellip:
Cool bananas! Thanks!

PR #749, opened 2017-08-30
Switch topicprio to dashboard2

2e2a:


meinberlin/apps/topicprio/views.py
```diff
@@ -96,80 +94,73 @@ class Meta:
...
+class TopicListDashboardView(mixins.DashboardComponentMixin,
+                             mixins.DashboardBaseMixin,
+                             mixins.DashboardContextMixin,
+                             filter_views.FilteredListView):
     model = models.Topic
-    template_name = 'meinberlin_topicprio/topic_mgmt_list.html'
+    template_name = 'meinberlin_topicprio/topic_dashboard_list.html'
     filter_set = TopicCreateFilterSet
     permission_required = 'a4projects.add_project'
     project_url_kwarg = 'slug'
```
kleingeist: there is no longer a valid project_url_kwarg
the list view is routed with the module_slug as named parameter as defined in .dashboard.py

meinberlin/apps/topicprio/templates/meinberlin_topicprio/includes/topic_dashboard_list_item.html
```diff
@@ -4,13 +4,13 @@
     <div class="lr-bar">
```
kleingeist: lit-item--squashed should be removed

kleingeist:
some classes have to be adapted for the list view. buti can't comment on them. i will add some commits

PR #755, opened 2017-08-30
Dashboard2 duplicate

### LINKED ISSUE #733: Number appended to module name on project duplication

2e2a:
Copied publicate button and mixin to dashboard2.
Also fixes #733 

CSS of dropdown menu is broken.


vellip:
@2e2a should work now

kleingeist:
it's weird, that the list items are so big and only the buttons are interactive. i'd expect that the image links somewhere. 
but this is another issue. the dropdown looks good to me.
i don't like the final double bottom border but that a general problem

PR #757, opened 2017-08-31
Add explicit is_effective check to components

kleingeist:
Everything on a component (get_menu_label and get_progress ...) may only be called if is_effective is True.
This allows effectively to replace get_menu_label(project_or_module) with a fixed label property.
Furthermore it prevents to have to double check on module types at many places


2e2a:
Nice.

PR #759, opened 2017-08-31
Add dashboard signals

kleingeist:
Note that the project_pre_publish signal may be used to prevent from publishing if
any connected handler returns an error message.
The create signals are derived directly from the db signals. Unfortunately this is not easily possible for the other cases as they require a before/after knowledge.

### DIFF
--- a/meinberlin/apps/dashboard2/views.py
+++ b/meinberlin/apps/dashboard2/views.py
@@ -84,10 +84,6 @@ class ProjectCreateView(mixins.DashboardBaseMixin,
 
     def form_valid(self, form):
         response = super().form_valid(form)
-
-        signals.project_created.send(sender=None, project=self.object)
-
-        # FIXME: maybe replace by dashboard signals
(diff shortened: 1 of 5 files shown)


meinberlin/apps/dashboard2/views.py
```diff
@@ -173,8 +173,18 @@ def publish_project(self):
                              'Required fields are missing.'))
             return
 
+        responses = signals.project_pre_publish.send(sender=None,
+                                                     project=self.project)
```
2e2a: According to django documentation `sender` must be set:

>  You must provide the sender argument (which is a class most of the time) ...

kleingeist: in the Signal source it states a class or None, so we are fine.

2e2a:
`sender=NONE` seems to be allowed by the code documentation and we dont select by sender anyway.

lgtm

PR #762, opened 2017-08-31
Remove and replace the old dashboard

kleingeist:
To split the dashboard into a meinBerlin and a Core component, a new feature was required, a blueprint registry.


vellip:
As discussed, we are not 100% certain about the blueprint registry. The PR looks good to me otherwise, so feel free to merge when the blueprint topic is resolved.

PR #763, opened 2017-09-04
translate leaflet draw

vellip:


meinberlin/assets/js/i18n-leaflet-draw.js
```diff
@@ -0,0 +1,106 @@
+window.L.drawLocal = {
+  draw: {
+    toolbar: {
+      actions: {
+        title: 'Bereich entfernen',
+        text: 'abbrechen'
```
2e2a: Some texts are lowercase, most uppercase. I would suggest to have everything uppercase.

2e2a:
lgtm

xi:
Why not use this with our regular translation system?

PR #766, opened 2017-09-05
announce file type to user

vellip:
required for accessibility: http://testen.bitv-test.de/index.php?a=di&iid=94


xi:
I am not sure whether "excel" is an accurate description of the file format. But I guess it's fine.

PR #771, opened 2017-09-07
add text for create project ux

### LINKED ISSUE #769: Dashboard: Info text on next steps
The new dashboard look vastly improved. One thing I noticed is, that when you create a new project, it would be good to let users now, that they can edit and save the project later. Currently, users who do not know how creating a project works could think that it would be published right away. I would propose to add a little note underneath the send button: „Im nächsten Schritt können Sie weitere Einstellungen vornehmen, bevor sie das Projekt als Entwurf speichern oder veröffentlichen“.

Also, the send button label „Senden“ needs to be changed to something more clear like „Fortfahren“ oder „Nächster Schritt“.

bildschirmfoto 2017-09-06 um 15 11 26

vellip:
My proposal for #769

closes #769


xi:
Nice!

PR #773, opened 2017-09-07
text review issues

### LINKED ISSUE #587: Small textreview usability and UI issues
Encountered some small usability issues:
- [ ] Chapter in the table of content are enumerated, but not in the editor. This should be unified.
- [X] When selecting a chapter for editing, jumping to the chapter part of the editor would be nice. As with many chapters, the initiator has to scroll a lot
- [ ] Table of contents design does not look that nice with actual content (e.g. https://mein.berlin.de/projects/arbeit-40-made-in-berlin/)

vellip:
fixes #587


xi:
much better already!

PR #779, opened 2017-09-11
fix action time format

### LINKED ISSUE #648: Action time is off by 12h
I just created a comment in my local dev environment. The related action was 12 hours in the past.

xi:
Also fixes #648. Note that the bug only occured after 12:00


2e2a:
lgtm

PR #781, opened 2017-09-12
only show feedback button on large screens

### LINKED ISSUE #518: feedback button overlaps content on zoom
Part of the BITV-Test: "1.4.4a Schriftgröße variabel" and "1.4.4b Bei Zoom auf 200% benutzbar"
----
It's not really a big thing, but the test fails because of it.

xi:
fixes #518

We already did hide the feedback button on small screens. But there was a rather large area in which we could not be reasonably sure that nothing is covered. This is fixed now.

Note that the feedback form is already linked in the footer, so the information is still available.


2e2a:
lgtm

PR #787, opened 2017-09-12
clearify connection between "notification" and "follow"

xi:
we got the feedback that the connection between "email", "notification", and "follow" is not completely clear.

@janaba what do you think of this change?


2e2a:
lgtm

PR #791, opened 2017-09-13
Dashboard project filter

2e2a:


xi:
LGTM. I have the feeling that sorting would be nice. But sort by what!?

PR #800, opened 2017-09-14
Maptopic permission tests

2e2a:


2e2a:
Merging as followup to #770

PR #801, opened 2017-09-18
Remove potential circular import

kleingeist:
`get_blueprints()` dashboard2/blueprints.py is importing the module from the settings (in our case dashboard/blueprints.py) which is importing `ProjectBlueprint` from dashboard2/blueprints.py

```
dashboard2/blueprints.py 
[ blueprints = _get_blueprints()  ->  importlib.import_module ]
dashboard/blueprints.py 
[ from meinberlin.apps.dashboard2.blueprints import ProjectBlueprint ]
dasboard2/blueprints.py
```

I am not sure why this has not been a problem yet. Maybe because only some metehods/classes are imported?  @2e2a  @xi  do you have any idea how python resolves this?


2e2a:
nice

PR #806, opened 2017-09-19
add address search to map_choose_point

xi:
Still missing:

- [x] Points should be filtered by the polygon that is provided with the map. This can happen in the frontend or backend.
- [x] Connect to the zbp geocoding API (see https://github.com/liqd/django_zbp/pull/54)
- [X] Update marker when value changes (see https://github.com/liqd/adhocracy4/pull/183)
- [x] Give user feedback after receiving the points: Either let them choose among different points, show a warning that no matches were found or signal success.
- [ ] Initially hide the address search and add a toggle somewhere


meinberlin/assets/js/map-address.js
```diff
@@ -0,0 +1,159 @@
...
+      var points = []
+      cb(points)
+    }
+  })
+}
+
+var renderPoints = function (points) {
+  if (points.length === 0) {
+    return $('<span class="complete__warning">')
+      .text(django.gettext('No matches found'))
```
kleingeist: maybe we should output "No matches found within the defined area" or something.

PR #807, opened 2017-09-19
use second breakpoint

xi:
We have had a second breakpoint for a while now, but only used it in few cases. This uses this second breakpoint to do some optimisations for medium sized screen (and potentially embeds).


kleingeist:
lgtm.
i think it's okay the image is not scaling steadily

PR #808, opened 2017-09-19
Accessibility fixes

kleingeist:
1. wraps all dates in time elements. provides the (python) iso8601 formatted  date to the datetime attribute
2. Make item title the first element in item listings

@xi i'm still not sure if we shouldn't better use the lr-bar also for the listings as lr-bar is used for similiar listings in the dashboard. we could either keep it's behaviour on small screens or replace lr-bar with flexboxes, too

Note the difference to the current master on small screens:

Flex (pr):
list-item-flex

Float (master):
list-item-float

### DIFF
--- a/locale/de/LC_MESSAGES/django.po
+++ b/locale/de/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: a4-meinberlin\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-09-18 16:33+0200\n"
+"POT-Creation-Date: 2017-09-19 15:28+0200\n"
 "PO-Revision-Date: 2017-09-18 15:04+0200\n"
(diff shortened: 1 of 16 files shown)


locale/de/LC_MESSAGES/django.po
```diff
@@ -1903,21 +1902,23 @@ msgstr "Projekt auf meinBerlin"
...
 msgid ""
 "\n"
-"                    Participation possible until %(date)s\n"
+"                    Participation possible until <time datetime=\"%(isodate)s"
+"\">%(date)s</time>\n"
 "                "
 msgstr ""
 "\n"
-"Beteiligung möglich bis zum %(date)s"
+"Beteiligung möglich bis zum <time datetime=\"%(isodate)s\">%(date)s</time>"
```
xi: Would it be possible to leave the string as-is and just insert the complete HTML?

kleingeist:
closed in favor of #811

PR #809, opened 2017-09-19
Revert to npm install

### LINKED ISSUE #799: Fix npm install/update
In ee490509e I changed `npm install` to `npm update --dev` in out Makefile. The reason was that `npm install` seemed not to update packages.

The issue with `npm update --dev` is that it scrambles the package.json file. It also seems to not install all required packages. Further issues may exist with linked packages in development, e.g. adhocracy4.

npm has had several new versions since that which, among other things, introduced the package-lock.json file.

We should check the above assumptions and re-evaluate our npm workflow. For example, it may be a good idea to go back to `npm install` and commit package-lock.json to the repo.

xi:
Fixes #799

I am not relatlivly sure that `npm install` is what we want. I am not
sure whether it is bug-free, but that is another question.

package-lock.json should not do any harm, but I do not see any benefit
in committing it to the repo. We did not have trouble with different
versions on different systems so far and it might cause issues when
using `npm link`. So I added it to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 test_db.sqlite3
 db.sqlite3*
 node_modules/
+package-lock.json
 meinberlin/config/settings/local.py
 .env/
 .idea/
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ help:
 	@echo
 
 install:
-	npm update --dev
+	npm install
 	if [ ! -f $(VIRTUAL_ENV)/bin/python3 ]; then python3 -m venv $(VIRTUAL_ENV); fi
 	$(VIRTUAL_ENV)/bin/python3 -m pip install -r requirements/dev.txt
 	$(VIRTUAL_ENV)/bin/python3 manage.py migrate
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ help:
 	@echo
 
 install:
-	npm install
+	npm install --no-save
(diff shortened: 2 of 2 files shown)


kleingeist:
If i understand it correctly we should use `npm install` to install packages as was reintroduces in this PR.
But we should also call `npm update --dev --save` and commit the new versions in package.json to our repo manually.
we could also use https://greenkeeper.io/ which seems like the pyup equivalent for npm to me

kleingeist:
also i wanted to note, that travis is currently setup to use nodejs 6 (stable) which includes npm 3.10.10
npm 3.10 does not use the package-lock.json
we should try to keep our local machines, travis and the buildserver in sync. 

As the buildserver is using npm 3.10.9 i think we should try to keep the nodejs 6 stable branch on our dev machines and travis, too.

For npm 3.10.9  `npm install` works fine and updates the packages when called.
Unlike npm 5.3 it does not touch the package.json. But it doesn't care about `--no-save` 

In total i'd suggest to use
* `npm install --no-save` in the `Makefile`
* keep our dev machines on npm 3.10 to be in sync with the buildserver (and travis)
* use greenkeeper or manually update the packages with `npm update --dev --save` from time to time

xi:
> keep our dev machines on npm 3.10 to be in sync with the buildserver (and travis)

I am not sure how much I agree with that. Yes, this should be the default. But I do think it makes sense to try to be compatible with newer versions as well.

PR #813, opened 2017-09-20
Update travis config

kleingeist:
* Change travis python version to 3.4 
  On our servers python 3.4 is running so we should use that version on travis, too
* Make travis fail if there are missing migrations


xi:
LGTM

PR #815, opened 2017-09-20
load new scripts in ajax modal

### LINKED ISSUE #795: Selected image not shown in the create project dialog

xi:
Fixes #795 

The fix is duplicated from embed.js: The content of the embed is loaded via ajax. Apart from the content, we also need to extract the script elements from the downloaded page and add them to the current one.


kleingeist:
lgtm

PR #819, opened 2017-09-21
Use datetimefield from core

2e2a:
Use https://github.com/liqd/adhocracy4/pull/177


meinberlin/apps/offlineevents/forms.py
```diff
@@ -1,7 +1,7 @@
 from django import forms
 from django.utils.translation import ugettext_lazy as _
 
-from meinberlin.apps.datetimefield.widgets import DateTimeInput
+from adhocracy4.forms.widgets import DateTimeInput
```
kleingeist: why are we using the widget here instead of the field? accident or decision?

2e2a: I was not sure and kept it as it was.

PR #820, opened 2017-09-21
Remove project type

2e2a:
The dashboard list items might need a better styling now, as they look emtpy without the project type.

Reflects: https://github.com/liqd/adhocracy4/pull/186


kleingeist:
lgtm

PR #821, opened 2017-09-25
Container projects basic backend

kleingeist:
This adds a new sub-project-type "container" which may include a number of projects via an m2m relation

missing (not sure if it should be done in this pr)
- [ ] dashboard component for inclusion of projects to containers (no concept yet, django admin for now)
- [ ] concept on how to show containers (should they show an ends in X ...)
- [ ] extended project detail view for containers (no concept yet)

Furthermore the project list is now filtered with the same rules as the view_project permission. Thus the `filter_has_perm` check in 'project_list.html' is redundant and *could* be removed. But i am not sure if this is reasonable as the view_project perm could possibly be extended with some logic not able to represent in a query.


kleingeist:
as many open questions regarding the US are remaining this PR is set to incomplete and should probably be closed.

Questions
* search/filter should maybe filter on the contents of the contained projects
* how to sort
* list contained projects individually or not
* what to show on the detail view
* may containers contain all projects or only from the same org (as the container)
* wagtail project block is different, as the project is not casted there (probably not a problem if the tile layout changes altogether)
* howto redirect to the container detail view from the project detail view
* Downloadmanager
* Howto link from projects back to their container

xi:
replaced by #878

PR #822, opened 2017-09-25
add language attribute to embed.html

### LINKED ISSUE #818: No comments in embedded document
The Chapter view in the embed does not include comments:
https://tempelhofer-feld.berlin.de/projekte/geschaftsordnung-der-feldkoordination/

xi:
fixes #818

the same was added to base.html in 6ed1f4c87


kleingeist:
i can reproduce the problem and the solution, but i don't know why-

xi:
Comment.jsx from adhocracy4 core contains the following code:

```
….toLocaleString(document.documentElement.lang)
```

This raises an exception of the `lang` attribute is not a valid BCP47 tag.

PR #833, opened 2017-09-25
work around layout bug in IE

### LINKED ISSUE #790: IE11: broken hero layout
The text in the hero unit on the front page should be at the bottom. In IE11 it is at the top.

xi:
fixes #790

see https://connect.microsoft.com/IE/feedback/details/802625/min-height-and-flexbox-flex-direction-column-dont-work-together-in-ie-10-11-preview


kleingeist:
lgtm but ie makes me sad :(

PR #835, opened 2017-09-25
Focus on input elements in react

### LINKED ISSUE #797: Set focus to new question/choice/paragraph in react forms
When adding a new question/choice/paragraph, the focus should directly jump to the new form elements. Stepping backwards through form fields feels awkward.

kleingeist:
Fixes #797


xi:
Seems to work fine for the most part. But I found one case where focus is not set properly: In polls, add an answer to a question that is not the last one.

xi:
Not necessarily nice, but as good as it will get. From a usability perspective, this is a huge improvement IMHO.

PR #836, opened 2017-09-26
Use react stateless components for poll forms

kleingeist:
depends on #835


kleingeist:
Fixed with commit d2f5a0a

PR #837, opened 2017-09-26
add asterisk to required fields

### LINKED ISSUE #406: Label required/optional form fields
I would suggest to add am indicator for optional / required fields. We could use an asterisk within the labels if a field is required.

xi:
fixes #406

There was the feeling that not all forms should have the asterisks (e.g. login). It was not clear though how to differeenciate the two cases. The decision was to just add it everywhere for now and add exceptions later (if required).


2e2a:
Looks good.

PR #841, opened 2017-09-26
translate follow button

### LINKED ISSUE #832: Following button not translated on productive
The text in the dropdown of the following button is still in English while the rest of the site is in German. This is on productive and can be seen here: https://mein.berlin.de/projects/vorbereitende-untersuchungen-ehemaliger-guterbahnh/?mode=list

It was tested on Safari 10.1 in German.

bildschirmfoto 2017-09-25 um 16 45 29

xi:
fixes #832
depends on https://github.com/liqd/adhocracy4/pull/189

### DIFF
--- a/locale/de/LC_MESSAGES/djangojs.po
+++ b/locale/de/LC_MESSAGES/djangojs.po
@@ -126,7 +126,23 @@ msgstr[1] "%s Antworten sehen"
 msgid "Answer"
 msgstr "Antworten"
 
-#: adhocracy4/follows/static/follows/FollowButton.jsx:55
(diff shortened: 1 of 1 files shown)


kleingeist:
i would prefer to try to find a general solution. There are additional places in the .jsx files where django.translate is used and i don't think we should workaround all of them :/

xi:
I would prefer a proper fix, too (potentially upstream). However, I don't know of any other issues. This is just a pragmatic workaround. We can merge this and still do a proper fix later.

kleingeist:
i have merged the core PR, so this is good to merge, too

PR #842, opened 2017-09-26
Enable scrolling in long modals

### LINKED ISSUE #804: overlay not scrollable
If I want to create a new project and my screen is too small, I cannot scroll down and thus I cannot submit the new project.

kleingeist:
The core problem was a missing `overflow-y: auto;` on the modal.

It was missing because the body was missing the `modal-open` class which
is set via the Bootstrap javascript when a modal is opened. As we did
not initialize the modal via javascript but rather tried to simply use
the classes directly on the html elements the scrolling has not been
enabled.

Using the javascript initialization also fixes the need for the custom
creation of a backdrop, as this will created and removed automatically
by bootstrap.

Furthermore the modal was not added to the body but rather somewhere in
the dom tree which is advised against by the bootstrap docs:

> Modals use position: fixed, which can sometimes be a bit particular
> about its rendering. Whenever possible, place your modal HTML in a
> top-level position to avoid potential interference from other
> elements. You’ll likely run into issues when nesting a .modal within
> another fixed element.

This PR fixes #804 but introduces the problem of two scrollbars, as we set the `overflow-y: auto;` on the `html` element instead of the `body` element. Bootstrap proposes to use scrolling on the body element

### DIFF
--- a/meinberlin/apps/contrib/static/js/ajax_modal.js
+++ b/meinberlin/apps/contrib/static/js/ajax_modal.js
@@ -12,7 +12,6 @@ $(function () {
       '</div>' +
     '</div>'
   )
-  var $backdrop = $('<div class="modal-backdrop" />')
 
   var extractScripts = function ($root, selector, attr) {
     var $existingValues = $('head').find(selector).map(function (i, e) {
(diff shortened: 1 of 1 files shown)


xi:
Nice!

PR #849, opened 2017-09-27
Remove alerts on CKEDITOR fields change

### LINKED ISSUE #754: React success message disappears only after reload
When a document or poll is saved the shown success message does not disappear on further react actions (e.g. adding a paragraph). It stays until a page reload.

This might be confusing for the user.

kleingeist:
Fixes #754

CKEDITOR does not fire change events in the DOM when the field is
modified. This is the simplest but kinda hacky solution to fix #754.
It requires knowledge about the child components inner structure.

Fixing the problem at its core would be possible if we'd handle change
events on the CKEDITOR editor instance, adapt the textarea accordingly
and fire change events manually. For example like:
```
window.CKEDITOR.on('instanceReady', (e) => {
	e.editor.on('change', (e2) =>  {
		$textarea.value = e2.value
		$textarea.change()
	})
})
```
But this is out of the scope of this issue

### DIFF
--- a/meinberlin/apps/documents/assets/DocumentManagement.jsx
+++ b/meinberlin/apps/documents/assets/DocumentManagement.jsx
@@ -210,7 +210,9 @@ class DocumentManagement extends React.Component {
       }
     }
     this.setState({
-      chapters: update(this.state.chapters, diff)
+      chapters: update(this.state.chapters, diff),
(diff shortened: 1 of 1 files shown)


xi:
For me the alert is removed by just focusing the ckeditor every other time. But I still think this is an improvement.

PR #850, opened 2017-09-27
Hide dropdown menu and disable button for external projects

### LINKED ISSUE #817: Fix or prevent project duplication for external and bplan projects
Currently duplicating external and bplan projects is broken as only the base project is duplicated but not the  additional one-to-one related external/bplan object.
We could either prevent from duplicating non standard projects (my favorite) 
or adapt the duplication logic to duplicate every special project type (bplan, external, possibly containers)

kleingeist:
Fixes #817

As the duplication code will be part of the a4 dashboard it would be
very complex to make the duplication code extensible and allow for
correct duplication of external and bplan projects.

Duplication of external and bplan projects is still possible but the
duplicated project looses its "type".

This PR simply hides the duplication from the user.


xi:
LGTM

PR #856, opened 2017-09-28
Sort modules without start dates to the future

kleingeist:
Modules without a start date (phases dates are not set) will be sorted
after every other module or offlineevent.


meinberlin/apps/offlineevents/templatetags/offlineevent_tags.py
```diff
@@ -12,10 +14,19 @@ def offlineevents_and_modules_sorted(project):
...
     res = modules + events
-    res_sorted = sorted(
-        res, key=lambda x: x.first_phase_start_date if
-        isinstance(x, Module) else x.date)
-    return res_sorted
+    return sorted(res, key=cmp_to_key(_cmp))
+
+
+def _cmp(x, y):
+    x = x.first_phase_start_date if isinstance(x, Module) else x.date
```
xi: redefining a variable is bad style IMHO

meinberlin/apps/offlineevents/templatetags/offlineevent_tags.py
```diff
@@ -12,10 +14,19 @@ def offlineevents_and_modules_sorted(project):
...
+    if x is None:
+        return 1
+
+    y = y.first_phase_start_date if isinstance(y, Module) else y.date
+    if y is None:
+        return -1
+
+    return (x > y) - (y < x)
```
xi: This looks wrong. I think the first one must be `<`.

But actually, I think this line is too dense. I would prefer a more verbose version:

```
if x > y:
  return 1
elif x == y:
  return 0
else:
  return -1
```

kleingeist: while investigating the python3 way of sorting items via the `cmp_to_key` function i stumbled upon the this snippet. i think even in a pep or something on the python domain (but i can't find it anymore)
but i have to agree, that it took some time for me to understand it again after one week of absence, so yes i think too it would be better to refactor :D

PR #859, opened 2017-10-04
Bplan archiving

2e2a:
Show information text when the bplan project is over

### DIFF
--- a/meinberlin/apps/bplan/views.py
+++ b/meinberlin/apps/bplan/views.py
@@ -1,4 +1,6 @@
 from django.core.urlresolvers import reverse_lazy
+from django.http.response import HttpResponseRedirect
+from django.urls import reverse
 from django.views import generic
 from django.views.generic import TemplateView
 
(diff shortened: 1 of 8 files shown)


meinberlin/apps/bplan/views.py
```diff
@@ -21,6 +23,12 @@ class BplanStatementFormView(ProjectContextDispatcher,
     template_name = 'meinberlin_bplan/statement_create_form.html'
     success_url = reverse_lazy('meinberlin_bplan:statement-sent')
 
+    def dispatch(self, request, *args, **kwargs):
+        project = self.get_project(*args, **kwargs)
+        if project.has_finished:
+            return HttpResponseRedirect(reverse('meinberlin_bplan:finished'))
```
xi: Why a redirect? I prefer showing the information text on the same URL.

meinberlin/apps/projects/views.py
```diff
@@ -127,6 +127,9 @@ def get_queryset(self):
                 Q(participants__pk=self.request.user.pk) |
                 Q(organisation__initiators__pk=self.request.user.pk) |
                 Q(moderators__pk=self.request.user.pk)
+            ) & (
+                Q(is_archived=False) |
+                Q(externalproject__bplan=None)
```
xi: This code deserves a comment.

meinberlin/apps/bplan/views.py
2e2a: The `finished` view has different permissions from the statement submission view. With the redirect I was able to use the PermissionRequired mixin.

meinberlin/apps/projects/views.py
kleingeist: i just looked over this again after my holidays and am not sure if it does really work is_archived is also set to True via a filter.
i would like to look into it again. so please don't merge yet

kleingeist: okay the expression is alright. all hail to the glory of databases :D

kleingeist:
lgtm. 
it's weird that the default project detail page is shown prior to the "statement" phases start and that it is opened on the information tab while there is no information, but that has been the same as before. maybe we should address this in another PR.

PR #869, opened 2017-10-09
Ignore missing docstrings

kleingeist:
Somehow the linting failed on travis with error codes D106 and D107

(see http://www.pydocstyle.org/en/latest/error_codes.html for a complete list of error codes)


kleingeist:
closed as the fix is already included in #855

PR #874, opened 2017-10-09
container dashboard

xi:
This add a container list and a container create for to the dashboard. The actual funcitonality is mostly missing (e.g. the container list currently simply lists all projects).


kleingeist:
just an early note: i think we should move containers from the dashboard2 app to a separate projectcontainers app or the existing meinberlin/apps/projects app.

xi:
replaced by #878

PR #881, opened 2017-10-10
lighter placeholders

xi:
placeholders were often confused for actual input


kleingeist:
looks good in chrome. i hope there are not that many users on old browsers.

PR #883, opened 2017-10-11
Show a simple notice on bplan embeds if participation has not started yet

kleingeist:
The wording is the same as in the timeline, but as bplan projects don't have an info etc. the default project detail (which is shown prior to phases) looked broken.

### DIFF
--- a/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
+++ b/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
@@ -7,6 +7,8 @@
     {% if project|is_container %}
         {% include 'meinberlin_projectcontainers/includes/container_detail.html' with container=project.projectcontainer %}
     {% elif project|project_type == 'bplan' %}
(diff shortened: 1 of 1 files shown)


meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -1,11 +1,21 @@
...
+    {% elif project|project_type == 'bplan' %}
+        <div class="l-wrapper">
+            <div class="l-center-6">
+                <h1 class="u-first-heading">{% trans 'Participation is not possible at the moment.' %}</h1>
+                <p>
+                    {% html_date project.modules.first.future_phases.first.start_date 'DATE_FORMAT' as start_date %}
+                    {% blocktrans with date=start_date %}It starts on {{ date }}.{% endblocktrans%}
+                </p>
+            </div>
+        </div>
```
xi: Could you please add a comment why this is here and how it fixes the issue?

xi:
LGTM. @2e2a any comments?

2e2a:
lgtm

PR #885, opened 2017-10-11
refine container tile

xi:


kleingeist:
the folder icon looks very uncentered on chrome. is that intended? 
icon-not-centered

kleingeist:
lgtm :)

PR #887, opened 2017-10-11
add logo to container detail page

xi:
the logos should be in a square box. In most cases we crop images, but
here we add a white background to make it fit.


kleingeist:
the logo is very small, but i think too that fitting is better for logos with graphical elements.
i'll merge but maybe we have to refine the logo size in the next iteration

PR #889, opened 2017-10-11
put container short description into a box

xi:
we keep piling more and more styling code onto project-header. That will som refactoring some day. But today is not that day.


kleingeist:
the header looks very dense now. especially the container title gets lost between the two pressing boxes

container-dense-header

xi:
replaced by #901

PR #890, opened 2017-10-11
overwrite form wording for containers

xi:
I am not totally sure if the ContainerInformationComponent works correctly like this.

### DIFF
--- a/meinberlin/apps/projectcontainers/dashboard.py
+++ b/meinberlin/apps/projectcontainers/dashboard.py
@@ -4,7 +4,6 @@ from django.utils.translation import ugettext_lazy as _
 from meinberlin.apps.dashboard import get_project_type
 from meinberlin.apps.dashboard2 import ProjectFormComponent
 from meinberlin.apps.dashboard2 import components
-from meinberlin.apps.dashboard2.dashboard import ProjectInformationComponent
 
 from . import forms
 from . import views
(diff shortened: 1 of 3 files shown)


meinberlin/apps/projectcontainers/dashboard.py
```diff
@@ -41,6 +42,15 @@ def get_urls(self):
         )]
 
 
+class ContainerInformationComponent(ProjectInformationComponent):
```
kleingeist: it would be better to duplicate the ProjectInformationComponent

there are just 4 additional properties to set, which would make the ContainerInformationComponent self-contained (weight, label, form_title and form_template). 

currently the page title and the h1 are still saying "edit project information".

meinberlin/apps/projectcontainers/forms.py
```diff
@@ -6,27 +6,35 @@
 
 from . import models
 
+LABELS = {
+    'name': _('Title of your container'),
+    'description': _('Short description of your container'),
+    'information': _('Description of your container'),
```
kleingeist: the information label should be set in a separate dictionary directly on the information form.

meinberlin/apps/projectcontainers/forms.py
```diff
@@ -36,6 +44,18 @@ class Meta:
...
+
+
+class ContainerInformationForm(ProjectDashboardForm):
+
+    class Meta:
+        model = models.ProjectContainer
+        fields = ['information']
+        required_for_project_publish = ['information']
+        labels = LABELS
+        help_texts = HELP_TEXTS
```
kleingeist: help text for "information" is missing

meinberlin/apps/projectcontainers/dashboard.py
xi: That was intentional

meinberlin/apps/projectcontainers/forms.py
xi: That was intentional

kleingeist:
lgtm :)

PR #891, opened 2017-10-11
Fix geo import from berlin fisbroker (fixes #863)

### LINKED ISSUE #863: Import of Bezirksregionen stopped working
`$ manage.py import_geodata --gdal-legacy`

Leads to a `KeyError`, probably the data format has changed.

kleingeist:
There were two problems fixed

1. The berlin WFS service changed its default interface version >2.0 and
the outdated gdal lib on our servers only supported version 1.1.0.
Thus it was neccessary to add the VERSION parameter to the url

2. The Feature keys have changed (BEZIRK was renames to BEZNAME)


2e2a:
lgtm

PR #893, opened 2017-10-12
Adapt to core category filters

kleingeist:
This adapts meinBerlin to liqd/adhocracy4#193 

It furthermore refactors the ChoicesRequestMixin in the same way to get rid of the necessity to add project / module to the request.


2e2a:
lgtm

PR #895, opened 2017-10-12
 Add AreaSettings to Django Admin

kleingeist:
When a multimodule is created via the django admin, the settings are not initialized automatically. To be able to select an area for map based modules in the dashboard it is required to initialize the areasettings manually by creating an entry in the database.
Note that an empty string is not valid serialized json. To initialize the AreaSettings either enter false or "".

Note that this may become irrelevant after settings have been refactored.


meinberlin/apps/maps/admin.py
```diff
@@ -1,7 +1,32 @@
...
 from . import models
 
 # Register your models here.
 admin.site.register(models.MapPreset, admin.ModelAdmin)
 admin.site.register(models.MapPresetCategory, admin.ModelAdmin)
+
+
+@admin.register(AreaSettings)
+class AreaSettingsAdmin(admin.ModelAdmin):
+    description = 'Foo'
```
kleingeist: ups...

2e2a:
lgtm

PR #896, opened 2017-10-12
Enable categories and disable reporting for map|topicprio 

kleingeist:
The configuration option was missing.

Reporting was disabled for map|topicprio as the content is generated by the trusted initiators.


2e2a:
lgtm

PR #899, opened 2017-10-16
use radio buttons for private/public setting

### LINKED ISSUE #868: Make private project setting more clear
Currently the checkbox is kind of hidden and it is not instantly clear what is meant by "Access to Project". The setting should be more obvious and it's consequences more clear.

Opin has another way of showing the setting which we could/should adopt

xi:
fixes #868

copied from https://github.com/liqd/a4-opin/pull/874


kleingeist:
nice. way better then the checkbox

PR #901, opened 2017-10-16
change container header layout

xi:
The implementation is not nice, but ok.


xi:
Displaying logos in a responsive layout is still not fully solved. With decorative photography, we usually just scale and crop it. With logos this does not work as well. This PR improves the situation for containers (where we expect the tile images to be logos) but a more comprehensive solution will have to be done separately.

kleingeist:
i know you were talking with the PM about this, but for example the stadt machen logo looks really blurry when down scaled to 160x160.
not sure how to fix that though :| we never now the actual size the image might consume :|
but maybe we could increase the logo size a little more to something like 320 maybe? as you may have guessed this is about the default pixel space i have in the project list with my setup.

xi:
I changed the tile yet again. The image is no longer filled with white, so it will be as big as possible without cropping.

This works best for images with a white background. Maybe we can get them to use a modified version of their logo with a white background like this:

leitlinien_white

kleingeist:
+1 for the white background image

PR #904, opened 2017-10-17
improve project list in django admin

xi:
see https://docs.djangoproject.com/en/1.11/ref/contrib/admin/

this is merely an experiment to see what is possible


2e2a:
This is also a nice feature, that we should enable everywhere:
https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.date_hierarchy

2e2a:
merged into #908

PR #906, opened 2017-10-17
Add multiform views

kleingeist:
Multiform views may replace multiforms in a (hopefully) more simple manner.
The idea is to create and pass multiple forms to the template and to save them separately from the view. 

This is a first draft.

P.S. this PR also adds the is_archived flag to budgeting proposals.


kleingeist:
Note: if merged we may remove multi forms from the product requirements, too

meinberlin/apps/contrib/forms.py
```diff
@@ -0,0 +1,197 @@
...
+        if form_class:
+            return form_class
+        elif model is not None:
+            if fields is None:
+                raise ImproperlyConfigured(
+                    "Using MultiModelFormMixin (base class of %s) without "
+                    "the 'fields' attribute is prohibited."
+                    % self.__class__.__name__
+                )
+            return model_forms.modelform_factory(model, fields=fields)
```
2e2a: I think this function is a bit hard to read. Maybe there is a clearer way to distinguish the allowed setting combinations.

meinberlin/apps/contrib/forms.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def get_form_kwargs(self, name):
+        """
+        Returns the keyword arguments for instantiating the named form.
+        """
+        kwargs = super().get_form_kwargs(name)
+        instance = self.get_instance(name)
+        if instance:
+            kwargs.update({'instance': instance})
```
2e2a: Could we derive an instance from the model here, if it is `None`? To avoid doing this in the `get_instance()` implementation.

meinberlin/apps/budgeting/views.py
```diff
@@ -80,17 +84,50 @@ class ProposalDeleteView(idea_views.AbstractIdeaDeleteView):
...
     def get_success_url(self):
         return self.get_object().get_absolute_url()
 
-    def get_form_kwargs(self):
-        kwargs = super().get_form_kwargs()
-        kwargs['item'] = self.object
-        kwargs['creator'] = self.request.user
-        return kwargs
+    def forms_save(self, forms, commit=True):
+        objects = super().forms_save(forms, commit=False)
```
kleingeist: if commit=False 
save_m2m is not called on the forms

meinberlin/apps/contrib/forms.py
kleingeist: This is analogous to the core django get_form_class implementation. I would keep it that way for uniformity.

meinberlin/apps/contrib/forms.py
kleingeist: actually if instance is not set, it defaults to None for ModelForms and django creates an empty instance. i will write a comment and adapt the budgeting view

2e2a:
nice

PR #908, opened 2017-10-18
Enable django admin for more apps

2e2a:


meinberlin/apps/maps/admin.py
```diff
@@ -1,31 +1,7 @@
...
-                             'To initialize a new areasetting enter the '
-                             'string "false" without quotation marks.')
-        })
-    )
-
-    def module_name(self, setting):
-        return setting.module.name
-
-    def project_name(self, setting):
-        return setting.module.project
```
xi: Maybe it would be better to leave this here until it is merged into core.

xi:
LGTM.

meinberlin/apps/maps/admin.py
kleingeist: if you are already preparing a PR for core i think it is okay to delete it with this PR in meinBerlin

kleingeist:
lgtm :)
thanks for the work

PR #909, opened 2017-10-18
Use ServiceKonto API certificate if configured

2e2a:


meinberlin/apps/servicekonto/views.py
```diff
@@ -83,7 +84,13 @@ def _unstash_state(request):
 
 
 def _get_service_konto_user_data_xml(request, token):
-    service_konto = zeep.Client(settings.SERVICE_KONTO_API_URL)
+    transport = None
+    if settings.SERVICE_KONTO_CERT:
```
kleingeist: raises if the variable is not set:

> AttributeError at /accounts/servicekonto/callback/
> 'Settings' object has no attribute 'SERVICE_KONTO_CERT'

kleingeist:
lgtm :)

PR #910, opened 2017-10-18
Fix fakers

kleingeist:
Replace the "text" faker with the "sentence" faker in places where a name/title is expected. 
Using "text" may result in newlines which does not work with email subjects.
This is related to liqd/adhocracy4#200 

Also the broken and unused ContentTypeFactory has been removed.


2e2a:
lgtm

PR #912, opened 2017-10-18
Move container logic to meinberlin dashboard app

2e2a:


meinberlin/apps/projects/views.py
```diff
@@ -327,3 +328,8 @@ class DashboardProjectParticipantsView(AbstractProjectUserInviteListView):
     success_message_removal = _('Participant successfully removed.')
 
     invite_model = models.ParticipantInvite
+
+
+class DashboardProjectListView(a4dashboard_views.ProjectListView):
```
kleingeist: i'd prefer to put this to the dashboard/ app as it overwrites functionality from the dashboard2

kleingeist:
site note: just noticed, that we can delete `dashboard/filtersets.py` as they are no longer used

kleingeist:
lgtm and works with product

PR #915, opened 2017-10-23
fix hero-unit min-height in IE

xi:


kleingeist:
:) nice. looks good even on ie
@2e2a should we cherry pick this into the 1.6 branch?

PR #916, opened 2017-10-23
show datetime instead of date in timeline

### LINKED ISSUE #905: The time is not shown if project has not started yet
There is only the date shown if a project has not started yet. The best should be to show the time also, so people visiting a project’s page which should start some time on the same day are better informed and do not suspect an error.

bildschirmfoto 2017-10-17 um 11 45 24

This was tested on meinberlin-stage

xi:
if participation is not possible

fixes #905


kleingeist:
the date within the phase list is without times. but i think thats okay and will merge. if want to unify we can easily change the datetime format for the phases overview, too

PR #922, opened 2017-10-23
trigger change event when changing a polygon

### LINKED ISSUE #918: Missing question if user wants to leave the page when editing polygons

xi:
this is required to trigger the unload warning. Fixes #918


kleingeist:
lgtm :)

PR #923, opened 2017-10-24
Allow moderators to edit projects

2e2a:


kleingeist:
sligthly off topic:
while looking at this PR i realized, that the `DashboardBaseMixin` is kind of old and duplicates the `ProjectMixin` logic. maybe we should rewrite it to check the `organisation_slug` and if thats not available the `self.project` from the `ProjectMixin`. 
furthermore i think we should keep the Permission checking out of the mixin, as we are overwriting it in mostly and it makes the permissions even more unreadable. at least we should not return "project" as the permission object, as it is not set by the mixin at all. 
for some kind of security we could keep the PermissionRequiredMixin dependency, but overwrite the get_permission_object with a default that raises an "NotImplemented" Exception or something like that if it is not overwritten by the child classes.

2e2a:
Good point. Refactored the dashboard mixin accordingly.

kleingeist:
if moderators should be able to modify (or complete) projects it should be possible to add moderators to projects that are still in draft mode. Although this may be weird, as the link in the email points to the unfinished and probably weird looking project.

PR #925, opened 2017-10-24
Idea view integration tests

kleingeist:
these tests are for detecting failures that result from code regressions (like wrong template tags, or access to in-existent view properties from within views) and merely check if the page is rendered with a 200 OK status.
it is tried to include in these test cases as many features as possible (e.g. categories with ideas)


xi:
> access to in-existent view properties

Just to be sure: If I remember correctly, django templates return None instead of raising Key- or AttributeErrors, don't they?

kleingeist:
> Just to be sure: If I remember correctly, django templates return None instead of raising Key- or AttributeErrors, don't they?
yes, the second part of the statement was in regard to views where python throws an exception if a property is missing.

those tests are very very broad, the don't prevent from non functional pages in regard to the user experience. they only prevent ugly 5xx errors which we missed previously when we refactored some parts of the platform

meinberlin/apps/maptopicprio/models.py
```diff
@@ -33,19 +46,6 @@ class MapTopic(module_models.Item):
...
-    slug = AutoSlugField(populate_from='name', unique=True)
-    name = models.CharField(max_length=120, verbose_name=_('Title'))
-    description = RichTextUploadingField(config_name='image-editor')
-    ratings = GenericRelation(rating_models.Rating,
-                              related_query_name='idea',
-                              object_id_field='object_pk')
-    comments = GenericRelation(comment_models.Comment,
-                               related_query_name='idea',
-                               object_id_field='object_pk')
-    category = CategoryField()
```
xi: Best fix ever

tests/budgeting/test_views_integration.py
```diff
@@ -0,0 +1,74 @@
...
+        client.login(username=user.email, password='password')
+        data = {
+            'name': 'Idea',
+            'description': 'description',
+            'category': category.pk,
+            'budget': 123,
+            'point': (0, 0),
+            'point_label': 'somewhere'
+        }
+        response = client.post(url, data)
```
kleingeist: i would also add a client.get() as we might miss template errors on the formpages.

kleingeist:
@2e2a i did change a lot again. could you take a look at it again and merge if you are okay?
especially check if the clear_caches fixture is okay: https://github.com/liqd/a4-meinberlin/pull/925/commits/6cc770e65e433b25b6c4518a70c6348c992d6772

PR #926, opened 2017-10-24
add select2

xi:
This is a proposal to add [select2](https://select2.org/getting-started/basic-usage) as a dependency. It greatly improves usability of select elements, especially for multi-selects and if there are many options. It even has decent accessibility!

On the downside, it depends on jquery and adds about 200kb to our frontend bundles (increase by 12% for js/css).

I have been thinking about adding this for a while. The trigger to do this now is the bad usability of the multi select element that is used to select projects in containers. We tried to imrpove that in 91470562b by adding resizing, but that is not enough and actually not supported in IE and safari.

### DIFF
--- a/meinberlin/apps/contrib/widgets.py
+++ b/meinberlin/apps/contrib/widgets.py
@@ -1,0 +1,26 @@
+from django import forms
+from django.contrib.staticfiles.storage import staticfiles_storage
+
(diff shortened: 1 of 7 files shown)


kleingeist:
so dynamic! like! :)

PR #927, opened 2017-10-24
overwrite datetime template

xi:
Fixup to ef88f2b


kleingeist:
should be done in the product, too

PR #929, opened 2017-10-24
allow admins to create modules in dashboard

xi:
This is mostly a copy-and-paste job that takes most of its code from project creation.

Note that the views are protected only with the `a4projects.add_project` permission, but the button to initiate module creation is only shown for superusers. This is because the process still has some rough edges, e.g. it is not possible to delete a module in the UI. You still need access to django admin for that.

One challange is that we do not have module types. So once you get back to the dashboard, all information about the blueprint is lost. This can make it hard to know which module is which. My strategy was to use the blueprint title as module name. Not sure if this is a good idea.

**This assumes that project blueprints and module blueprints are equivalent and that wording works for both cases.**


kleingeist:
Note that this will probably conflict with #923

meinberlin/apps/dashboard/urls.py
```diff
@@ -38,4 +38,11 @@
     url(r'^organisations/(?P<organisation_slug>[-\w_]+)/projects/$',
         views.DashboardProjectListView.as_view(),
         name='project-list'),
+    url(r'^projects/(?P<project_slug>[-\w_]+)/blueprints/$',
```
kleingeist: we should move those to url to the top and rename the block from "# Overwrite the ProjectUpdateView with meinBerlin urls" to "# Overwrite adhocracy4 core urls with meinBerlin urls"

meinberlin/apps/dashboard/views.py
```diff
@@ -52,3 +59,64 @@ def get_success_url(self):
...
+            phase = phase_models.Phase(
+                type=phase_content.identifier,
+                name=phase_content.name,
+                description=phase_content.description,
+                weight=phase_content.weight,
+                module=module,
+            )
+            phase.save()
+
+    def get_next(self, module):
```
kleingeist: i would prefer `get_success_url` which is normally used by us (due to the FormMixin)

kleingeist: actually i take it back.this is not in the context of a module, so we have to pass it along. sorry

kleingeist:
nice :) although we are moving away from the whole blueprints are the only origin of projects. but i think thats a great improvment over django admin

PR #932, opened 2017-10-24
Filter deleted, archived and private projects from wagtail block

### LINKED ISSUE #235: deleting a featured project causes a 500 error on home
^

2e2a:
Fixes #235

A cleaner fix would be in the block directly, but fixing template was simpler.


kleingeist:
kind a hacky, but i like it. and it solves the problem. modifying wagtail blocks did not work out fine at all :(

PR #936, opened 2017-10-25
Allow user invites via file upload

kleingeist:
The uploaded file has to be a utf8 compatible text file that may contain emails. 
The addresses are scraped without further regard to a maybe existent logical document structure. 
While everything that looks like an email is found only valid emails are returned.


2e2a:
nice

Issue #750, opened 2017-08-30
Switch between map and list view is not intuitive


2e2a:
@hmkay did not see that there is a list view in the map projects

damusp:
Hmm…
I too get the feeling that this needs to be much more obvious. 

Making the button stand out more (most likely with color, maybe with size and positioning?) would help a bit. If we want to get further with it, we would need to make some layout changes…

kleingeist:
there is a new concept to merge the list and map view in PR #937

PR #937, opened 2017-10-25
combined map/list

xi:
First, some general notes about this topic:

-   We need to show map and list next to each other. This can be difficult on small screens.
-   We want filters to work with both
-   If simple, we want the map to act as a filter for the list (only items that are visible in the current map excerpt are included)
-   It is not clear what will happen if you click on a marker/list item: Will it mark the current item as selected or will it link to the detail page directly?

This PR is a low-fi solution to the problem. It simply displays a map with markers for each list item next to the list. When you click on a marker, the corresponding list item is selected. If necessary, it is also scrolled into view. The list item itself links to the detail page.

Note that all filters are still applied server-side. So when I activate a filter a request is triggered and the result again contains a list with a corresponding map.

In terms of accessibility this means that users who cannot use the map can ignore it for the most part, as the list is not changed at all.

I consider this PR as a starting point for discussion rather than a finished implementation.


xi:
Related to #750

xi:
Closing in favour of #984

PR #938, opened 2017-10-30
change container icon

xi:
I am not totally convinced of this one. The candidates are.

-   http://fontawesome.io/icon/folder-open/ - what we use now
-   http://fontawesome.io/icon/clone/ - doesn't really work semantically
-   http://fontawesome.io/icon/th-large/ - looks like the windows logo
-   http://fontawesome.io/icon/th/ - okayish

I went with `th`, but it does not really looks centered in chromium (it looks fine in firefox).


kleingeist:
for reference a screenshot on chrome 
container-new-icon

not sure if i should merge or not. how do we proceed?

PR #939, opened 2017-10-30
show warning on draft project detail

xi:
I also took the time to change the styling of our alerts. Before, alerts were often made up ad-hoc and were very inconsistent. With this PR they now look a lot like the [bootstrap alerts](https://getbootstrap.com/docs/4.0/components/alerts/). This includes adjusting some colors, most notably `$brand-danger` which is now a bit darker and therefore more distinguishable from our `$brand-primary` magenta.


2e2a:
lgtm

PR #940, opened 2017-10-30
allow to filter user list in admin by last login

xi:
This allows to see which users are actually active.

The default django date filter is surprisingly good for our purpose here.


2e2a:
nice

PR #942, opened 2017-10-30
make sure next parameters are urlencoded

fuzzylogic2000:
We had a problem in the user profiles in ae and opin: https://github.com/liqd/a4-advocate-europe/issues/527

Currently there shouldn't be a problem in meinBerlin, but just to make sure, this PR urlencodes all next-parameters.

### DIFF
--- a/meinberlin/apps/users/templates/meinberlin_users/indicator.html
+++ b/meinberlin/apps/users/templates/meinberlin_users/indicator.html
@@ -37,7 +37,7 @@
         <li>
             <form class="u-inline" action="{% url 'account_logout' %}" method="post" aria-label="{% trans 'Logout' %}">
                 {% csrf_token %}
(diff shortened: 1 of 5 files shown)


meinberlin/apps/users/templates/meinberlin_users/indicator.html
```diff
@@ -37,7 +37,7 @@
         <li>
             <form class="u-inline" action="{% url 'account_logout' %}" method="post" aria-label="{% trans 'Logout' %}">
                 {% csrf_token %}
-                <input type="hidden" name="next" value="{{ redirect_field_value }}">
+                <input type="hidden" name="next" value="{{ redirect_field_value|urlencode }}">
```
xi: I think it gets url-encoded automatically when used in an input field.

PR #943, opened 2017-10-30
avoid inline script

xi:
This will allow us to apply a strict content security policy in the future.

See https://github.com/liqd/adhocracy4/issues/86.

~~Depends on https://github.com/liqd/adhocracy4/pull/207.~~


kleingeist:
classy pr. very british :)
very british

i really had to look up  initialise as i was so sure it would be wrong. but it seems as if the british are never wrong. also it is used in opin, too.

PR #947, opened 2017-10-30
Revert part of 3d8b15e

xi:
This is a fixup to #942

Steps to reproduce the issue:

-   Go to http://localhost:8000/?foo
-   Log in/out
-   With this path you will end up back on http://localhost:8000/?foo
-   Without this patch, you will end up on http://localhost:8000/%3Ffoo because the URL was encoded twice


2e2a:
nice

PR #954, opened 2017-11-01
update npm dependencies

### LINKED ISSUE #789: IE11: broken shariff styling
screenshot from 2017-09-13 14-16-25

Shariff does not render correctly on IE11. I think this is because some classes are missing (e.g. `facebook`, `fa-facebook`). Looking through the shariff code I think this might be connected to calls like `$(…).addClass('fa fa-facebook')` where only the first class seems to be added.

xi:
especially shariff, which fixes #789.


2e2a:
lgtm

Issue #953, opened 2017-11-01
Confusing alerts on invite if emails are already invited


xi:
I uploaded the following CSV to invite moderators (see #936):

```
hallo,bar@example.com,asd,01.01.2019
hallo,baz@example.com,asd,01.01.2019
```

It took quite a while and in the end I got two alerts:

- warning: `Die folgenden Nutzerinnen und Nutzer wurden bereits eingeladen:baz@example.com, bar@example.com`
- success: `0 moderators invited.`

2e2a:
Could not reproduce locally and on dev

kleingeist:
the "Die folgenden Nutzerinnen und Nutzer wurden bereits eingeladen:" message is from the existing behaviour and has nothing to do with file uploads. if you enter an already invited email adress to the input field the same messages appear. i'll adapt the the issue title accordingly.

for the second issue abut slow upload: where did you test? locally? can you reproduce it on the dev machine? how long did it take? if it is reproducible slow on the dev machine we should add another issue.

kleingeist:
actually there is also an issue with the "pending" emails.
if an email is [found in the inivite_model ](https://github.com/liqd/a4-meinberlin/blob/1edc8cd48a1e325e91a44e66009f8444e1d384c0/meinberlin/apps/projects/views.py#L255) it is filtered. 
But it is not checked if the email adddress is invited to the same or a different project. Thus it is not possible to invite the same email to different projects. This should be fixed, but is probably another issue, as the alerts may still be confusing.

xi:
Fixed with #957

PR #957, opened 2017-11-02
Filter pending emails on project

kleingeist:
as found out when investigating #953 pending email invites had not been filtered by project. thus an email could be invited only to a single project. this pr fixes this by filtering the invite on the project.


2e2a:
lgtm

PR #958, opened 2017-11-02
disable poll for unauthenticated user

fuzzylogic2000:


meinberlin/assets/scss/components/_radio.scss
```diff
@@ -1,6 +1,10 @@
 
 .radio__input {
     @include sr-only;
+
+    &:disabled + .radio__text {
```
fuzzylogic2000: @xi May I use the sibling-selector in meinBerlin? And what should happen to the colours?

xi: In general, I think that sibling selectors are ok, but should be avoided. In practice this means that you can use them, but are among the first things to be refactored.

In this specific case, I don't think this styling is useful. We do not generally change the styling of labels for disabled inputs. [Bootstrap](https://getbootstrap.com/docs/4.0/components/forms/#disabled-forms) also does not do this.

fuzzylogic2000: Ok, so it's not clickable, but I can't see that before I click. That seems a bit annoying, but if that's the way to do it, I will of course do it like that.

meinberlin/assets/js/app.js
```diff
@@ -43,6 +43,9 @@ var dynamicFields = require('../../apps/contrib/assets/dynamic_fields.js')
 var getCurrentHref = function () {
   return location.href
 }
+var getCurrentPath = function () {
+  return location.pathname
+}
```
xi: Unfortunately this is not that simple. `getCurrentHref()` is part of the embed API (see comment above). Changing this means extending that API.

xi:
Nice!

PR #961, opened 2017-11-02
Fix maptopicprio field order

2e2a:
Point needs to be defined after item_ptr, as its pk is required by the json
field to convert the string to json.


xi:
This has been cherry-picked from #925 (3309997)

kleingeist:
bad bug :bug: !

PR #963, opened 2017-11-06
 List container projects only if user may view them

kleingeist:
The project list within containers will be filtered by the permission of
the currently logged in users. Only projects that may be viewed by the
user are listed.

Furthermore this adapts the number of active and total projects.
For public containers it only counts published projects that are public
and not archived.
In the future (if containers may be private) the count for private
containers should reflect the total number of published, not archived
projects.

Furthermore this makes  the ORM project list filtering reusable.

### DIFF
--- a/meinberlin/apps/projects/query.py
+++ b/meinberlin/apps/projects/query.py
@@ -1,0 +1,21 @@
+from django.db.models import Q
+
+
+def filter_viewable(queryset, user):
+    # FIXME: has to be in sync with a4projects.view_project and should
+    #        be implemented on the Project's QueryManager/QuerySet.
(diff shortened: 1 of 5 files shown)


meinberlin/apps/projects/query.py
```diff
@@ -0,0 +1,21 @@
...
+        return queryset.filter(
+            Q(is_public=True) |
+            Q(participants__pk=user.pk) |
+            Q(organisation__initiators__pk=user.pk) |
+            Q(moderators__pk=user.pk)
+        )
+    else:
+        return queryset.filter(
+            is_public=True
+        )
```
2e2a: Nice. Much more intuitive than the query before.

PR #965, opened 2017-11-06
open link to container outside of embed

xi:
-   If you embed a project that is inside of a container, the link back to the container should be treated as an external link.
-   If you embed a container, the link from a project back to the container should be handled as an internal link.

As we do show embed code for projects but not for containers, I think it is reasonable to ignore the second case.


2e2a:
lgtm

PR #966, opened 2017-11-06
2017 11 ka add multiple choice option to poll

fuzzylogic2000:
- adds field to model

- enables to set field in dashboard


xi:
LGTM.

xi:
Just one request: There should be a better label/help text in the dashboard.

fuzzylogic2000:
@xi Yes! What do you propose?

PR #969, opened 2017-11-06
Send signals when components update

kleingeist:
A signal is send if the form did update without raising an error on save. 
This also extends the existing signals by the user who triggered the action.

### DIFF
--- a/meinberlin/apps/dashboard2/mixins.py
+++ b/meinberlin/apps/dashboard2/mixins.py
@@ -9,11 +9,13 @@ from django.shortcuts import get_object_or_404
 from django.shortcuts import redirect
 from django.utils.translation import ugettext_lazy as _
 from django.views.generic import base
+from django.views.generic.edit import FormMixin
 
 from adhocracy4.projects import models as project_models
 from adhocracy4.rules import mixins as rules_mixins
 
 from . import get_project_dashboard
+from . import signals
 
 Organisation = apps.get_model(settings.A4_ORGANISATIONS_MODEL)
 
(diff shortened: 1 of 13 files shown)


kleingeist:
i did use the following to log he signals for testing

```
def log_signal(sender=None, **kwargs):
    print("Signal: ", kwargs)

project_created.connect(log_signal, weak=False)
module_created.connect(log_signal, weak=False)
project_pre_publish.connect(log_signal, weak=False)
project_published.connect(log_signal, weak=False)
project_unpublished.connect(log_signal, weak=False)
project_component_updated.connect(log_signal, weak=False)
module_component_updated.connect(log_signal, weak=False)
```

meinberlin/apps/dashboard2/mixins.py
```diff
@@ -95,6 +97,24 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class DashboardComponentUpdateSignalMixin(FormMixin):
+    def form_valid(self, form):
+        from . import components
```
kleingeist: ma yprobably moved to top

meinberlin/apps/maptopicprio/views.py
```diff
@@ -156,3 +159,12 @@ def get_success_url(self):
...
         return self.get_object()
+
+    def delete(self, request, *args, **kwargs):
+        response = super().delete(request, *args, **kwargs)
+        a4dashboard_signals.module_component_updated.send(
+            sender=None,
+            module=self.module,
+            component=self.component,
+            user=self.request.user)
+        return response
```
2e2a: Could we extend the DashboardComponentUpdateSignalMixin to handle the delete too?

kleingeist: good idea, this would overwrite the DeletionMixin

meinberlin/apps/polls/serializers.py
```diff
@@ -69,6 +76,15 @@ def _update_choices(self, question, question_instance):
                 }
             )
 
+    def _send_component_updated_signal(self, question_instance):
+        component = components.modules['polls']
+        a4dashboard_signals.module_component_updated.send(
+            sender=None,
```
2e2a: sender is None

PR #973, opened 2017-11-06
UI adjustments for plans

xi:


xi:
> Somehow the map form field for the plan create view got lost between master and here, but I can't find how.

That was fixed separately in #972. This branch is actually older than the fix.

fuzzylogic2000:
Oh, sorry! Everything else looked fine!

PR #977, opened 2017-11-06
maptopicprio: hide places dashboard component if polygon is not set

### LINKED ISSUE #970: Maptopicprio: Polygon may not be set
When I try to create a maptopic in the dashboard, it is not guaranteed that a polygon is already set. The map for setting a point therefore fails to display.

xi:
fixes #970

This may not be the best approach from a usability perspective, but it is simple and better than just showing a broken map widget.


2e2a:
nice

PR #979, opened 2017-11-07
Move module add views to core dashboard

2e2a:
The button was added to the core template, so moving the views there too. Fixes a4-product dashboard.


kleingeist:
lgtm.
we will see what happens when we try to merge it to the core :D

PR #980, opened 2017-11-07
use has_or_would_have permission logic in polls

fuzzylogic2000:


xi:
We definitely want to show results when the module is finished. I am not so sure that this is also true if users simple are not allowed to vote. Then again, what else would we show? So I think this is completely fine. Nice work!

PR #981, opened 2017-11-07
Notify initiators on project creation

2e2a:


kleingeist:
lgtm. wording needs to be checked back with the pm but thats easier when the changes are deployed on dev.

PR #982, opened 2017-11-07
mark all make targets as phony

xi:


kleingeist:
lgtm. altough a little phony. but i still trust you ;)

PR #983, opened 2017-11-07
Create Admin Log in DB

kleingeist:
The Admin Log should contain information about projects modified from within the dashboard.
The app listens to the dashboard signals and creates localized database log entries for them.
The log entries may be viewed in the django admin.


2e2a:
Is there a way to have to show the translated component names, instead the identifier?

2e2a:
> Is there a way to have to show the translated component names instead the identifier?

My mistake. Is translated.

Nice solution.

PR #984, opened 2017-11-07
dynamic combined list/map

xi:
See also #937

~~This PR is incomplete, even as a spike. Expect rebases! I just pushed so you can see the progress.~~

This version tries to use leaflet within react. I took some inspiration from [react-leaflet](https://github.com/PaulLeCam/react-leaflet/). I think we would only need a fraction of that project, so I tried to duplicate the relevant bits instead.

My main concern here is the interplay between filtering, scrolling the list, zooming/moving the map, and selecting an item.

-   Scrolling
    -   Scrolling should not have any effect on anything else. Users expect scrolling to be non-destructive.
    -   When you filter, the scrolling position must necessarily change. If the selected item is still in the list, it should be scrolled into view. Otherwise, the list should scroll back to the top. (`filter -> scroll`)
    -   When you select an item on the map, it should be scrolled into view. (`select -> scroll`)
-   Zooming
    -   Zooming is very similar to scrolling. However, we want the map's bounding box to act as a filter to the list. This changes quite a bit.
    -   When you zoom, you also filter the list, so the effect on scrolling is the same as discussed above (`zoom -> scroll`)
    -   When you filter, the current part of the map may be empty. In order to avoid confusion it would be a good idea to reset the zoom on filter. (`filter -> zoom`)
    -   When you select an item in the list, it is already guaranteed that its marker is in the visible part of the map. So zooming is not required. However, it may still be a nice feature to zoom in on the selected item. We need to discuss wether this outweighs the added complexity. (optional: `select -> zoom`)
-   When filtering (or zooming, which also is a kind of filter), the selected item may be excluded from the list. We might want to un-select it then. However, I believe that it would be better to keep it selected: Imagine a user moving the map so that the selected item is excluded and then moving it back. It would be nice if the item would still be selected. (optional: `filter -> select` and/or `zoom -> select`)

Other notes:

-   Sorting adds even more complexity. I thing it is sufficient to define one default sorting and not allow users to change it (e.g. by date)
-   Pagination does feels strange with a map. We will probably have to show everything. If performance gets too bad we will have to restrict the list to the `n` latests entries or something like that.


meinberlin/assets/js/plans_map.jsx
```diff
@@ -92,14 +96,16 @@ class PlansMap extends React.Component {
         <ul className="u-list-reset map-list-combined__list">
           {
             this.props.items.map((item, i) => {
-              let className = 'list-item'
-              if (i === this.state.selected) {
-                className += ' selected'
+              if (this.isInFilter(item)) {
```
kleingeist: leaflet layers/markers should be filtered, too

kleingeist:
very nice :) i like it and think it's fine to merge to get a broader audience for testing

PR #991, opened 2017-11-09
Always activate participation tab first

2e2a:
Reverts https://github.com/liqd/a4-meinberlin/pull/756

... as is breaks offline-event navigation.


kleingeist:
i did like the idea, but we should re-evaluate the behavior. maybe we can find a solution working for projects and events.

PR #993, opened 2017-11-13
Option to include email-form responses directly into email

2e2a:


meinberlin/apps/cms/templates/meinberlin_cms/emails/form_submission.en.email
```diff
@@ -7,4 +7,12 @@
 
 {% block content %}
     {{ emailformpage.email_content }}
+
+    {% for key, value in field_values.items %}
+
+        {{key}}:
+
```
kleingeist: i would skip this empty line. looks weird and makes the field (keyvalue) blocks less distinguishable

kleingeist:
i am not sure, if it is understandable that selecting "-----" will send the input within the body of the email.

PR #994, opened 2017-11-13
Adminlog dashboard component

2e2a:


kleingeist:
not shown for external and bplan projects (the admin component has to be returned from the TypedDashboard)
and i just realized, that we are probably missing a log entry for bplan api modifications.

kleingeist:
oh and within this context it may look weird if a project / module is renamed, as the old log messages will not reflect the changes. but maybe this is only a wording issue.

2e2a:
> oh and within this context it may look weird if a project / module is renamed, as the old log messages will not reflect the changes. but maybe this is only a wording issue.

I would ignore this issue in favor of having simple and fixed log messages. We can do it later, when requested.

PR #995, opened 2017-11-13
moved test factories to apps/test

fuzzylogic2000:


kleingeist:
we probably have to use the `A4_ORGANISATION_FACTORY` and the `A4_USER_FACTORY` in every factory we want to reuse in the product, as the product has different models from meinberlin

kleingeist:
lgtm :)

PR #997, opened 2017-11-13
added styling for checkboxes for polling

fuzzylogic2000:
While this works fine and styles the checkboxes analogous to the radio buttons, it didn't seem right to copy the css and only change the border-radius. @xi What do you think/how would you do that?


meinberlin/assets/scss/components/_checkbox.scss
```diff
@@ -0,0 +1,73 @@
...
+    @include fake-focus-shadow;
+}
+
+.checkbox__input:disabled + .checkbox__text {
+    &:before,
+    &:after {
+        background-color: $border-color;
+        cursor: not-allowed;
+    }
+}
```
xi: I would prefer to use a variant like `.radio__text--checkbox` instead of duplicating all this code. This also makes sense semantically: We want these two components to look alike.

xi:
Much wow!

PR #999, opened 2017-11-14
Enable clustering on plans

kleingeist:
The clusters are basically working.
If many clusters are on the same place, the may be spiderfied onclick when the highest zoom level is selected.

### DIFF
--- a/meinberlin/assets/js/plans_map.jsx
+++ b/meinberlin/assets/js/plans_map.jsx
@@ -105,11 +105,9 @@ class PlansMap extends React.Component {
     if (prevState.selected !== this.state.selected) {
       if (prevState.selected !== null) {
         this.markers[prevState.selected].setIcon(icon)
-        this.markers[prevState.selected].setZIndexOffset(0)
       }
       if (this.state.selected !== null) {
         this.markers[this.state.selected].setIcon(activeIcon)
-        this.markers[this.state.selected].setZIndexOffset(1000)
       }
     }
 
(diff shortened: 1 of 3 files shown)


xi:
I tried to clear the cluster layer completely and adding in the visible markers again as recommended in the [docs](https://github.com/Leaflet/Leaflet.markercluster#bulk-adding-and-removing-markers), but that resulted in an unstable clustering (selecting the same marker could result in different clustering).

I also tried to always draw the selected marker on top of everything else. My approach was to add the `selectedMarkers` layer to a different [pane](http://leafletjs.com/reference-1.2.0.html#layer-pane), but that seems to be broken (or at least did not work as I had expected).

xi:
In a3 there was also some custom styling for clusters: https://github.com/liqd/adhocracy3/blob/54d4df7143e34d82e98b969c3ec6a7ecaf61bc1d/src/adhocracy_frontend/adhocracy_frontend/static/stylesheets/scss/components/_map.scss#L87

xi:
z-index does not work properly for me. Steps to reproduce:

-  Create several markers on the exact same spot
-  Select one
-  Zoom out

For me, the z-index seems to change with the zoom level.

kleingeist:
i can, kind of reproduce, 
the zIndex is reset if a marker is clicked when it is part of a spiderfied cluster. 

Steps to reproduce:
* Create several markers on the exact same spot
* Zoom In to the max
* Click cluster to spiderfy markers
* Click marker 
  -> zIndex is not set :(
* Zoom Out

kleingeist:
@xi i did fix the zIndex by setting it after the marker is removed from the cluster
furthermore i changed the code to only style and move the marker between layers if it is not already there.
this is based on the assumption, that the state from stored in the layer information is always correct. not sure about this, as it bypasses reacts state :(

xi:
Nice!

PR #1007, opened 2017-11-15
add component library

xi:
A component library has 3 important benefits:

-   It documents all available components. This also means establishing consistent naming that is accessible for all stakeholders, including designers and clients.
-   You can think of it as "unit tests for CSS" as you can simple look at the component library in order to see if any component is broken. You can also add "regresion tests" if you find that a specific combination of classes causes issues.
-   Creating a component library is an opportunity to revisit all the CSS code. This means that we can get fimiliar with the parts of the code that we have not looked at for a long time. We may also find opportunities for refactoring.

To get more background information on component libraries and component-base web design in general, I strongly recommend the first chapter of [Atomic Design](https://atomicdesign.bradfrost.com/chapter-1) by Brad Frost. Also the [bootstarp's component library](https://getbootstrap.com/docs/) is a good reference and inspiration.

I would like to do a workshop on this topic with the whole team.


kleingeist:
> It might be a good idea to make this more self-explanatory, maybe with description texts.

yes! but we will address this in a different PR. presumably ;)

PR #1008, opened 2017-11-15
refactor alert styling

xi:


kleingeist:
lgtm. would merge.

but out of curiosity: why did you use `alert--error` etc. instead of [`alert-error` as bootstrap does](https://v4-alpha.getbootstrap.com/components/alerts/)?

xi:
We use BEM. There are other cases already where we use BEMified versions of bootstrap class names.

PR #1009, opened 2017-11-15
Translations

2e2a:


locale/de/LC_MESSAGES/django.po
```diff
@@ -465,7 +465,7 @@ msgstr "Bereich"
 
 #: meinberlin/apps/adminlog/dashboard.py:13
 msgid "Log"
-msgstr ""
+msgstr "Log"
```
xi: Historie? Änderungsprotokoll? Verlauf?

locale/de/LC_MESSAGES/django.po
```diff
@@ -1157,40 +1157,40 @@ msgstr "Organisationsinformationen bearbeiten"
 
 #: meinberlin/apps/dashboard/templates/meinberlin_dashboard/plan_confirm_delete.html:4
 msgid "Confirm Plan Deletion"
-msgstr ""
+msgstr "Bestätigung der Vorhabenlöschung"
```
xi: Sounds strange, but is consistent with similar cases.

kleingeist:
lgtm. 
just noticed something general. we are translating "User" as "Nuzer und Nutzerinnen" or "Nutzer*innen" when it is used in a sentence, but we translate "User" as single string as "Benutzer"

PR #1010, opened 2017-11-15
extend plan model

xi:
This adds many more fields to the plans model. We are not quite sure if these fields will be in the final version. However, we definitely want to add them now so we can get better feedback from the clients.

The only field that has been omitted is "timespan" as the format is very unclear. A human readable version of that information can be included in the description for now.


meinberlin/apps/plans/models.py
```diff
@@ -22,6 +33,30 @@ class Plan(UserGeneratedContentModel):
...
+        blank=True,
+        default='',
+        max_length=255,
+        verbose_name=_('Label of the location'),
+        help_text=_('This could be an address or the name of a landmark.'),
+    )
+    contact = models.TextField(max_length=255, verbose_name=_('Contact'))
+    cost = models.PositiveIntegerField(blank=True, null=True,
+                                       verbose_name=_('Cost'))
+    description = RichTextField(verbose_name=_('Description'), blank=True)
```
kleingeist: i think richtextfields need some additional handlin in the save method:
see https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/ideas/models.py#L44

although i don't now why o.0

kleingeist: i did create issues on that topic #1011

PR #1018, opened 2017-11-16
Style markercluster and improve printing of plans

kleingeist:
The styling is analogues to the bplan map: https://github.com/liqd/django_zbp/blob/master/bplan/static/scss/components/_markers.scss


xi:
LGTM

PR #1019, opened 2017-11-16
Select map marker on list click

kleingeist:
Simply selects a marker when users click on the list item.


xi:
This solution is not accessible for keyboard users. (Note that the map is only relevant for sighted users. However, there are users who are sighted *and* need to use the keyboard.)

I think this is an improvement. However, I am reluctant to merge this as it would tempt us to not do a proper implementation.

To enable keyboard support, we would need an element that can receive focus. For example, we could add a button to the list item. Alternatively, we could use the `tabindex` property to make the list item itself focussable.

kleingeist:
@xi thanks for the feedback. i did try to make it accessible for keyboard users by giving a tabindex=0 to the li and select the marker on focus. Thus the behavior is the same for mouse users, but keyboard users may also navigate the list and see the associated highlighted marker. 
but i am not sure if this solution is preferable to an additional button. but for me as a mouse centric visual user it behaves as i would expect it.

xi:
Nice!

Issue #948, opened 2017-10-30
Activate cache busting


xi:
Django has built-in cachebusting via [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/#manifeststaticfilesstorage). AFAIK we already use `collectstatic` in production, so it would be a simple matter of adding the following line to production.py:

```
STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'
```

However, I got the following error:

```
ValueError: The file 'a4images/imageUploader.js' could not be found with <django.contrib.staticfiles.storage.ManifestStaticFilesStorage object at 0x7f5f3d33c780>.
```
I got it to work by running `collectstatic` first with the default storage and than again with this one. But I don't think this is how it was intended to be used.

fuzzylogic2000:
This works, right?

fuzzylogic2000:
Hmm, or not that well with running collectstatic twice?! We should also do it like that: https://github.com/liqd/a4-opin/pull/1008/files

PR #1021, opened 2017-11-16
ManifestStaticFilesStorage Cachebusting

2e2a:
Trying out cachebusting provided by djangos `ManifestStaticFilesStorage`  as suggested in https://github.com/liqd/a4-meinberlin/issues/948.

Notes:
-  `collectstatic` needs to be called twice. Once with the default storage and once with  the ManifestStaticFilesStorage.
- The strict standard ManifestStaticFilesStorage did not work. As far I understood, this works only when the manifest file is predefined.
- Worked for CSS, Images and JS, but did break our react components.


2e2a:
Will merge and try it out on dev.

PR #1031, opened 2017-11-20
disable circlemarker in leaflet draw

### LINKED ISSUE #1028: Remove circlemarker tool from Areasettings map
Somehow the circlemarker tool from leaflet-draw made it in our polygon widget. 
This is probably due to a leaflet update.

leaflet-draw-circlemarker

xi:
fixes #1028


kleingeist:
lgtm

PR #1032, opened 2017-11-20
add 'change vote' on poll if user already voted

fuzzylogic2000:
The design says to write 'Stimme ändern' if user already voted.
I also refactored the footer generation a bit and hope to make it more readable?!

### DIFF
--- a/meinberlin/apps/polls/assets/Question.jsx
+++ b/meinberlin/apps/polls/assets/Question.jsx
@@ -128,38 +128,31 @@ class Question extends React.Component {
     const total = this.state.question.totalVoteCount
     const max = Math.max.apply(null, this.state.question.choices.map(c => c.count))
 
-    let showResultOrVote
+    let showTotalOrVoteButton
     let toggleShowResultButton
     let toggleShowResultButtonText
 
     if (this.state.showResult) {
-      showResultOrVote = (
+      showTotalOrVoteButton = (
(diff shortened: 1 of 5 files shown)


meinberlin/apps/polls/assets/Question.jsx
```diff
@@ -128,29 +128,39 @@ class Question extends React.Component {
...
-              { django.gettext('To poll') }
-            </button>
-          }
-        </div>
+      showResultOrVote = (
+        `${total} ${django.ngettext('vote', 'votes', total)}`
       )
+      toggleShowResultButtonText = (
+        `${django.gettext('To poll')}`
+      )
```
xi: Why did you wrap this string in parantheses?

meinberlin/apps/polls/assets/Question.jsx
```diff
@@ -128,29 +128,39 @@ class Question extends React.Component {
...
+      showResultOrVote = (
+        <span>
           {this.getVoteButton()}
-          &nbsp;
-          <button type="button" className="btn btn--link" onClick={this.toggleShowResult.bind(this)}>
-            { django.gettext('Show preliminary results') }
-          </button>
-        </div>
+        </span>
+      )
```
xi: Why did you wrap this in a `<span>`?

PR #1034, opened 2017-11-20
add plan filters

xi:
~~Styling still missing~~

UI can still be improved, but I propose to do that separately:

- [X] The select element is a bit smaller than the free text input.
- [ ] Every option in the select element is currently prefixed with "Status: ".


kleingeist:
nice :)

p.s. one small thing. we never show the participation status. neither in the list, nor in the detail. but i will add a task for that

PR #1035, opened 2017-11-20
add edit link to plan detail

xi:
I know this is a lot of HTML. This is simply how we do edit links in the
item-detail UI.

I am not completely sure which permission needs to be checked here.

### DIFF
--- a/meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
+++ b/meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
@@ -72,7 +72,7 @@
                             <ul class="dropdown-menu" aria-labelledby="idea-{{ object.pk }}-actions">
                                 <li>
                                     <a
-                                        href="{% url 'a4dashboard:plan-update' organisation_slug=object.project.organisation.slug slug=object.slug %}"
+                                        href="{% url 'a4dashboard:plan-update' organisation_slug=object.organisation.slug slug=object.slug %}"
                                         class="dropdown-item"
                                         data-embed-target="external"
                                     >{% trans 'Edit' %}</a>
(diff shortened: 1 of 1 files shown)


meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -53,6 +53,34 @@ <h1 class="item-detail__title">
                         <p>{% trans 'Participation is planned, but has not started yet.' %}</p>
                     {% endif %}
                 </div>
+
+                <div class="item-detail__actions lr-bar">
+                    {% has_perm 'a4projects.change_project' request.user view.project as user_may_change %}
```
kleingeist: although this would check the correct rule it is inconsistent with the view [where we check the 'meinberlin_plans.add_plan' permission](https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/dashboard/views.py#L106) (which btw. is also not correct there, as we should check the change_plan perm on the update view).

i'd suggest to check the change_plan permission in the template, too. 
As always with our rules/permission system its a pain to find out which object is expected by permission. In this case we can determine from the code, that `change_plan` uses `is_superuser` and `is_initiator`. While `is_superuser` expects only a user object, `is_initiator` expects an organisation like object with a `has_initiator` method, or an object with an `organisation` property.
Thus i would suggest to always check `change_plan` (and `add_plan`) with an user and a organisation object.

kleingeist:
i get a 'NoReverseMatch' on the plan detail page

> NoReverseMatch at /vorhaben/letzter-tpoller-plan/
> Reverse for 'plan-update' with keyword arguments '{'slug': 'letzter-tpoller-plan', 'organisation_slug': ''}' not found. 1 pattern(s) tried: ['dashboard/organisations/(?P<organisation_slug>[-\\w_]+)/plans/(?P<slug>[-\\w_]+)/$']

kleingeist:
if the reverse match is fixed i would merge.
i'd refactor the permission issued in another issue, as it will also touch the views

kleingeist:
lgtm :)

PR #1039, opened 2017-11-21
Fix html_date template tag timezone handling

2e2a:
The `date` and `time` template filters expect the value in localtime.


kleingeist:
lgtm

PR #1041, opened 2017-11-21
Plans export

kleingeist:
currently there is no button in the ui to export. 
for testing navigate  to `/vorhaben/export/format/xslx/` 
exports are available for every user that may view the vorhaben list: which is currently everyone


2e2a:
works

Issue #1026, opened 2017-11-20
Extend email import to allow for <foo@bar.de> mails


kleingeist:
this would allow to upload an emails body (for example from rundmails) as email addresses are given as `"John Doe <foo@bar.de>"`

this would require to extend the regular expression used for email extraction

PR #1042, opened 2017-11-21
Extend email regex

2e2a:
Extract emails of the form `<foo@bar.org>`

Fixes https://github.com/liqd/a4-meinberlin/issues/1026

### DIFF
--- a/meinberlin/apps/users/fields.py
+++ b/meinberlin/apps/users/fields.py
@@ -37,7 +37,7 @@ class EmailFileField(forms.FileField):
     widget = widgets.FileInput
     # Find possible email strings. Emails may be quoted and separated by
     # whitespaces, commas or semicolons.
-    email_regex = re.compile(r'[^\s;,"\']+@[^\s;,"\']+\.[a-z]{2,}')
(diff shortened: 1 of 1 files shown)


meinberlin/apps/users/fields.py
```diff
@@ -37,7 +37,7 @@ class EmailFileField(forms.FileField):
     widget = widgets.FileInput
     # Find possible email strings. Emails may be quoted and separated by
     # whitespaces, commas or semicolons.
```
kleingeist: would adapt the comment, too

meinberlin/apps/users/fields.py
```diff
@@ -37,7 +37,7 @@ class EmailFileField(forms.FileField):
...
-    email_regex = re.compile(r'[^\s;,"\']+@[^\s;,"\']+\.[a-z]{2,}')
+    email_regex = re.compile(r'[^\s;,"\'<]+@[^\s;,"\']+\.[a-z]{2,}')
```
kleingeist: the regex should match the `>` after the add, too.
although i cant find a usefeul example where this would be a problem o.0

PR #1043, opened 2017-11-21
status markers

xi:
this change the plans map so that markers with different icon depending on status are used.

Note that the first commit is actually a fix for a regression that was introduced in #1034.

I am not totally sold on the icons I picked. However, I like the general idea.

I think we should change the logic of marker selection to show a selected marker *in addition* to the regular marker. Currently, we just change the marker icon and shift it to another pane.


kleingeist:
> I think we should change the logic of marker selection to show a selected marker in addition to the regular marker. Currently, we just change the marker icon and shift it to another pane.
i support that idea. should we try to add that to this PR or a separate one?

kleingeist:
we are missing a legend for the icons that are used.
either a classic one or a selectable like at bplans https://bplan-stage.liqd.net/berlin/#

PR #1046, opened 2017-11-22
small fixes for plans map

xi:


kleingeist:
nice. i like the find-as-you-type functionality. i hope it won't become a problem if there will be a lot of plans. do you have any idea, on how many plans we may expect and when searching might become slow?

xi:
>  i like the find-as-you-type functionality.

Honestly I was surprised by that. Usually change events are only fired when you hit return or focus out of the input. I guess this is some react magic. If this becomes an issue we could add some throttling.

PR #1047, opened 2017-11-22
Show the selected/active marker supplementary

kleingeist:
The original base marker is currently shown too
otherwise it's weird that the active marker is shown besides the cluster where the base cluster is counted in. for example with a cluster of two markers if one of it is selected it is not clear if there are 3 vorhaben (2 in the cluster, 1 selected) or two

It may be better to just hide the base marker though


xi:
Nice!

PR #1049, opened 2017-11-22
add intro text for plans

xi:


kleingeist:
i am not sure about the widths.
on large screens the filter bar looks which is aligned left in relation to the wrapper looks deplaced between the smaller intro text and the full screen map.
vorhaben-intro-filter-deplaced


not sure how to fix this. my first thought was to move the filters within the map like in the draft below. but i am not sure how this would work out in regard to responsiveness
plans-filter-within

kleingeist:
i'll address the filterbar position in another task. for now this is a great improvement :)

PR #1050, opened 2017-11-22
Fixup 5591ddc28

xi:
Explicitly setting height fixes cases where line-height is ignored (e.g.
select) but breaks cases where the padding or line-height is changed.

I am not sure which is the best solution. For now I went with setting
the height every time the padding is overwritten.


kleingeist:
although i am not sure if this is the best  solution either, it fixes the problem and is explicit. thus i think it's an improvement which will not bear high debts in the future

PR #1055, opened 2017-11-22
Fix grid for right aligned filters within control bars

### LINKED ISSUE #1052: Filters aligned to the right of the control bar overlap on small screens
filter-overlap

kleingeist:
The float is set for larger screen to position the filter to the right.
The grid does not work as expected if the element does not float.
As grid-position does not include grid-cell automatically it is required
to include it manually to reset the correct grid floating and margins.
fixes #1052

### DIFF
--- a/meinberlin/assets/scss/components/_filter_bar.scss
+++ b/meinberlin/assets/scss/components/_filter_bar.scss
@@ -25,10 +25,9 @@
         float: right;
 
         @media (max-width: $breakpoint) {
-            // grid-position does not include grid-cell automatically
-            @include grid-cell();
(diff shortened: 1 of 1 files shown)


meinberlin/assets/scss/components/_filter_bar.scss
```diff
@@ -25,7 +25,8 @@
         float: right;
 
         @media (max-width: $breakpoint) {
-            float: none;
+            // grid-position does not include grid-cell automatically
+            @include grid-cell();
```
xi: I would maybe add this below the next comment

Issue #1040, opened 2017-11-21
Consistent filter UI


xi:
#1034 introduces a new filter interface that is optimized for client-side filtering. However, it is not consistent with the existing filter interface.

For filters, we need to show the name of the filter, the available options, and the currently selected option. The options are typically displayed in a dropdown.

We currently have 3 different options how this can be done. I do not have a particular preference for any of these, so I will just list some pros and cons.

### Server-side filters

The simplest way to do a dropdown is a `<select>` element. However, filtering server-side means navigation, and triggering navigation on the select's change event is unexpected, especially for keyboard users. Adding a separate submit button is also not desired.

So instead of using a select element, we ended up using a button that opens a dropdown of links. Note that WAI-ARIA is used to add some semantics. However, the keyboard navigation is very restricted compared to select elements.

The button contains both the name of the filter as well as the currently selected option. Unfortunately, the selected option is often cut off due to width limitations.

![2017-11-21-120421_1440x900_scrot](https://user-images.githubusercontent.com/202576/33069303-67b40658-ceb4-11e7-9e92-4660a98a7b96.png)

### Client-side filters

When filtering is done on the client, no navigation is needed. This allows us to use select elements for filters. The only downside here is that the element always shows the selected option. So in order to show the name of the filter, we have to add it to every option.

![2017-11-21-120404_1440x900_scrot](https://user-images.githubusercontent.com/202576/33069302-668ccecc-ceb4-11e7-985b-794405b074f9.png)

### Advocate Europe

Advocate Europe found an elegant solution by displaying the name of the filter in a label outside of the dropdown element. However, the layout is very different there, so it is not clear how it would translate to meinberlin.

![2017-11-21-120451_1440x900_scrot](https://user-images.githubusercontent.com/202576/33069304-68ec460c-ceb4-11e7-99e6-0d945b4c3020.png)

kleingeist:
i'd suggest to drop support for client site filters from within control(filter)-bars as they are hard to style.
furthermore i would limit the allowed elements within control(filter)-bars to a very small set:
`btn`, `dropdown` and `input-group` which we may style using the child `>` selector

using `dropdowns` from the client side is not really supported yet, as they need a lot of html boiler code, but we could add react compont to encapsulate the functionality. it could accept a list of item value, a onClick callback and maybe an item styling callback (or an item label list)

fuzzylogic2000:
Yes, I think that's a good way to go.

xi:
[...]

PR #1064, opened 2017-11-27
Revert "Fix select height"

xi:
As part of #1034 I set a fixed height on input elements to fix the height of select elements, which seem to ignore the line-height.

This negatively affects buttons, checkboxes, and textareas. We could attempt to set an explcit height for each of them, too. However, that is hard to maintain.

Since we diceded to not use select elements for filters in #1040, I now propose to roll back the previous changes and remove the fixed height from input elements.


kleingeist:
works nice :)
although it feels like loosing against css 
maybe we can reconsider the baseinput styling in the future^TM

PR #1065, opened 2017-11-27
2017 11 ka add to component library

fuzzylogic2000:
A few more components: nearly managed to finish the D


meinberlin/apps/contrib/templates/meinberlin_contrib/component_library.html
```diff
@@ -304,6 +328,282 @@ <h2 id="cl-commenting">Commenting</h2>
...
+                </div>
+
+                <h2 id="cl-component-library">Component library</h2>
+
+                <div class="cl-example">
+                    <div class="cl-example">
+                        Every component in the component library has a border and some padding.
+                    </div>
+                    <div class="cl-example cl-example--grey">
+                        And some have the same background as the module pages.
```
xi: There seems to be some typo here.

fuzzylogic2000: Do you mean grey? Gray, grey, all gray (or grey) to me!

xi: Nevermind. I misread.

meinberlin/apps/contrib/templates/meinberlin_contrib/component_library.html
```diff
@@ -422,6 +434,68 @@ <h2 id="cl-complete">Complete</h2>
...
+                            <li class="dashboard-nav__page">
+                                <a href="#"
+                                    class="dashboard-nav__item dashboard-nav__item--interactive">
+                                    Interactive Dashboard nav item
+                                </a>
+                            </li>
+                            <li class="dashboard-nav__page">
+                                <a href="#"
+                                    class="dashboard-nav__item dashboard-nav__item--interactive">
+                                    These interactive items are shown in a list
```
kleingeist: maybe we should also add the error case with the red exclamation mark icon, as it is specifically layouted to be on the right side of the menu item

meinberlin/apps/contrib/templates/meinberlin_contrib/component_library.html
kleingeist: i think i would prefer to actually use the ".module-content" class as a wrapper if we wan't to show components in the modules context for some reason. thus if the module-content changes we will have a real representation even if the background won't be gr[ea]y anmyore
altough this is also kind of weird, as the '.module-content' does also add some layout (paddings :()

fuzzylogic2000: Yes, that was my first shot, but looked sooo strange because of the padding. But I think you're right. I will change that.

kleingeist:
nice :) hope we won't forget to finish after xi leaves

PR #1066, opened 2017-11-27
Overlap vorhaben filterbar with the map

kleingeist:
The first commit replaces the `<select>` input element with a dropdown and the second applies the control-bar__top-overlap styling.
But i am not longer really convinced the second is a good idea, as the filterbar overlaps with the zoom buttons and the first element of the list, too (see screenshot).

Maybe we should refrain from the overlap, but still keep the dropdown commit.

vorhaben-map-overlap


kleingeist:
closed in favor of #1067

PR #1072, opened 2017-11-28
redirect on registration

fuzzylogic2000:
The changes are from AE and add the next parameter to the confirmation link in the email.


xi:
Very nice!

kleingeist:
nice :) i really like the feature and we have been missing it forever. maybe we have to adapt the moderator/participant invitation,too to include a next param

and we should do the same in the product :)

PR #1075, opened 2017-11-29
Import emails from files only if ascii encoded

kleingeist:
As it is difficult to guess the correct encoding of a
file, email addresses are restricted to contain only ascii
letters.

This works for every encoding which is a superset of
ascii like utf-8 and latin-1. Non ascii chars are simply ignored.


xi:
Seems like a reasonable approach to me. We have to remember that non-ascii is stripped if we ever want to reuse this code for something other than email.

PR #1076, opened 2017-11-29
de translation adjustments

xi:
see also https://github.com/liqd/a4-product/pull/139


kleingeist:
merci :)
i'm curious if it will build now without errors

PR #1077, opened 2017-11-29
add map to plan detail page

xi:


kleingeist:
looks good, but i changed it to pass the polygon as a template variable instead of a model attribute

PR #1078, opened 2017-11-29
extend plans freetext filter

xi:
also filter for organisation name and category


kleingeist:
currently all of the words have to be found in at least one of the fields. so it is not possible to match 
"great idea" of the "liqd" organisation when entering "great liqd"
not sure which behvaioru would be expected

kleingeist:
nice. works as expected :)

PR #1079, opened 2017-11-29
add datalist to category field

xi:
Not completely sure about this implementation:

- I created a new widget that wraps `TextInput` and requires some knowledge of django internals
- Not completely sure when `get_category_options()` is executed.

Still I think this version is usable.


fuzzylogic2000:
But why do I get the list twice? Is one produced by my browser (firefox) and one by the widget?

screenshot from 2017-12-19 10-04-12

kleingeist:
args. now i merged before github showed the comment.
not sure why that happened. this is only a browser element. will make a separate issue out of it

PR #1086, opened 2017-12-04
sort list items by last modified date

xi:
Note that the sorting can currently not be done in sql due to https://github.com/liqd/adhocracy4/issues/218


kleingeist:
sql is not nice :( 
but this is a good workaround and total fine until we have to use pagination

PR #1095, opened 2017-12-05
refactor document styling

xi:
facilitates styling in a4-product


kleingeist:
lgtm. does not break anythin in meinberlin

PR #1098, opened 2017-12-06
set sass variables with !default

xi:


kleingeist:
the css ends up in a identical state. so we are safe :)

PR #1103, opened 2017-12-12
2017 12 ka show actor in adminlog

fuzzylogic2000:


kleingeist:
lgtm :) thanks for spotting the missed translations

PR #1110, opened 2017-12-15
2017 12 ka fix tab breaking

fuzzylogic2000:
This was a problem in product (https://github.com/liqd/a4-product/issues/204) and meinberlin. In meinBerlin it is in the projectlist, the container view and the dashboard. The mobile view is fixed in the project list and the container view, but in the dashboard the tabs still break into the next line because the text on the tabs is too long. We have to keep that in mind for tabs to follow.

### DIFF
--- a/meinberlin/assets/scss/components/_tabs.scss
+++ b/meinberlin/assets/scss/components/_tabs.scss
@@ -67,3 +67,11 @@
 .tablist__left {
     float: left;
 }
+
+@media (max-width: $breakpoint) {
+    .tablist__right {
+        margin-bottom: 1em;
+        float: none;
+        text-align: right;
+    }
+}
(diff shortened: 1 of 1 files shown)


kleingeist:
i am not 100% sure about this PR as i think it breaks to soon and having to lines one for the tabs and one for example the follow button looks weird while there is still some place left. 
we could introduce another smaller breakpoint but we would just shift the problem. there is currently an ongoing discussion about removing the result tab and to expand the cms functionality to add initiator generated tabs. 

@fuzzylogic2000 would it be possible to move the whole breakpoint logic for tabs to the product and keep meinberlin the way it is?

fuzzylogic2000:
Yes, I will move this to beteiligung.in, where we have an extra xs breakpoint!

PR #1115, opened 2017-12-19
Style the reply button

kleingeist:
Very shameful css :/

this is required due to https://github.com/liqd/adhocracy4/pull/216


kleingeist:
closed in favor of #1123

PR #1118, opened 2017-12-20
Basic wagtail events/calendar

kleingeist:
This adds basic calendar/event pages from tempelhof to meinberlin.
It does not include the dynamic map yet.
and the styling is wrong but i don't now how to fix it. i am missing a concept for general info pages like that :(

not sure if we should merge or not

Note, that the actual requirement from the customer was realized with offline events now


fuzzylogic2000:
Oh, I just changed the calendar styling (and behaviour) on tempelhof. What did we want to accomplish here? Also have a block or add the dynamic calendar somewhere else?

kleingeist:
this PR may be closed as we are using a separate wagtail instance for the leitlinien prozess now

PR #1119, opened 2017-12-20
Add reference number to plans and use it from urls

kleingeist:
This adds a reference number for plans which is constructed from the year the plan was created and the global plan id `{year}-{pk}`. To have such a number is a requirement from our customers.

This replaces routing by slugs. Whenever needed the plan is routed via the pk
To get a plan based on the reference number the RelatedManager was extended with a `get_by_reference_number` method. This helps to keep the logic for building and parsing reference numbers in one place.

We fear/expect to have more customer requests for reference numbers. Thus it would be useful to decide on a common scheme.

And i am not sure if this one is the best, as it is very implicit and complicated.
The alternative would be to use the `reference_number` as a real unique charfield that acts like a `AutoSlug` and is autofilled from the fields `{created.year}-{pk}`. This approach would be more explicit and would allow for simple use of django views as one could just set the `slug_field` and rely on the default `get_object` method

now after writing this text the second approach seems way better to me, but i would love to get some feedback from @slomo as long as it is still possible ;)


kleingeist:
closed in favor #1120

PR #1121, opened 2017-12-21
Move invite functionality to a different app

kleingeist:
This follows https://stackoverflow.com/a/26472482 and http://marcela-campo.blogspot.de/2015/01/django-migrations-moving-django-model.html although it does not rename the sequence table yet.

Note that this will probably be moved to core again.
Not sure if it makes sense to merge now


kleingeist:
closed, as the invite will move to core and we will do the migration manually on the server

PR #1122, opened 2017-12-21
Add project information sections

kleingeist:
This adds a new model which contains project information sections.
They are intended to be used as a accordion in a project with the requirement of extensive information.

This PR adds the basic functionality based on Django Formsets and the Docs Block Styling.
It could be improved by 
- styling the dashboard formset
- styling the frontend
- explain the difference between information and extended information
- allow for dynamic addition of formset forms with javascript
- allowing to set the sections weight or dynamically move formsets with javascript

Note that the `ProjectInformationComponent` should move to core. Thus it is not advisable to use berlin specific features like information section in that component.


kleingeist:
closed in favor of #1131

PR #1124, opened 2017-12-22
Split webpack config into prod and dev

slomo:


kleingeist:
lgtm. will add the same to the product, too to unify all a4 projects

PR #1131, opened 2018-01-02
Extend ckEditor to allow for collapsibleItems

kleingeist:
This is an alternative approach to #1122 which is extending the ck-editor to allow for collapsible items instead of adding the sections as separate model to the database.

IMHO this allows for an easier interface as the collapsibles may be defined at the same place as the other information. And it is more flexible as it allows to add the collapsible at any place within the text.

But there are also downsides:
- to prevent from allowing to many elements/attributes in user entered fields the ckeditor part is kept simple and contains only 3 divs with classes
   - still bleach had to be extended to allow for divs and the class attribute. it is not possible to restrict the class to a specific value. i am not sure if setting classes could be a security concern
 - for rendering the simple html has to be transformed to the more complex dom with correct collapsible definitions. the code for this is kind of complex:
   - the text has to be parsed and an element tree has to be built
   - all the collapsible elements have to be replaced by the correct html. which is involves a render_template call. (advantage: this allows for custom elements/classes as needed by bs3 vs bs4)
   - the original tree has to be modified and serializes again
- currently the block are not collapsible within the ckeditor as this would require a lot of additional markup to retain a11y
- django-ckeditor is broken and won't allow to add additional plugins from the global config, thus a new field type is required

I think it would be a good idea to focus on the editor itself instead of building around it and i would like to switch to another editor in the future.
For example https://quilljs.com/ on which some experiments had been already started by @slomo on a [opin dev branch](https://github.com/liqd/a4-opin/compare/ym-2017-12-richtext-editor-demo?expand=1)
or https://draftjs.org/ which will will [be used by wagtail](https://github.com/springload/draftail)

@MagdaN  @fuzzylogic2000 @mkind   i would be happy to get some feedback on this approach


meinberlin/apps/ckeditor_collapsible/templatetags/ckeditor_collapsible_tags.py
```diff
@@ -0,0 +1,41 @@
...
+            title.tag = 'span'
+            del title.attrib['class']
+
+            body.tag = 'div'
+            del body.attrib['class']
+
+            final_html = render_to_string(
+                'ckeditor_collapsible/collapsible_fragment.html',
+                dict(
+                    id='ck-collapsible-{}'.format(i),
```
kleingeist: temporary: a unique id should be generated but i was too lacy to do it right for this element

MagdaN: is it possible, that the index is not unique? This would happen, if the RichTextCollapsibleField or the RichTextCollapsibleUploadingField is used more then once on the same page, right? How would you make it unique?

mkind:
> i am not sure if setting classes could be a security concern

Well yes. The [django documentation](https://docs.djangoproject.com/en/2.0/topics/security/#cross-site-scripting-xss-protection) has a nice example.

MagdaN:
I would assume that bleach removes 'invalid' classes

kleingeist:
@mkind we already have user entered input and rely on bleach/html5lib to sanitize the input. they promise to strip invalid tags and attributes such as `class=foo  onmouseover=javascript:func()`
the only thing we have to allow for additional is a valid class name. but i can't imagine an attack based on setting a valid classname for a div

kleingeist:
i'll close this PR and will create a new one in the a4-core

PR #1142, opened 2018-01-11
Mn 2018 01 use export from a4

MagdaN:
only works when this PR in a4 is merged: https://github.com/liqd/adhocracy4/pull/233


meinberlin/apps/exports/views.py
```diff
@@ -1,19 +1,21 @@
+# a4 classes need to be imported here as # noqa isort:skip
+# otherwise the export register would not work # noqa isort:skip
+
+from adhocracy4.exports.mixins import ItemExportWithCommentCountMixin  # noqa isort:skip
+from adhocracy4.exports.mixins import ItemExportWithCommentsMixin  # noqa isort:skip
+from adhocracy4.exports.mixins import ItemExportWithLocationMixin  # noqa isort:skip
+from adhocracy4.exports.mixins import ItemExportWithRatesMixin  # noqa isort:skip
+from adhocracy4.exports.views import ItemExportView  # noqa isort:skip
```
kleingeist: this was required as the plans/views.py did still import the Mixins from this package instead of the a4 core package. i fix it with a commit.

kleingeist:
lgtm. thanks for the work :)

PR #1150, opened 2018-01-17
Polygon import and export

kleingeist:
work in progress

missing
- [x] error handling
- [x] shapefile upload
- [x] ui/ux concept (e.g. popups to explain the upload)
- [ ] clean webpack config
- [x] use modalsomething befor uploading
- [x] show errors in modal


webpack.common.js
```diff
@@ -37,16 +37,19 @@ module.exports = {
...
-      'leaflet-draw/dist/leaflet.draw.css',
-      './meinberlin/assets/js/i18n-leaflet-draw.js'
-    ],
     embed: [
       './meinberlin/assets/js/embed.js'
     ],
     'popup-close': [
       './meinberlin/assets/js/popup-close.js'
+    ],
+    'map_choose_polygon_with_preset': [
```
kleingeist: should we keep this filename for the bundle?

meinberlin/apps/maps/assets/map_choose_polygon_with_preset.js
```diff
@@ -0,0 +1,247 @@
+/* global django */
+
+const $ = require('jquery')
+// const L = require('leaflet')
```
kleingeist: this is an optimisation to only load Leaflet once. but i am not sure if i like it. 
what is your opinion? packaging leaflet to the map_with_preset bundle or using leaflet from the `window.L` scope while ensuring from the python side that is is loaded (`maps/widgets.py`)

meinberlin/apps/maps/assets/map_choose_polygon_with_preset.js
```diff
@@ -0,0 +1,247 @@
...
+    $input.val(JSON.stringify(group.toGeoJSON()))
+    $input.trigger('change')
+  }
+}
+
+(function (init) {
+  $(init)
+  $(document).on('a4.embed.ready', init)
+})(function () {
+  // Prevent from including leaflet in this bundle
```
kleingeist: see above

meinberlin/apps/maps/assets/map_choose_polygon_with_preset.js
```diff
@@ -0,0 +1,247 @@
...
+      const exportLink = L.DomUtil.create('a', '', container)
+      const exportIcon = L.DomUtil.create('i', 'fa fa-download', exportLink)
+      exportLink.setAttribute('title', django.gettext('Export as GeoJSON'))
+      exportIcon.setAttribute('aria-label', django.gettext('Export as GeoJSON'))
+
+      if (this.options.onExportLinkClick) {
+        exportLink.onclick = this.options.onExportLinkClick
+      }
+
+      // FIXME: is this accessible doe we need a label or something?
```
kleingeist: is this file input accessible or do we need a label or something?

meinberlin/apps/maps/assets/map_choose_polygon_with_preset.js
```diff
@@ -0,0 +1,247 @@
...
+      const importInput = L.DomUtil.create('input', 'sr-only', container)
+      importInput.setAttribute('type', 'file')
+
+      const importLink = L.DomUtil.create('a', '', container)
+      const importIcon = L.DomUtil.create('i', 'fa fa-upload', importLink)
+      importLink.setAttribute('title', django.gettext('Import shapefile or GeoJSON file'))
+      importIcon.setAttribute('aria-label', django.gettext('Import shapefile or GeoJSON file'))
+
+      importLink.onclick = function (e) {
+        // FIXME: is preventing the default event required?
```
kleingeist: not sure if/when this is required :|

kleingeist:
unfortunately this does currently not work with IE 10 and i am not sure why >:(
the example at http://leaflet.calvinmetcalf.com/#0/-81/103 doesn't work either, but there is an issue which states someone made it work with ie ( https://github.com/calvinmetcalf/shapefile-js/issues/53 )
we probably have to look into that again :(

kleingeist:
after talking with the pm it seems we should use a popupsomething prior to the actual import where we may explain the allowed formats

kleingeist:
not sure about the separation of the `onImport` and `onExport` functions from the controls. 
The button title within the control makes only sense if we now the actual onExport/Import functions. This will get even worse if we add a modal. Not sure how to design it different though, as the onExport has to access variables from the surrounding scope. So maybe it doesn't make sense to try to separate the control. Maybe it would be better to just add a `createControlHtml` function (equivalent to the `onAdd`) to create the control elements

PR #1152, opened 2018-01-17
Mn 2018 01 add filters to map

MagdaN:
still missing: custom marker for the address search

to get this running you need to run 
`python manage.py import_geodata` (GDAL has to be installed)


MagdaN:
@rittermo we would need another marker that is set when jumping to an address, i would suggest to use the same one as the default one but switch colors (blue background, white circle) -> the default marker can be found here: https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/assets/images/map_pin_01_2x.png

meinberlin/apps/plans/assets/plans_map.jsx
```diff
@@ -122,6 +135,79 @@ class PlansMap extends React.Component {
...
 
+  onDistrictFilterChange (event) {
+    this.setState({
+      filters: update(this.state.filters, {
+        $merge: {district: parseInt(event.currentTarget.value, 10)}
+      })
+    })
+  }
+
+  onAddressFilterChange (event) {
```
fuzzylogic2000: This should be onAddressSearchChange

PR #1159, opened 2018-01-22
make districts required

MagdaN:


meinberlin/apps/plans/migrations/0006_make_plan_district_required.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def delete_plans(apps, schema_editor):
+    plan_model = apps.get_model('meinberlin_plans', 'Plan')
+    for plan in plan_model.objects.all():
+        plan.delete()
```
kleingeist: delete may also be called on a queryset like `plan_model.objects.all().delete()` but i would not update this PR. just catched my eye

MagdaN: Ah, ok, good to know, thanks!

PR #1164, opened 2018-01-22
don't add berlin polygon to the list of district polygons

MagdaN:
On the development server there is an extra polygon for the border of berlin in mappresets, which does not get loaded when running 'import_geodata'. Because of this, on the plans map not only the district borders but also the city are displayed and therefore added twice which looks strange. 

To remove the district borders on the map I excluded the Berlin map preset from the query.


kleingeist:
lgtm :)

PR #1169, opened 2018-01-23
adding a first CSP version

mkind:
* a first version of the content security policy
* basis is the discussion in https://github.com/liqd/adhocracy4/issues/86 especially comment https://github.com/liqd/adhocracy4/issues/86#issuecomment-342775744
* due to [some issues of wagtail](https://github.com/wagtail/wagtail/issues/1288) I needed to exclude the `/admin` url


kleingeist:
csp makes problems on local machines,
as the django debug toolbar uses an external jquery ressource and
the local `webpack.dev.js` config generates `eval` based code.

my first thought was to enable csp only via the `config/settings/production.py` settings. but i am not sure if this would be a good idea as we'd miss problems. we could set it to report only by default (for dev) and set `CSP_REPORT_ONLY=False` from the `production.py`

kleingeist:
thank for these improvements :) they were long overdue

PR #1170, opened 2018-01-23
introducing some http security headers

mkind:
introducing the following headers

* `HttpOnly` to disable cookie access via js
* `x-xss-protection: 1; mode=block` to enable client side xss blocking
* ~HSTS to instruct the browser to use a secure connection. for the beginning, we start with a low age. however, it should be increased continuously~ already implemented via nginx
* `x-content-type-options: nosniff` to prevent the browser from guessing any mime types


kleingeist:
i thought we'd already using hsts via the nginx gateway
@treba123 do you know which hsts headers are sent on which domain?

mkind:
@treba123 @kleingeist 

ah sure, I see

`Strict-Transport-Security: max-age=15552000; includeSubdomains; preload`

treba123:
This should fix our C rating on the mozilla observatory https://observatory.mozilla.org/analyze/mein.berlin.de

PR #1176, opened 2018-01-29
Ka 2018 01 use distinct ordering filter

fuzzylogic2000:
Uses the distinct ordering filters from a4.

I searched for ListViews with pagination not using the ordering filter. I found:
- Actions -- use timestamp for ordering
- LogEntry -- also timestamp
- Dashboard2 -- uses a filterset, but currently without ordering filter. I guess, we will remember for a wile that we have to be careful if we add one, but should we put that down somewhere as well? or hope, that django filters fixes it before it can go wrong? To order, it uses created
- ProjectContainers -- uses pagination, but I couldn't find an ordering. But this is shown in the normal project list view and ordered?! I can only find how the projects from the containers are filtered out of the project list view, not how the containers are put into the list.

### DIFF
--- a/meinberlin/apps/contrib/filters.py
+++ b/meinberlin/apps/contrib/filters.py
@@ -1,6 +1,6 @@
-import django_filters
 from django.utils.translation import ugettext_lazy as _
 
+from adhocracy4.filters.filters import DistinctOrderingFilter
 from adhocracy4.filters.widgets import DropdownLinkWidget
 
 from . import mixins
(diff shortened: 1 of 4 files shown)


kleingeist:
@fuzzylogic2000 
so we should give them ListViews an distinct ordering, too? 
should we add them to this PR?

fuzzylogic2000:
@kleingeist Yes, I think so. But I hoped timetsamp and created would be distinct enough?! Or we just add the pk to every ordering to make sure nothing can ever go wrong.
And what about the ProjectContainers? Do they get the ordering somewhere?

kleingeist:
yes i think it is okay to regard timestamps with microsends as distinct. 
what do you mean by project containers? you mean the projects within the container? they should get the default ordering from `Project` which is `created`

fuzzylogic2000:
As the models used in the views then all have a distinct ordering, I think we're fine?!

kleingeist:
did try on a few examples but i don't have enough ideas created for testing. couldn't find a problem. will merge and see what happens on the server :)

PR #1182, opened 2018-01-31
Greenkeeper/react flip move 3.0.1

kleingeist:
Not sure why, but it flip-move does not work with `require` anymore. Using `import` is fine though and works with babel.


kleingeist:
we could alternatively use `require('react-flip-move').default`. which would u prefer

MagdaN:
@kleingeist I prefer the require and changed it

PR #1189, opened 2018-02-02
use reference number for all items (prio processes as well) and show in detail view

fuzzylogic2000:


fuzzylogic2000:
@kleingeist could you check that and merge in case all else is fine?

kleingeist:
lgtm.

PR #1194, opened 2018-02-07
Make (map)-ideas and topicprio moderateble (Ampel)

kleingeist:
This PR adds the moderatable variant (ampel) to the following apps (as defined in the US)
* ideas  (Brainstorming, Idea Collection, AgendaSettings)
* mapideas  (Spatial Brainstorming, Spatial Idea Collection)
* topicprio  (Topic Prioritization)

Note that **maptopicprio (Place Prioritization) is not moderateble yet**. As bürgerhaushalt and kiezkasse  were moderateble before it is the only app/process left
@CarolingerSeilchenspringer should the maptopicprio have the option to give moderator feedback, to?


kleingeist:
I could move the Moderateble to the `AbstractIdea` model as every model inheriting from it is using the moderateable, too. But  i am not sure if this is a good idea, as this limits our possibilities if we have to introduce a new itemlike app

what do you think?

fuzzylogic2000:
I think it is a good idea to move it to the AbstractIdea model. I think in most cases the new app would also need it anyway. And if not, we can either split it up again or make a new model.

And I think, we can also make the maptopicprio moderatable right away? It doesn't make a lot of sense to use it everywhere but there.

meinberlin/apps/contrib/templatetags/item_tags.py
```diff
@@ -44,14 +45,19 @@ def get_item_delete_url(item):
...
-    return reverse(url_name, kwargs={
-        'year': item.created.year,
-        'pk': '{:05d}'.format(item.pk)
-    })
+    try:
+        return reverse(url_name, kwargs={
+            'year': item.created.year,
+            'pk': '{:05d}'.format(item.pk)
+        })
+    except NoReverseMatch:
```
fuzzylogic2000: Nice!

kleingeist:
i think the migration is missing in the commit

CarolingerSeilchenspringer:
we don't need the Ampel for Topic- & Placeprio! BUT: For Story #165-[mB] internal categories for moderation we need moderation option for ALL ideas and comments.

kleingeist:
ups. i will remove the ampel from (map)topicprio again. i have no idea how i came up with the idea

PR #1199, opened 2018-02-12
Ka 2018 02 translations

fuzzylogic2000:


kleingeist:
:) thanks for the translations. this time i would have totally forgottem them

PR #1204, opened 2018-02-14
smaller font-size for item detail meta

fuzzylogic2000:
Meta gets has same font-size as label, as discussed with @CarolingerSeilchenspringer and Sandy.


kleingeist:
for some reasons the timeline is using ` .item-detail__meta` too 0.o
imho the timeline module description looks weird with the small text

timeline-small

kleingeist:
lgtm :)

PR #1205, opened 2018-02-14
Enforce german l18n

kleingeist:
As meinBerlin is directed to a german speaking audiency and the user generated contents (for example project descriptions etc.) are not translateable the default language will be changes to german and the users browser settings ignored.
This change has been approved by the pm and the po


kleingeist:
ich war jetzt kurz davor ein schlaaaaaand gif einzubinden. aber nenenene. das geht dann doch nicht ;)

PR #1230, opened 2018-02-28
Ka 2018 02 change pins

fuzzylogic2000:


meinberlin/apps/mapideas/forms.py
```diff
@@ -16,11 +17,17 @@ def __init__(self, *args, **kwargs):
...
+        self.fields['category'].widget = widgets.SelectCategoryWithIconWidget(
+            choices=self.fields['category'].choices,
+            attrs={'qs': self.fields['category'].queryset})
 
     class Meta:
         model = models.MapIdea
         fields = ['name', 'description', 'category', 'point', 'point_label']
 
+    class Media:
+        js = ('js/category_formset.js',)
```
kleingeist: we don't need the `category_formset.js` the `select_dropdown.js` is already required in `adhocracy4/index.js`
but we need a small javascript to initialize the select onLoad (similiar but simpler then in `category_formset.js`

meinberlin/assets/scss/components/_select_dropdown.scss
```diff
@@ -18,9 +18,11 @@
     margin-right: 0.6em;
 }
 
-.select-dropdown__item:first-child {
-    padding-top: 0.8em;
-    padding-bottom: 0.8em;
-    color: $text-color-gray;
-    background-color: $bg-secondary;
+.select-icon {
```
kleingeist: not sure why this is necessary. but i probably broke it while renaming the css classes in core

meinberlin/apps/plans/assets/plans_map.jsx
```diff
@@ -439,7 +443,7 @@ class PlansMap extends React.Component {
                 <li>
                   <button
                     type="button"
-                    className="dropdown-item select-item"
+                    className="dropdown-item"
```
kleingeist: note to myself: i should check the classes in the form widgets, category filters and this jsx filter

kleingeist:
yipie :)

PR #1241, opened 2018-03-14
Add export button to vorhaben list 

kleingeist:
also adds the district and a link to the project to the export

### DIFF
--- a/.polylintrc
+++ b/.polylintrc
@@ -1,5 +1,5 @@
 py:flake8 --exclude */migrations/*,*/settings/* %s
-py:isort -rc -c %s
+py:isort --diff -rc -c %s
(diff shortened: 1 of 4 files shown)


kleingeist:
what kind of styling? its a standard button

MagdaN:
its a bit unconnected to the rest of the content
bildschirmfoto 2018-03-14 um 16 31 57

kleingeist:
ah i see what you mean. 
i oriented the layout at the concept attached to:
https://tree.taiga.io/project/liqd-meinberlin/us/10

maybe we should get back to design/pm . the problem with the big whitespaces will occur with the new layout on large screens in general :(

PR #1253, opened 2018-03-21
Use core offlineevents

kleingeist:
depends on https://github.com/liqd/adhocracy4/pull/248

note that uploading and viewing documents is not supported in mB yet (missing concept/layout...)


MagdaN:
closing as we have not moved offline events to a4

PR #1254, opened 2018-03-21
Ka 2018 03 add image fields to idea models

fuzzylogic2000:


meinberlin/apps/ideas/forms.py
```diff
@@ -1,15 +1,32 @@
...
 
 from . import models
 
 
 class IdeaForm(CategorizableFieldMixin, forms.ModelForm):
+    right_of_use = forms.BooleanField(required=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.instance.name != '' and self.instance.image:
```
MagdaN: I think you can check for `self.instance.pk` here, as the instance won't have a pk when it has not been saved

meinberlin/apps/mapideas/forms.py
```diff
@@ -18,10 +17,26 @@ def __init__(self, *args, **kwargs):
             polygon=self.settings.polygon)
         self.fields['point'].error_messages['required'] = _(
             'Please locate your proposal on the map.')
+        if self.instance.name != '' and self.instance.image:
```
MagdaN: the same here

meinberlin/apps/maptopicprio/forms.py
```diff
@@ -20,10 +21,23 @@ def __init__(self, *args, **kwargs):
             polygon=self.settings.polygon)
         self.fields['point'].error_messages['required'] = _(
             'Please locate your proposal on the map.')
+        if self.instance.name != '' and self.instance.image:
```
MagdaN: and here

meinberlin/apps/topicprio/forms.py
```diff
@@ -10,7 +11,23 @@ class TopicForm(CategorizableFieldMixin, forms.ModelForm):
 
     description = fields.RichTextUploadingFormField(
         config_name='image-editor', required=True)
+    right_of_use = forms.BooleanField(required=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.instance.name != '' and self.instance.image:
```
MagdaN: and here

meinberlin/apps/ideas/forms.py
fuzzylogic2000: Yes, thanks! But even just checking for the image should be enough!

PR #1257, opened 2018-03-22
Enable HTTP2-push for static files

treba123:
This enables the http2 feature of pushing files before they get
requested by the browser. It can speed up browser rendering,
especially on on cold caches/first page landings.

Important to note: this does NOT disable browser caching and
browsers do cancel unnessecary file transmissions. But in cases
where the push is faster than the cache (which does happen quite
often, even on SSDs but more often on HDDs), some browsers
rather redownload the file. The network overhead should be
negitable, as media images make up for most of the traffic and
the pushes use streams instead of new connections (HTTP2, yeah!)

### DIFF
--- a/meinberlin/config/settings/base.py
+++ b/meinberlin/config/settings/base.py
@@ -112,6 +112,7 @@ INSTALLED_APPS = (
 )
 
 MIDDLEWARE = (
+    'django_cloudflare_push.middleware.push_middleware',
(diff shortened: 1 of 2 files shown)


kleingeist:
yeah. push it real good!

https://www.youtube.com/watch?v=vCadcBR95oU

PR #1260, opened 2018-03-23
remove scrollTo from code as it breaks in IE

MagdaN:
the scrollTo breaks the code completely in IE 11 (map disappears) and  as I could not really figure out why and when this is actually needed, I removed it.


kleingeist:
scrollTo is used to scroll the plans list to the correct entry when the user clicks on the pin in the map. kind of a core feature of the current plans layout

PR #1275, opened 2018-04-04
add label model and label to idea model

fuzzylogic2000:


fuzzylogic2000:
did it all in a4 now!

PR #1293, opened 2018-04-16
Mn 2018 04 add plans creation to dashboard

MagdaN:
depends on https://github.com/liqd/adhocracy4/pull/256


meinberlin/apps/plans/forms.py
```diff
@@ -36,3 +38,31 @@ class Meta:
...
+class CustomMultipleChoiceField(forms.ModelMultipleChoiceField):
+
+    def clean(self, value):
+        value = [value]
+        return super().clean(value)
+
+
+class ProjectPlansDashboardForm(ProjectDashboardForm):
+    plans = CustomMultipleChoiceField(
+        widget=forms.RadioSelect,
```
ihucos: Maybe we can have setting the widget inside the CustomMultipleChoiceField

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -59,11 +59,15 @@ <h1 class="item-detail__title">
                         <dd class="u-whitespace-pre">{{ object.contact }}</dd>
                     </dl>
 
-                    {% if object.project and not object.project.is_draft and object.project.is_public %}
-                        <p><a href="{{ object.project.get_absolute_url }}">{% trans 'Go to participation' %}</a></p>
-                    {% elif object.participation == 1 %}
-                        <p>{% trans 'Participation is planned, but has not started yet.' %}</p>
-                    {% endif %}
+                    {% for project in object.projects.all %}
+                        {% if not project.is_draft and project.is_public %}
```
ihucos: possibly too much logic in the template? I think it's still fine

meinberlin/test/factories/plans.py
```diff
@@ -25,3 +24,12 @@ class Meta:
...
     participation = plan_models.PARTICIPATION_UNDECIDED
+
+    @factory.post_generation
+    def projects(self, create, extracted, **kwargs):
+        if not create:
+            return
+
+        if extracted:
+            for project in extracted:
+                self.projects.add(project)
```
ihucos: I am not very familiar with pytest yet
self.projects is a django relation, right? then we could do:
`
self.projects.add(*extracted)
`
See https://docs.djangoproject.com/en/2.0/ref/models/relations/#django.db.models.fields.related.RelatedManager.add

If extracted is a an empty something the `if extracted` could also be removed

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -59,11 +59,15 @@ <h1 class="item-detail__title">
...
+                            <p><a href="{{ project.get_absolute_url }}">{% trans 'Go to participation in ' %}{{ project.name }}</a></p>
+                        {% endif %}
+                    {% empty %}
+                        {% if object.participation == 1 %}
```
ihucos: magic number, should be constant.
use something like:
`
if object.participation = object.PARTICIPATION_XXXXX
`

meinberlin/apps/plans/forms.py
fuzzylogic2000: Yes, just change it!

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
fuzzylogic2000: Yes! But I think, but we kind of wait here for a design and a concept as well.

meinberlin/test/factories/plans.py
fuzzylogic2000: I don't know. Does it work?

meinberlin/apps/plans/forms.py
```diff
@@ -58,10 +58,8 @@ class Meta:
         required_for_project_publish = ['plans']
 
     def save(self, commit=False):
-        self.instance.plans.clear()
-        for plan in self.data['plans']:
-            self.instance.plans.add(plan)
-        self.instance.save()
+        plans = self.data['plans']
```
fuzzylogic2000: Aha! If you review my stuff, I like when you tell me how to do it properly and then I have to do it myself and learn how to, so that you don't have to tell me again next time. :) But I guess, not everyone likes that.

ihucos: Doch, we all like it. Than as we discussed we do pull request the "propper" way and the reviewer just requests changes.

ihucos:
@MagdaN die Aenderungen sind drin. Leider bleibt der button da, weil er von der parent template kommt, und wir es zumindestens nicht in unserer custom componenten template ueberschreiben koennen.

fuzzylogic2000:
Do you mean the save-button in the dashbord? I think, that's fine for now. We should also link the titles of the vorhaben to the vorhaben in case someone wants to check? Or would that be confusing? I will make an issue for it and merge though!

PR #1308, opened 2018-04-26
Ka 2018 04 face to face

fuzzylogic2000:


fuzzylogic2000:
How is the plugin.js loaded and the right form field shown for the collapsible thingy?

PR #1313, opened 2018-04-30
More fine grained call to action in notifications, closes #1250

### LINKED ISSUE #1250: Notification links to project instead of idea for ideas
https://github.com/liqd/a4-meinberlin/blob/master/meinberlin/apps/notifications/templates/meinberlin_notifications/emails/notify_moderator.de.email#L11 tries to get the correct link, but somehow fails.
not sure if this is true for new ideas or only new comments for ideas

ihucos:
I am actually not so happy with this solution.
We use similar code in other templates: https://github.com/liqd/a4-meinberlin/tree/master/meinberlin/apps/notifications/templates/meinberlin_notifications/emails

And it's too much logic in the template. But somehow I did'nt came up with a better solution for this one.


ihucos:
(Duh, I forgot to add `ihc-` in the branch of my last two prs)

PR #1317, opened 2018-05-02
Pm 2018 05 style activties detail

### LINKED ISSUE #1318: Accordion styling looks off
left padding to the content text not matching title text.

phillimorland:
fixes #1318

### DIFF
--- a/meinberlin/assets/scss/components/_accordion.scss
+++ b/meinberlin/assets/scss/components/_accordion.scss
@@ -38,7 +38,7 @@
 }
 
 .accordion__body {
-    padding: $padding / 2;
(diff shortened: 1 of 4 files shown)


meinberlin/assets/scss/components/_accordion.scss
```diff
@@ -38,7 +38,7 @@
 }
 
 .accordion__body {
-    padding: $padding / 2;
+    padding: $padding/2 1em;
```
fuzzylogic2000: Yeah! That looks much nicer! You could use $padding here instead of 1em, as that is also used in the first bit.

PR #1321, opened 2018-05-03
Mn 2018 05 update plans list

MagdaN:
the retrieval of the status string in the view feels a bit clumsy, maybe there is a more elegant way to do that?

### DIFF
--- a/meinberlin/apps/plans/assets/plans_map.jsx
+++ b/meinberlin/apps/plans/assets/plans_map.jsx
@@ -364,10 +364,7 @@ class PlansMap extends React.Component {
     if (i === this.state.selected) {
       itemClass += ' selected'
     }
-    let statusClass = 'list-item__status--active'
-    if (item.participation_string === item.participation_display) {
-      statusClass = 'list-item__status--inactive'
-    }
+    let statusClass = (item.participation_active === true) ? 'list-item__status--active' : 'list-item__status--inactive'
     return (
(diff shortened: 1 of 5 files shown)


meinberlin/apps/plans/views.py
```diff
@@ -47,6 +47,36 @@ def get_districts(self):
         except ObjectDoesNotExist:
             return []
 
+    def _get_status_string(self, projects):
```
fuzzylogic2000: like that:
`if project:
    find status_string and status_class
else:
    if item.participation:
        status_string = planned
     else:
        something else`

meinberlin/apps/plans/assets/plans_map.jsx
```diff
@@ -360,24 +360,26 @@ class PlansMap extends React.Component {
...
 
   renderListItem (item, i) {
-    let className = 'list-item list-item--squashed'
+    let itemClass = 'list-item list-item--squashed'
     if (i === this.state.selected) {
-      className += ' selected'
+      itemClass += ' selected'
+    }
+    let statusClass = 'list-item__status--active'
+    if (item.participation_string === item.participation_display) {
```
fuzzylogic2000: I find all this logic a bit distributed and well hidden.
Can we use the view to get a status string and the statusClass?

MagdaN:
@fuzzylogic2000 I removed the list comprehension and added an additional field to check which class to set in react

PR #1323, opened 2018-05-03
Ihc 2018 04 new plan detail

ihucos:


ihucos:
Mockups here: https://tree.taiga.io/project/liqd-liquid-software/us/190?milestone=182434
Let's say it like this: overall I learned a lot about how the frontend at meinberlin and can do it even better in the next issue.

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -24,57 +24,70 @@
...
+                            <dt>{% trans 'Participation' %}</dt>
+                            <dd>
+                                    {{ object.get_participation_display }}{% if object.participation == object.PARTICIPATION_YES and object.published_projects.count %},
+                                    <a href='#participation-plans'>{% trans 'see participation plans' %}</a>
+                                {% endif%}
+                            </dd>
+                        </dl>
+
+                        <dl>
+                            <dt>{% trans 'Planed cost' %}</dt><dd>{{ object.cost }}€</dd>
```
MagdaN: I think, cost is not required so  I guess it would be better to check if it is set, otherwise there would be None€

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -24,57 +24,70 @@
...
-                    {% if object.point_label %}
-                        <span class="label"><i class="fa fa-map-marker" aria-hidden="true"></i> {{ object.point_label }}</span>
-                    {% endif %}
                 </div>
 
-                <div class="item-detail__content">
-                    {% if object.description %}
-                        {{ object.description|richtext }}
-                    {% endif %}
+                <div class="container">
```
MagdaN: is the container class needed here?

PR #1336, opened 2018-05-08
Ka 2018 05 change face to face module detail view

fuzzylogic2000:
- [x] https://github.com/liqd/a4-product/issues/354

- [x] https://github.com/liqd/a4-product/issues/355


fuzzylogic2000:
@MagdaN Did you mean quite like in my last commit? This is not yet working for projects with only one face-2-face phase that hasn't started yet, because the project view doesn't know what module (or item-type) is used.

MagdaN:
@fuzzylogic2000 yes, that is exactly how I meant it, thanks, not so good that it does not work

PR #1337, opened 2018-05-09
By default the user is not registered for the newsletter

ihucos:
Tested it by creating a user, in it's setting page the newsletter is then not enabled.


ihucos:
@MagdaN check it out

PR #1338, opened 2018-05-09
face to face tests

ihucos:
thaaaaat was somehow more difficult than it looks. there is actually still some face to face code not covered but since i heard that this will be replaced or so I think it's fine.

I learned a lot about phases and modules and so on.


meinberlin/apps/activities/views.py
```diff
@@ -25,22 +24,13 @@ def form_valid(self, form):
...
 
     def get_object(self, queryset=None):
-        try:
-            first_activity = models.Activity.objects \
-                .filter(module=self.module) \
-                .first()
-            return first_activity
-        except AttributeError:
-            return None
+        return models.Activity.objects.filter(module=self.module).first()
```
fuzzylogic2000: Why is this still here? I thought you said it was needed after all?

ihucos: What was not needed is the try except clause, since there is no occasion where an AttributeError is raised. As we tested together .first() already returns None when there are no entries, so it's not needed to handle this case separately.... hmmm so from the behaviour the code that is there now should do exacly the same, I only removed the try except clause and the result is directly returned.

ihucos:
> What of the code will be replaced?
I don't get the question. In the merge I am taking your version where the merge conflicts are (and a manual adaptation)

fuzzylogic2000:
The code will not be replaced. We will delete it from the blueprints, because we have to work on the timeline before it can and should be (properly) used.

What is still not tested? Do you think it should be or is it fine like that? We could merge these tests first anyway and add the other tests soon?!

meinberlin/apps/activities/views.py
fuzzylogic2000: Ah alright, than I misunderstood you.

PR #1344, opened 2018-05-14
Added ellipsis for maps pop up

ihucos:
**Before**
screenshot from 2018-05-14 18-35-15

**After**
screenshot from 2018-05-14 18-34-58

### DIFF
--- a/meinberlin/assets/scss/components/_map.scss
+++ b/meinberlin/assets/scss/components/_map.scss
@@ -24,6 +24,8 @@
     font-size: $font-size-lg;
 
     line-height: normal;
+    text-overflow: ellipsis;
+    overflow: hidden;
 
     a {
         color: $link-color;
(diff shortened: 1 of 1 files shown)


ihucos:
A little out of process..... but nice

fuzzylogic2000:
Yeah! I was so sure this didn't work for two lines, but it works differently than I thought. Still nice. But as we do not cut off long texts, using word-break might be a better idea? I just commited that to your branch. Revert if you find your option better.
It would actually be awesome if the text would break like in my commit, but then be ellipsed after two lines. I couldn't find an easy solution for that, though. So let's hope for sensible users with short idea titles (with long words if they like). :)

ihucos:
Yeah, it's also good, I think both are much better then it was before. I don't know if hyphenation (silbentrennung) is an issue at all.

ihucos:
screenshot from 2018-05-15 09-44-49

PR #1348, opened 2018-05-15
Export emails

ihucos:
the format is one email per line, not xlsx
with xlsxwriter we would need to buffer the entire thing before the download...
lets see if we need that at all

oh no my yapf did changes that where not needed (gardening *aham*)


ihucos:
Incomplete because waiting if this is needed and possibly tests needed, xlsx format is also a think

ihucos:
Closing as there is another solution

PR #1386, opened 2018-05-22
Ihc 2018 05 registration tos link

ihucos:


ihucos:
Don't forget to check my german to english translation

fuzzylogic2000:
@ihucos Your translations would be so much easier to check if you commited the result of makemessages first!

PR #1394, opened 2018-05-24
Attempt tasks only one time

ihucos:


ihucos:
rrrghhh, I can't test it

fuzzylogic2000:
Ok, then we should merge, right?!

PR #1414, opened 2018-05-31
Using send mail to admin mixin

ihucos:


ihucos:
Don't merge yet: first I need to update requirements.txt and packages.json after https://github.com/liqd/a4-meinberlin/pull/1414 is merged.

PR #1415, opened 2018-05-31
One task per mail

ihucos:


ihucos:
Don't merge yet :-)

ihucos:
Still needs a little bit of work and cleanup, but the future of newsletter mail sending is here :-)

You can see on the newsletter object how many mails where send and how many had errors
screenshot from 2018-05-31 18-00-07

with a simple click on the mail status you will be redirected to a more detailed status page
screenshot from 2018-05-31 18-03-26

and all mail is send in one task! no error handling needed

treba123:
I'm not a dev but I'm against this. It doesn't scale well. What if we have 50k users? How long would it take to even schedule a task?

ihucos:
> UNIX admin style 
Which I totally prefer in most cases, I am happy that we don't use too many cloudy external services.

ihucos:
TODO: communicate too many mails in _from_ limitation

ihucos:
This pull request adds django admin reporting to the newsletter with the ParallelEmailMixin
revert admin mail reporting before merging.
There is one limitation: all recepient mails together for one mail should not be bigger than 255 charachters

ihucos:
I totally think we need this, but right now there is not so much time and in the future there also will not be, so I am closing this for now and hope that we can merge it another day.

PR #1416, opened 2018-05-31
css for IE11

phillimorland:
Added a width to the box, comparable to the button, fine in chrome, firefox, IE and on responsive, I know you said something about only using min-width? can't work with percentage. Added a min height to text area as Carolin requested it be larger, specially as IE cannot resize.


fuzzylogic2000:
@phillimorland Nice! I just chnaged it to min-width instead of width. Like you already proposed. [MDN says min-width is fine with percent](https://developer.mozilla.org/en-US/docs/Web/CSS/min-width) and it also looks fine in ie11.

PR #1423, opened 2018-06-05
Configure sentry

ihucos:
One small behavioral side effect: `python manage.py process_tasks` now makes more output in production.
The local settings then should be something like this:

```
RAVEN_CONFIG = {
    'dsn': 'https://username:password@server.io/something',
    'release': git_commit_hash,
}
```
@treba123 will take care of a self hosted sentry server.

### DIFF
--- a/meinberlin/config/settings/base.py
+++ b/meinberlin/config/settings/base.py
@@ -56,6 +56,7 @@ INSTALLED_APPS = (
     'ckeditor_uploader',
     'capture_tag',
     'background_task',
+    'raven.contrib.django.raven_compat',
 
     'adhocracy4.actions.apps.ActionsConfig',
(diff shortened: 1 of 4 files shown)


ihucos:
Merging now and deploying without the sentry configuration SHOULD not result in any error

fuzzylogic2000:
Ok, cool! When do we merge and make sure, that SHOULD is also does? Now or Monday morning?

ihucos:
> Ok, cool! When do we merge and make sure, that SHOULD is also does? Now or Monday morning?
Everything no problemo, click on the green merge button, now :-)

ihucos:
Robert already made the config in locals.py, so it should then work in dev

PR #1439, opened 2018-06-07
addition of rest of components, added one tag to hero

phillimorland:
I have done an indent check, hope its ok! The following components I have not added:
category-formset__form - Cannot locate a use of it
simple_page - Cannot locate a use of it
select2 -Only referenced in js files
lr_bar - Only referenced in import doc
select_dropdown - Only referenced in import doc
map - javascript


fuzzylogic2000:
@phillimorland The category-formset and the select_dropdown are actually from the same js-stuff (that puts the icons in the category-dropdowns either to select icons that are connected to the category or put them in the dropdown with the category for the user to select.
The simple page is a wagtail thing. The classes in the component lr_bar are all called something lr-bar. I will add these components.
I am unsure about the select2.
And I don't know how to add the map. I guess there are things like popups and so on that are worth to add, but I think without actually adding the map. Hrrmpf. Will use my fresher brain tomorrow.

PR #1442, opened 2018-06-11
polls to a4

ihucos:


meinberlin/apps/contrib/predicates.py
```diff
@@ -1,11 +0,0 @@
-import rules
```
ihucos: I moved this file to adhocracy4/projects/predicates.py and refactored imports

meinberlin/config/settings/base.py
```diff
@@ -203,7 +203,7 @@
 
 IMAGE_ALIASES = {
     '*': {
-        'max_size': 5*10**6,
+        'max_size': 5 * 10**6,
```
ihucos: style

meinberlin/config/settings/base.py
```diff
@@ -316,14 +316,14 @@
 }
 
 BLEACH_LIST = {
-    'default' : {
-        'tags': ['p','strong','em','u','ol','li','ul','a'],
+    'default': {
+        'tags': ['p', 'strong', 'em', 'u', 'ol', 'li', 'ul', 'a'],
```
ihucos: style

meinberlin/config/settings/base.py
```diff
@@ -316,14 +316,14 @@
...
         'attributes': {
             'a': ['href', 'rel'],
         },
     },
     'image-editor': {
-        'tags': ['p','strong','em','u','ol','li','ul','a','img'],
+        'tags': ['p', 'strong', 'em', 'u', 'ol', 'li', 'ul', 'a', 'img'],
```
ihucos: style

meinberlin/config/settings/base.py
```diff
@@ -460,8 +460,7 @@
 
 A4_DASHBOARD = {
     'PROJECT_DASHBOARD_CLASS': 'meinberlin.apps.dashboard.TypedProjectDashboard',
-    'BLUEPRINTS': 'meinberlin.apps.dashboard.blueprints.blueprints'
-}
+    'BLUEPRINTS': 'meinberlin.apps.dashboard.blueprints.blueprints'}
```
ihucos: style

meinberlin/config/settings/base.py
```diff
@@ -471,7 +470,11 @@
 DEFAULT_LANGUAGE = 'de'
 
 CSP_STYLE_SRC = ("'self'", "'unsafe-inline'")
-CSP_IMG_SRC = ("'self'", "data:", "*.tile.openstreetmap.org", "https://maps.berlinonline.de")
+CSP_IMG_SRC = (
+    "'self'",
+    "data:",
+    "*.tile.openstreetmap.org",
+    "https://maps.berlinonline.de")
```
ihucos: style

MagdaN:
I was thinking about this and I am in favour of leaving the polls as they are in mein Berlin. Using them in opin was a compromise but in general we would like to implement something different. So it is great that they are in A4 now so we can use them and change them the way we want them for opin and beteiligung.in but those changes should not effect meinberlin - and the other way around, changes in meinberlin polls should not affect all polls.

fuzzylogic2000:
Yes! We discussed the polls so much, and wanted them to be different from the mB ones on our platform(s). I think @MagdaN proposal is a good solution. It's great, that we have a first version of polls in the core then, but we can change them there if need be and keep the ones with the given requirements in mB.

PR #1443, opened 2018-06-11
Ka 2018 06 component library

fuzzylogic2000:


MagdaN:
@fuzzylogic2000 whats needs to be done here?

PR #1445, opened 2018-06-11
Contributions listing

ihucos:
Something that bothers me when I am using mein.berlin.de personally: I can't see a listing of all activities in the platform.

screenshot from 2018-06-11 18-57-06

@CarolingerSeilchenspringer 
Give me the OK for this, then maybe some different content and translations, then we do the technical review.


fuzzylogic2000:
I think we shouldn't merge this. It is not for the users, but only for initiators and clogs up the dashboard (or whatever you like to call it). And just putting in stuff like that, when mein.berlin doesn't have a user profile-thingy on purpose doesn't seem like a good idea to me.
Maybe we can put a real user profile a bit like the one in ae into bet.in?

Issue #1449, opened 2018-06-13
Bearbeiten button on Vorhaben detail page


phillimorland:
In dev: button is active on page load displaying 1 menu option, "Bearbeiten", clicking has no effect

PR #1455, opened 2018-06-14
dropdown fix

phillimorland:
separate dropdown and lr-bar__right classes into 2 divs due to styling on lr-bar__right


fuzzylogic2000:
fixes #1449

PR #1466, opened 2018-06-27
Pm 2018 06 list item fixes#1463

phillimorland:
change background colour to lighter gray


phillimorland:
first commit should be on branch pm-IE11-moderator-button, shall I delete that commit from this branch?

phillimorland:
not finished as just noticed odd behavior that I want to check

PR #1468, opened 2018-06-27
Ka 2018 06 add comment export to idea modules and polls

fuzzylogic2000:
This PR fixes the problem with the related query names @ihucos and me identified.
Additionally it adds the comment export to ideas, mapideas, kiezkasse_proposals, budget_proposals, topics, and maptopics. I only tested it for brainstorming, mapbrainstorming, and budgeting so far.

Still missing:

- [x] manual tests for the finction in the untested modules
- [x] comment export for polls
- [x] dashboard view with two export buttons
- [ ] tests

I briefly looked at the export-button-thing. It's in meinberlins export-app, but seems not that easy to solve.


ihucos:
> I briefly looked at the export-button-thing. It's in meinberlins export-app, but seems not that easy to solve.
Yes

meinberlin/apps/kiezkasse/exports.py
```diff
@@ -32,3 +34,34 @@ def get_queryset(self):
...
+
+    def get_queryset(self):
+        comments = (Comment.objects.filter(
+                    kiezkasse_proposal__module=self.module) |
+                    Comment.objects.filter(
+                    parent_comment__kieszkasse_proposal__module=self.module))
+
+        return comments
+
+    def get_base_filename(self):
```
ihucos: I am thinking that we don't need that part, that is already implemented in BaseItemExportView: https://github.com/liqd/adhocracy4/blob/029bd5fb27236f9c1c042e32b126e41fe4f9a73e/adhocracy4/exports/views.py#L103

meinberlin/apps/ideas/exports.py
```diff
@@ -31,3 +33,33 @@ def get_queryset(self):
...
+                            a4_export_views.BaseItemExportView):
+
+    model = Comment
+
+    fields = ['id', 'comment', 'created']
+
+    def get_queryset(self):
+        comments = (Comment.objects.filter(idea__module=self.module) |
+                    Comment.objects.filter(
+                    parent_comment__idea__module=self.module))
```
ihucos: Just out of coriosity, that could also be written diffrently:

`Comment.objects.filter(Q(idea__module=module) |  Q(parent_comment__idea__module=module`

Confirming this:
```
>>> q2 = str((Comment.objects.filter(idea__module=module) |  Comment.objects.filter(parent_comment__idea__module=module)).query)
>>> q1 = str(Comment.objects.filter(Q(idea__module=module) |  Q(parent_comment__idea__module=module)).query)
>>> q1 == q2
True
```

I think that with set operations the django orm sometimes uses subqueries, wich are inneficient, but in that case he optimises that away, so its the same

PR #1571, opened 2018-09-25
Pm 2018 09 idea project ref

### LINKED ISSUE #246: No project title when creating new idea
When I try to create a new idea, the project the idea is created in, is not referenced. This might be confusing, especially if you leave the tab in my browser open and come back later. It would be great if we show at least the project’s title, short description and maybe organization.

bildschirmfoto 2017-04-24 um 15 33 41

phillimorland:
fixes #246


phillimorland:
Agreed! I changed the class name of the heading from "item-detail-2__project-title" to "project-header__title" on last commit as it seemed like a more appropriate name, didn't realise it would change the size, thought that was done by the h3.
In hindsight it doesnt need a class name at all

PR #1624, opened 2018-10-24
Mn 2018 10 add vector tiles

MagdaN:
This PR only works when disabling the CSP middleware even on local machines.

@treba123 do you have an idea how to fix this?


treba123:
Locally the csp middleware should be report-only...if that still breaks the PR, something deeper must be happening :/
Will look into it on tuesday

PR #1631, opened 2018-10-25
adding css for newly added class names from adhocracy to change colou…

### LINKED ISSUE #1368: "unpublish" could easily be hit by mistake
If I want to hit the preview, I could easily miss and unpublish the project.
There could be an javascript alert with a are you sure question.

And the german translation "unpublizieren" seems awkward to me.

phillimorland:
…r of progress box

fixes #1368

relates to adhocracy - https://github.com/liqd/adhocracy4/pull/288


phillimorland:
ahh didn't know we used labels for pull requests as well as issues, will do so in future.
I forgot the circle is pink while the form is being filled, I will change the pull request to only be the tick, I think a new colour would be confusing.

PR #1648, opened 2018-10-31
Streamline app initialization

treba123:
This brings us in line how other packages work and avoids bugs.
For example background tasks fail to discover tasks.py if an
app is not initialized like this.


treba123:
Hm interesting...why did that one fail? Will have another look, actually it shouldn't be necessary :/
Anyway, thanks for reviewing!

PR #1657, opened 2018-11-05
adding accordion to homepage cms

phillimorland:
still need the design but it's all there
Also need to commit migration once questions are cleared up


meinberlin/apps/cms/models.py
```diff
@@ -73,6 +73,7 @@ class HomePage(Page):
         ('columns_text', cms_blocks.ColumnsBlock()),
         ('projects', cms_blocks.ProjectsWrapperBlock()),
         ('activities', actions_blocks.PlatformActivityBlock()),
+        ('accordion', cms_blocks.DocsBlock())
     ])
```
phillimorland: Is it ok to call it accordion when block has different name?

meinberlin/apps/cms/templates/meinberlin_cms/blocks/docs_block.html
```diff
@@ -16,8 +19,8 @@ <h2>
 
     <div
         class="accordion__body collapse"
-        id="accordion-{{ index }}-body"
-        aria-labelledby="accordion-{{ index }}-title">
+        id="accordion-{{ identifier }}-body"
+        aria-labelledby="accordion-{{ identifier }}-title">
```
phillimorland: index wasn't working so added the identifyer stuff from SPD accordion

meinberlin/apps/cms/templatetags/menu.py
```diff
@@ -11,3 +11,8 @@ def get_menu(title):
 
     if menu is not None:
         return menu.items.all()
+
+
+@register.simple_tag()
+def get_identifier(item):
+    return id(item)
```
phillimorland: tried adding this to it's own block_tag.py file but it wouldn't recognise it and listed the acceptable places to add a template tag, did I miss a step?

meinberlin/apps/cms/models.py
MagdaN: yes

meinberlin/apps/cms/templates/meinberlin_cms/blocks/docs_block.html
```diff
@@ -1,9 +1,12 @@
 {% load wagtailimages_tags wagtailcore_tags %}
+{% load wagtailcore_tags block_tags %}
```
fuzzylogic2000: you can and should add all the templatetags to one line. So ehre you just add block _tags to the first line between the brackets and all template tags are loaded.

meinberlin/apps/cms/templatetags/menu.py
fuzzylogic2000: You just need to add the tag to a file named whatever you ioad in the template and add that file to a folder templatetags and it should be found. Did that not work?

fuzzylogic2000:
@phillimorland Oh no! You were very smart to leave the migrations out! I forgot about the storefront...

PR #1658, opened 2018-11-05
simplifying map component - first step

phillimorland:
Ok this is mostly a mess, I added 2 functions from the old plans_map and stripped them down in hope that would at least pass the data which i could access from console but i just ended up confusing myself more. 
The current structure is same as the wireframes but the sticky behaviour is not correct, due to nesting I think we will need to use JS maybe [stickybits](https://dollarshaveclub.github.io/stickybits/) or [react example](https://reactjsexample.com/makes-native-position-sticky-statefull-and-polyfills-the-missing-sticky-browser-feature/)
Sorry its such a mess!


MagdaN:
@phillimorland because you asked how to do this, I moved each component to a separate file. Let me know if something is unclear to you!

PR #1659, opened 2018-11-05
revert updating mapbox-gl

fuzzylogic2000:
The [mapbox-gl update](https://github.com/liqd/a4-meinberlin/commit/d8e3b03d1a20db88ee366a836a13ff2acdf5cd36) gave us strange linting errors for all the jsx-files. Downdating fixes it. Should be investigated further.


fuzzylogic2000:
found a much better fix

PR #1683, opened 2018-11-27
dev: show more logging output by default

treba123:
Update dev logging again to make it more expressive


treba123:
Jep, raven is just for sentry

treba123:
Updated the PR to silence the raven warning

PR #1685, opened 2018-11-28
adding hompage classname to rich text bloks for custom styling

phillimorland:
adding a class name to div so we can style rich text blocks according to design


phillimorland:
I have applied the changes to all accordions, the classname was for richtext styling for landing page to enable us to adapt it to the design, sorry, should be more clear in title. Does that help?

PR #1688, opened 2018-11-29
Ka 2018 11 add async task for bplan location info

fuzzylogic2000:


meinberlin/apps/bplan/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+import urllib
+
+from background_task import background
+
+from adhocracy4.administrative_districts.models import AdministrativeDistrict
+from meinberlin.apps.bplan.models import Bplan
+
+
+@background(schedule=0)
+def get_location_information(bplan_id):
```
MagdaN: you could split up the code into several functions, might be easier to read

fuzzylogic2000:
Yeahhh, I also thought of that, but was tired looking at the code... Will do that!

PR #1691, opened 2018-11-29
Pm 2018 11 storefront cols

phillimorland:
i've used too much nesting, I need to neaten and add classnames where possible also the way i did the image is crappy, I forgot about that till the commit, was wondering is we could use a tag to add it as making an img a background image seems to be tricky. 
Hope it's ok and kind of understandable!


meinberlin/apps/cms/templates/meinberlin_cms/home_page.html
```diff
@@ -32,16 +32,30 @@ <h1 class="hero-unit__title u-first-heading">{{ self.title }}</h1>
...
-                      <a href="{{ item.link_page.url }}"></a>
-                      {% image item.header_image max-1920x560 %}
-                   </div>
-             {% endfor %}
-           </div>
+            {% image self.storefront.image max-400x600 as storefront_image %}
+            <div class="l-wrapper">
+                <div class="storefront">
+                    <div class="storefront__col">
+                        <div class="storefront__item storefront__img" style="width=100%; background:url({{ storefront_image.url }}) no-repeat; background-size: cover;">
```
fuzzylogic2000: we can add it as a background-image and style it further in the css, like here: https://github.com/liqd/a4-meinberlin/blob/77887fa95aad2f559fc12781af91f325f4ddb220/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html#L36 and https://github.com/liqd/a4-meinberlin/blob/77887fa95aad2f559fc12781af91f325f4ddb220/meinberlin/assets/scss/components/_project_header.scss#L16

PR #1698, opened 2018-12-03
restructure header

fuzzylogic2000:
There are a few things still missing (mainly the styling and font of the headline next to the logo, but also the check how well it works with all the pages and alerts). But I propose that we still merge it into master if there are no other big issues, so that @CarolingerSeilchenspringer and @szukowski can look at it and talk about the issues and changes that still need to be done with us.

Already in this PR are:
- the new styling of the header (apart from the text next to the logo...)
- the help and feedback pages are settable in wagtail

Right, and sorry for all the migrations! But as I already commited and pushed I thought I'd just leave them.... And wagtail is always migration-cluttered anyway.


meinberlin/assets/scss/components/_header.scss
```diff
@@ -7,27 +7,52 @@ $header-margin-bottom: 25px;
...
+    justify-content: space-between;
     line-height: 1;
 }
 
+.main-nav__nav-item {
+    flex: 1;
+}
+
+.main-nav__nav-item--right {
+    display: flex;
```
fuzzylogic2000: this breaks the positioning of the mobile dropdown

meinberlin/templates/header.html
```diff
@@ -3,36 +3,51 @@
...
+                                <a class="main-nav__link" href="{{ item.url }}">{{ item.title }}</a>
+                            </li>
+                        {% endfor %}
+                    </ul>
+                </div>
+            </div>
+            <div class="main-nav__logo-item">
+                <a href="/" class="main-nav__logo" rel="home">
+                    <img src="{% static 'images/LOGO.png' %}" alt="mein.berlin"/>
+                    <div class="main-nav__headline">{% trans 'Participation platform of Berlin!' %}</div>
```
fuzzylogic2000: As the image next to it has to be centered, this has (?) to be positioned absolute, but I just can't get it vertically centered.

PR #1725, opened 2018-12-10
Pm 2018 12 map issues

phillimorland:
need to look into ipad pro size, it's too big to be a tablet with our current breakpoints.

### DIFF
--- a/meinberlin/assets/scss/components/_map.scss
+++ b/meinberlin/assets/scss/components/_map.scss
@@ -127,7 +127,6 @@
 
 .map-list-combined__list {
     flex: 0 1 44vh;
-    background-color: $bg-secondary;
 
     @media (min-width: $breakpoint-xs) {
         flex-grow: 2;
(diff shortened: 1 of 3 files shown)


meinberlin/assets/scss/components/_map.scss
```diff
@@ -95,6 +95,9 @@
 
     @media (min-width: $breakpoint) {
         flex-direction: row;
+    }
+
+    @media (min-width: $breakpoint-md + 0.1em) {
```
fuzzylogic2000: You don't need to add anyting for min-width media-queries

meinberlin/assets/scss/components/_switch.scss
```diff
@@ -12,36 +12,38 @@
...
+        width: 47px;
+        border: 1px solid $gray;
     }
 
     label:after {
-        background: rgb(255, 255, 255);
+        background: $body-bg;
         border-radius: 16px;
-        box-shadow: 0 0 5px rgba(0, 0, 0, 0.3);
+        // box-shadow: 0 0 5px rgba(0, 0, 0, 0.3);
```
phillimorland: have commented this stuff as final design for switch not in zeplin, only wireframe, so might need some of this

PR #1726, opened 2018-12-10
Ka 2018 12 project count

fuzzylogic2000:
I wasn't quite sure which projects we said we wanted to count for the storefront anymore. So I counted the ones with active or future phases.
@CarolingerSeilchenspringer Do we also want to add the active plans to the count or did we decide that we didn't want to do that?

### DIFF
--- a/meinberlin/apps/cms/models/storefronts.py
+++ b/meinberlin/apps/cms/models/storefronts.py
@@ -74,9 +74,13 @@ class Storefront(ClusterableModel):
 
     @cached_property
     def num_projects(self):
-        return Project.objects.all()\
-            .filter(is_draft=False, is_archived=False, is_public=True)\
-            .count()
+        projects = Project.objects.all()\
+            .filter(is_draft=False, is_archived=False, is_public=True)
+        active_project_count = 0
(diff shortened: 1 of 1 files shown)


meinberlin/apps/cms/models/storefronts.py
```diff
@@ -74,7 +74,13 @@ def num_entries(self):
...
     @cached_property
     def num_projects(self):
-        return Project.objects.all().count()
+        projects = Project.objects.all()\
+            .filter(is_draft=False, is_archived=False, is_public=True)
+        active_project_count = 0
+        for project in projects:
+            if project.active_phase or project.future_phases:
+                active_project_count += 1
+        return active_project_count
```
MagdaN: would it be easier to just count active phases?

fuzzylogic2000: Yes, a bit. But I think, we said, we also wanted to count the future ones? @CarolingerSeilchenspringer Currently we count all projects with active or future phases, that are published, not archived and not private. Is that like we want it? Or do we only want the active, not the future ones?

PR #1747, opened 2018-12-17
Pm 2018 12 style filter

### LINKED ISSUE #1721: switch broken (goes with #744)
can't use switch
Seen on ipad 5/safari 

bildschirmfoto 2018-12-10 um 14 44 30

### LINKED ISSUE #1722: design not as responsive as it should be (goes with #744)
Samsung Galaxy tab 4, Chrome

bildschirmfoto 2018-12-10 um 15 26 22

### LINKED ISSUE #1723: map should not be sticky or is too big on mobile (goes with #744)
If I use the map I cannot get back to the header/switch (I have to reload).
Seen on Iphone 7

bildschirmfoto 2018-12-10 um 15 32 11

phillimorland:
the max and min heights need some looking at, i reverted them to begin with but when that didn't work i added a min height but it's just a random number, need work! will add issue.
Also no translations 
fixes #1721
fixes #1722
should fix #1723


fuzzylogic2000:
screenshot from 2018-12-17 17-58-46

phillimorland:
What browser do you use again? because the map is still sticky for me and the list isn't over the footer although that is currently an issue for some browser screen sizes, I don't think it ever stopped being an issue did it?  

- Font size is from design and yes I agree it's pretty big! 
- There is no mobile design so only added the reduced padding but thought anything more would be time waste?

phillimorland:
think thats everything fixed except footer issue

PR #1752, opened 2018-12-18
Pm 2018 12 mobile header

phillimorland:
added for small and big mobiles as well as small tablets


phillimorland:
ok, not sure i totally follow, you mean with a display none class name? I thought of that butrest of the header uses media q's for the disply non instead of class names

phillimorland:
fixed :p

PR #1759, opened 2018-12-20
styling filters

phillimorland:
- [ ] find way to add topics to filter?
- [ ] style for mobile (still awaiting final designs)


meinberlin/apps/plans/assets/ListMapBox.jsx
```diff
@@ -28,7 +28,7 @@ class ListMapBox extends React.Component {
   }
 
   isInFilter (item) {
-    return (this.state.topic === '-1' || this.state.topic === item.topic) &&
+    return (this.state.topic === '-1' || this.state.topic === item.topic || this.state.topic === item.topic.toLowerCase()) &&
```
fuzzylogic2000: Let's also set item.topic to lower case, just to be sure.

PR #1785, opened 2019-01-07
Pm 2019 1 storefront statusbar

phillimorland:


fuzzylogic2000:
Good morning, I will try to fix it, so that we can also have the real percent!

PR #1800, opened 2019-01-10
Pm 2019 01 switch map adjustments

phillimorland:
a bit of a bitty one, sorry, should have done another pull request for the storefront bits but it only need tiny change. This is what shoudl have been done. 
Map:
- Adding closable info box 
- Removing bottom margin between search input and dropdown
Map Switch
- Colour update
- Wrapping in l-wrapper as per designs
Storefront
- Adding min height to md tiles
- Adjusting status position
General: related to Storefront
-  Increasing small font size (storefront labels)
- Changing link colour on site (tile text colour)

Sorry!!!! I got carried away


MagdaN:
@phillimorland could you please remove the merge conflicts? We are not sure right now, what needs to stay and what needs to go in the scss ...

phillimorland:
oupps sorry didn't see that, will do! not sure by what you mean with what needs to stay?

meinberlin/apps/plans/assets/PlansMap.jsx
```diff
@@ -260,6 +273,12 @@ class PlansMap extends React.Component {
 
           </form>
         </div>
+        {this.state.showInfoBox &&
+        <div className="map-infobox">
+          <button className="infobox__close" id="close" onClick={this.closeInfoBox.bind(this)}><i className="fa fa-times" /></button>
```
fuzzylogic2000: I think this should have an aria-label like "close infobox", maybe you can check how we did it for other things closing something (by greping fa-times).

meinberlin/apps/plans/assets/PlansMap.jsx
```diff
@@ -260,6 +273,12 @@ class PlansMap extends React.Component {
...
+          <i className="fa fa-info-circle " /><span>{django.gettext('Not all plans are shown on map, please check list view.')}</span>
```
fuzzylogic2000: As this has a text, it can be aria-hidden

meinberlin/apps/plans/assets/PlansMap.jsx
phillimorland: the ones that weren't hidden were labelled close but as this could be confusing i added your suggested text also the correct text for the info box, all ready I hope!

PR #1814, opened 2019-01-15
plans list - style tablet filter bar

fuzzylogic2000:


MagdaN:
@fuzzylogic2000 in order to remove the warning I did a lot of refactoring (which I think in the end was not needed, but I did not want to undo it). Basically I moved some of the code from your functions to extra components (but they are in the same file). But what was actually fixing the problem was adding a div with a key around each list and the key is the current index.

fuzzylogic2000:
@MagdaN Nice! Makes it also easier to read!

PR #1816, opened 2019-01-15
adding corner image

phillimorland:
add image to corner of list items


fuzzylogic2000:
@MagdaN Didn't you implement a way to identify external projects in your last PR? Or is there currently only the right url, and we still need to add a way to do so?

MagdaN:
@phillimorland I will add the information on the whether a project is external to this PR and as soon as I am finished you can go on?

PR #1817, opened 2019-01-15
Pm 2019 01 map teaser

phillimorland:
- [x] add filter

- [x] update onclick link

- [x] style filter


meinberlin/assets/scss/utility.scss
```diff
@@ -80,6 +80,14 @@
     position: relative;
 }
 
+.u-z-index-top {
```
fuzzylogic2000: This should be called u-z-index-2. :) And the other one u-z-index-1

MagdaN:
@phillimorland I added the filter now to your wagtail block. Please let me know if I should explain what I did.

PR #1823, opened 2019-01-17
working basic popup, possible code in commented out if statement

phillimorland:
Very rough, but wanted to push incase someone gets round to it tomorrow, the commented out code is in correct formate but not sure how to do it with if statements or seperate functions or something?!


MagdaN:
@phillimorland I looked at your code and realised that leaflet wants a html-string (and you already transformed already a lot of your code to it, sorry, I did not see that before), but I found a way to still use a react component so that you can work at the jsx as before. I also transformed your code back to jsx so that you should be able to keep on working. There is an new file now PopUp.jsx -> The PopUp react component just has one property which is item. So via this.props.item... you can access all the information you need.

phillimorland:
Still have an issue with the z-index but don't want to sink time into it until we know what should, have left comment on zeplin, is that correct place? 
Have also commented about the project with plan popup style as that has to be slightly different to designs as we are trying to not over complicate things. 
Other than those 2 things this is ready I think :)

meinberlin/assets/scss/components/_map.scss
```diff
@@ -16,8 +16,8 @@
     flex-direction: column;
     justify-content: space-between;
 
-    margin: 0.5em;
-    width: 14.5em;
+    padding: $padding;
+    width: 14.5rem;
 }
```
phillimorland: is this ok? couldn't think of where else we used map popups so thought this would be a fine solution?

PR #1835, opened 2019-01-22
checking for projectcontainer attr to get project count for container tile

### LINKED ISSUE #1832: Container tile does not show number of projects

phillimorland:
fixes #1832

### DIFF
--- a/meinberlin/apps/plans/assets/PlansList.jsx
+++ b/meinberlin/apps/plans/assets/PlansList.jsx
@@ -27,7 +27,7 @@ class PlansList extends React.Component {
   }
 
   renderListItem (item, i) {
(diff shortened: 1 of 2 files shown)


meinberlin/apps/plans/assets/PlansList.jsx
```diff
@@ -27,7 +27,7 @@ class PlansList extends React.Component {
   }
```
phillimorland: I didn't change participationStatusClass to statusClass, i'm a bit confused if this is an artifact from somewhere? what is the best procedure in situations like this?

fuzzylogic2000: Hmm, don't know. Change it back?

PR #1842, opened 2019-01-24
removing un-needed element to mapteaser block

phillimorland:
removing richtext icon block from mapteaser block
also added tiny fix by removing z-indexes I added, i thought I needed them but I think that was before I had the background image behaving properly, they don't seem to be needed now and they mess with user dropdown


MagdaN:
@phillimorland looks good but how are we adding the icon? Should that be a static file and always the same or just add it inside the richtextfield?

MagdaN:
ah I see we can use the richtextfield

PR #1845, opened 2019-01-24
restructure mobile btn group to be radio btns for a11y

phillimorland:
is it ok to leave the btn class names as we want them to look like that? 
Also added tiny fix quick to accordions which I think I broke with map teaser styling

### DIFF
--- a/meinberlin/apps/plans/assets/MapListSwitch.jsx
+++ b/meinberlin/apps/plans/assets/MapListSwitch.jsx
@@ -25,15 +25,15 @@ class MapListSwitch extends React.Component {
       return (
         <div>
           <div className="u-spacer-left u-spacer-right">
-            <div className="switch-group" role="group" aria-label={django.gettext('Map list switch')}>
-              <div className="radio-group">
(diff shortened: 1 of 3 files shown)


meinberlin/apps/plans/assets/MapListSwitch.jsx
```diff
@@ -26,9 +26,13 @@ class MapListSwitch extends React.Component {
         <div>
           <div className="u-spacer-left u-spacer-right">
             <div className="switch-group" role="group" aria-label={django.gettext('Map list switch')}>
-              <div className="btn-group">
-                <button className={!this.props.displayMap ? 'btn btn--light switch--btn active' : 'btn btn--light switch--btn'} onClick={this.props.showList}><i className="fa fa-list" /></button>
-                <button className={this.props.displayMap ? 'btn btn--light switch--btn active' : 'btn btn--light switch--btn'} onClick={this.props.showMap}><i className="fa fa-map" /></button>
+              <div className="radio-group">
+                <label className={!this.props.displayMap ? 'btn btn--light switch--btn active' : 'btn btn--light switch--btn'} onClick={this.props.showList} htmlFor="show_list">
```
fuzzylogic2000: the label should also have an sr-only span actually labelling it. Maybe if we find something that makes sense, we can also remove the aria-label from the outer div.

PR #1879, opened 2019-01-30
add title search to secondary participation filters

fuzzylogic2000:


fuzzylogic2000:
@phillimorland It's fine now! The trim does what I thought it would: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim
And the ids match.

PR #1887, opened 2019-01-31
Mn 2019 01 more filter buttons

MagdaN:


meinberlin/apps/plans/assets/ListMapBox.jsx
```diff
@@ -81,11 +81,9 @@ class ListMapBox extends React.Component {
   }
 
   isInTitle (title) {
-    let titleLower = title.toLowerCase()
-    if (titleLower.indexOf(this.state.titleSearch.toLowerCase()) === -1) {
-      return false
-    }
-    return true
+    let titleLower = title.toLowerCase().trim().replace(/\s/g, '')
+    let searchString = this.state.titleSearch.toLowerCase().trim().replace(/\s/g, '')
```
fuzzylogic2000: The titleSearch is already stored in lower case, but better safe than sorry!

meinberlin/apps/plans/assets/ListMapBox.jsx
```diff
@@ -81,11 +81,9 @@ class ListMapBox extends React.Component {
   }
 
   isInTitle (title) {
-    let titleLower = title.toLowerCase()
-    if (titleLower.indexOf(this.state.titleSearch.toLowerCase()) === -1) {
-      return false
-    }
-    return true
+    let titleLower = title.toLowerCase().trim().replace(/\s/g, '')
```
fuzzylogic2000: Do we want to replace all whitespace? What if someone searches for "der Verwaltung" and then this can't be found because "derverwaltung" is not in the title?
Oh, maybe we need to think about doing different searches for every word. Otherwise "verwaltung für bla" might have a hit while "verwaltung bla" doesn't.

meinberlin/apps/plans/assets/FilterSecondary.jsx
```diff
@@ -47,9 +47,8 @@ class FilterSecondary extends React.Component {
 
   changeTitleSearch (e) {
     let value = e.target.value
-    let searchTerm = value.toLowerCase().trim()
```
fuzzylogic2000: alright, forget the other comment. :)

fuzzylogic2000:
Ah, good morning! By replacing whitespace in searchterm and title we still find two words.

PR #1942, opened 2019-02-06
Pm 2019 02 mobile overlay

phillimorland:


meinberlin/assets/scss/components/_filter_bar.scss
```diff
@@ -33,6 +33,11 @@ $divider-margin: 1.25rem;
     margin-bottom: $r-spacer;
 }
 
+.filter-bar--stacked.filter-bar--open {
+    position: absolute;
+    z-index: 2000;
+}
+
```
phillimorland: is this the best way to do the overlay? feels like I may have oversimplified a bit?

meinberlin/apps/plans/assets/FilterNav.jsx
```diff
@@ -56,6 +56,12 @@ class FilterNav extends React.Component {
     })
   }
 
+  closeSecondaryFilters () {
+    this.setState({
+      displaySecondaryFilters: !this.state.displaySecondaryFilters
```
fuzzylogic2000: if it's just to close it, I would set it to false to make sure it's always doing that

meinberlin/assets/scss/components/_filter_bar.scss
```diff
@@ -144,6 +144,11 @@ $divider-margin: 1.25rem;
         padding-left: 0;
     }
 
+    .btn-primary {
```
fuzzylogic2000: Don't we have a btn--primary normally? This seems to be quite special to something? Maybe we need to add another class and not use that one?!

meinberlin/assets/scss/components/_filter_bar.scss
```diff
@@ -289,3 +295,23 @@ $divider-margin: 1.25rem;
...
+.filter-secondary__modal-header {
+    border-bottom: none;
+    padding-bottom: 0;
+}
+
+.filter-secondary__modal-content {
+    background-color: $bg-secondary;
+    font-size: $font-size-base;
+
+    .btn-primary {
```
fuzzylogic2000: Ah, you use that to style it differently! But still, maybe another class-name and try what happens if you set it to btn--primary

Issue #1967, opened 2019-02-08
add missing dashboard-tabs to external projects


fuzzylogic2000:
in both missing:
- plans

missing in external projects:
- district and location
- topics

(This was part of #1933 )

Issue #1933, opened 2019-02-05
district, topic and localisation for external projects and bplans


CarolingerSeilchenspringer:
external projects also need district, topic and localisation as the other projects do. 

same is true for b-plans as not all of them come via imperia

fuzzylogic2000:
@CarolingerSeilchenspringer Obviously true for external projects!
But kind of only partly for bplans. We need to add the topics for the Bplans as well, but I think we shouldn't add the location and district to the dashboard as we get that automatically!?

CarolingerSeilchenspringer:
@fuzzylogic2000 the senatsverwaltung für Stadtentwicklung sometimes also have bplans but don't use imperia. They start them via meinberlin.

fuzzylogic2000:
@CarolingerSeilchenspringer That should be fine as long as they fill in the bezeichner-field, which should be there. As soon as they save the bplan, we try to get district and localisation from the other bplan-API, just as we do for the stuff filled in with imperia. So, we should be fine!

CarolingerSeilchenspringer:
@fuzzylogic2000 I will think of a helptext for the Bezeichner (Identifier) that works for both, SenStadt and Imperia

fuzzylogic2000:
We said we will:
- add both tabs (Ort and topics) to external projects
- add topics to Bplans
- add a helptext for the identifier-field of bplans
- fix #1929 at the same time (wording change tab name/title)

CarolingerSeilchenspringer:
@fuzzylogic2000 as just discussed: no need to add topics to Bplans, just always set them on Stadtentwicklung, no matter if they come from imperia or via meinBerlin

fuzzylogic2000:
@CarolingerSeilchenspringer Do you already have a helptext for the identifier?

fuzzylogic2000:
@CarolingerSeilchenspringer Can you please give me the helptext for the identifier?

CarolingerSeilchenspringer:
"Der Bezeichner des Bebauungsgplans muss identisch mit dem Bezeichner im FIS-Broker sein. So werden Informationen zum Bezirk und zur Verortung automatisch hinzugefügt."

PR #1968, opened 2019-02-08
dashboard: add topics and plans to external projects

fuzzylogic2000:
Ok, I am already spending much too much time on this. I tried to add the same tabs for topics and location (and plans) to the external projects as in the projects. But I couldn't get it to work. Now I added the topics to the main settings tab of the bplans. The same didn't work for the point for external projects. Even though I also added the right widget to the external-project form, it only looked right and didn't work (existing points were not shown).
This was part of #1933, but I made a new issue to solve later: #1967


fuzzylogic2000:
Ok, as the topic is set automatically after a Bplan is saved, this PR does not make any sense. Will close.

PR #1987, opened 2019-02-11
working fix for removing label when nothing selected

phillimorland:
label caused switch issue


phillimorland:
should now be fixed

PR #1994, opened 2019-02-12
Ka 2019 02 another shot at the shadow

fuzzylogic2000:


fuzzylogic2000:
@phillimorland Yeah, don't know which is better

PR #2031, opened 2019-02-14
Project overview performance

treba123:
Some cheap optimizations


treba123:
The latest round doesn't enable cachalot by default, only when explicitly activated via the `local.py`. So I'd say we can merge it

fuzzylogic2000:
Yes, or we just wait another hour and are absolutely sure that we're fine! :)

PR #2034, opened 2019-02-14
Pm 2019 02 list width

phillimorland:
bit annoying having to add - margin but i think it's better than trying to remove just the outer margins on the tiles?


phillimorland:
huh I'll take another look, was that chrome?

phillimorland:
ok that is hopefully fixed!

phillimorland:
it very much is! um fuck

phillimorland:
can't reproduce error in chromium, @fuzzylogic2000 was that browser you were using?

fuzzylogic2000:
Haha! Today it's gone! So, it was my cache... Sorry!

PR #2078, opened 2019-03-06
package.json: Update babel

treba123:
This should allow us to use ES6/2015 features.

And some drive-by additions


treba123:
em vs rem error... @fuzzylogic2000 can you have a look when you have time?

treba123:
Seems to be because I updated bootstrap as a drive-by (security warning...)

MagdaN:
I get the same error as in the other project (euth): Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'
    at Module.eval (Modal.jsx:61)

treba123:
Alright, in the end everything turned out to be much simpler than expected...but some lessons learned :)
This should work now and the solution should be easily portable to all our projects without changes to .jsx files

PR #2083, opened 2019-03-07
adding if and class to only display horizontal images on lg screens

### LINKED ISSUE #2063: project title overlap with progress bar in project overview
on smaller screens (here: Mac 13,3-Zoll 2560 x 1600) a long title overlaps the progress bar in the horizontal tiles. 

bildschirmfoto 2019-03-04 um 14 17 51

This is what it looks like with the longest title possible with only one topic. 

bildschirmfoto 2019-03-04 um 14 23 25

@szukowski What would be your idea to solve this?

### LINKED ISSUE #2064: store window: district tile announces more results than there are if I click on them
store window tile shows ALL project of district, if I click and get to project overview, the default filter takes out all old projects and plans without beteiligung.

can we only count running projects with participation?

phillimorland:
fixes #2063 
fixes #2064

### DIFF
--- a/meinberlin/apps/cms/models/storefronts.py
+++ b/meinberlin/apps/cms/models/storefronts.py
@@ -50,12 +50,17 @@ class StorefrontItem(models.Model):
 
     @cached_property
     def district_project_count(self):
-        return Project.objects\
+        projects = Project.objects\
             .filter(administrative_district=self.district,
                     is_draft=False,
                     is_public=True,
                     is_archived=False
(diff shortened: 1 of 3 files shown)


phillimorland:
@fuzzylogic2000 the coverage failure is due to not enough of the site being tested, should I look into how to write a test for the thing i did?

fuzzylogic2000:
No, you don't need to write the first tests for that. I will write an issue. :)

PR #2095, opened 2019-03-12
Pm 2019 03 event tracking

phillimorland:
I think this will work, I will set up a report for the dev site so I can check it before the next release


MagdaN:
@phillimorland, thanks, yes, I think that should work (at least as far as I remember matomo). I just not totally understand what we need it for in this place?

phillimorland:
@MagdaN I spoke to @CarolingerSeilchenspringer about ways to track where we get drop offs in the signup process, there is a full form-analytics plugin that tracks this stuff but it's a yearly fee so I was hoping that once we could track clicks on the forms we might be able to do some rudimentary analysis maybe add some user journeys/funnels. Also we have no event tracking currently and if we want to use this more then I figured it was just as good a place to start as any, might help us see what we can get from event tracking. Is there a big disadvantage to having it in place?

fuzzylogic2000:
@phillimorland Are you still on this?

phillimorland:
I had forgotten about this to be honest! let me take another look at it after i've finished docs on cookie cutter

phillimorland:
We wanted to just try this on dev to begin with so I'll leave this pull request open till after the release then will finish and we can test it to see if it tracks what i'm hoping it will track

MagdaN:
@phillimorland I am fine with this, but we have functionality that makes it possible to only add the js code to the page it tracking is enabled and I was wondering if we could add because we would get error messages in our browser on pages where it is not enabled

PR #2100, opened 2019-03-14
add newsletter form for group members

MagdaN:


meinberlin/apps/newsletters/views.py
```diff
@@ -51,12 +52,13 @@ def get_form_kwargs(self):
     def get_success_url(self):
         return reverse('meinberlin_newsletters:newsletter-create')
 
+    def _check_permission(self, organisation, user):
```
fuzzylogic2000: Yeah! That's much easier to find now! For next time...

meinberlin/apps/newsletters/views.py
```diff
@@ -120,13 +135,24 @@ def get_email_kwargs(self):
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         kwargs['organisation'] = self.organisation
-        kwargs.pop('user')
+        user = self.request.user
```
fuzzylogic2000: Yeah, you are right, this change seems a bit scary, but worked locally for me! So, let's try around a LOT on dev!

PR #2111, opened 2019-03-19
Mn 2019 03 add tests

MagdaN:


fuzzylogic2000:
Should we already merge this to add the tests that you wrote and add an issue for the missing ones?

MagdaN:
@fuzzylogic2000 yes, we could! I had a very strange problem here: I could not add the group subfactory to the plans factory. Well I did add it but when I did something like `pans_factory(group=group)` the group was not added.

I was thinking that my test db did not migrate properly, so maybe we could try again with your db?

PR #2134, opened 2019-03-27
Webpack optimizations

treba123:


treba123:
Travis is too slow today, merging

PR #2135, opened 2019-03-27
change wording in form and information tab

### LINKED ISSUE #2101: testing 1822 show responsible administration
I tested story #1822 and here is what I've noticed:

* please add helptext to "Webseite der Organisation" (as we did last sprint for external projects): "Bitte geben Sie eine vollständige URL ein, die mit https:// or http:// beginnt."
* delete "Stelle und", it should only be Postadresse as the organisation is not mentioned again in this field and the Abteilung should be entered via Ansprechpartner
* @szukowski: do we have or need a design for the headline on mobile? it has the same size as the organisation name (this is also true for the contact person)

nothing bizzare on browserstack :-)

MagdaN:
fixes #2101


fuzzylogic2000:
@MagdaN Did you forget to commit the migrations?

MagdaN:
@fuzzylogic2000 I added the migrations

PR #2136, opened 2019-03-27
display zoom button on the topright corner on desktop

### LINKED ISSUE #2066: zoom button outside view on project map
can we put the zoom button in the top right corner for desktop and tablet and low right for smartphones? client put this on his issue list, medium priority.

@szukowski could you live this this as a flexible zoom button seems to be way more difficult to be implemented?

MagdaN:
fixes #2066


phillimorland:
Works nicely but can get a little confused when swapping between sizes but as this is not a likely behavior of users I think its fine

PR #2153, opened 2019-04-03
Pm 2019 04 container

### LINKED ISSUE #2016: running container isn't shown as running (can't see when filter running is set)

phillimorland:
fixes #2016
Also fixes colour issue on ended banner, was not in issues but was discussed in meeting with Sandy and Caro

### DIFF
--- a/meinberlin/assets/scss/components/_portal_header.scss
+++ b/meinberlin/assets/scss/components/_portal_header.scss
@@ -2,7 +2,7 @@
 
 .portal-header {
     @include clearfix;
-    background-color: #bfd3ff;
+    background-color: $blue-light;
     background-image: linear-gradient(to bottom, #c7d9ff, #b2caff);
     border-top: 5px solid #2d509a;
     min-height: 40px;
--- a/meinberlin/assets/scss/components/_status_bar.scss
+++ b/meinberlin/assets/scss/components/_status_bar.scss
@@ -34,7 +34,7 @@
 .status-bar__past {
     padding: $padding;
     background-color: $blue-light;
-    color: $print-black;
+    color: $text-color;
     text-align: center;
     bottom: 0;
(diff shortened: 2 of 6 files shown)


meinberlin/assets/scss/components/_status_bar.scss
```diff
@@ -33,8 +33,8 @@
 
 .status-bar__past {
     padding: $padding;
-    background-color: $brand-secondary;
-    color: contrast-color($brand-secondary);
+    background-color: $blue-light;
+    color: $print-black;
```
fuzzylogic2000: This is a colour, we use for the printing media-query. Isn't there another black we use everywhere else? Like $text-colour? Alright, I checked that. So, we are sure, that this should use another bg and another text-colour than everywhere else on the site?

fuzzylogic2000: Just say yes, we are very sure and I will merge! :)

phillimorland:
@fuzzylogic2000 it now fails tests?! is this because the test coverage is low for the site?

meinberlin/assets/scss/components/_status_bar.scss
fuzzylogic2000: Cool! Which header? The berlin-one? It should be in variables already then! Maybe bg-secondary or bg-tertiary?

meinberlin/assets/scss/components/_portal_header.scss
```diff
@@ -2,7 +2,7 @@
 
 .portal-header {
     @include clearfix;
-    background-color: #bfd3ff;
+    background-color: $blue-light;
```
fuzzylogic2000: OMG! There was a colour that wasn't a variable! Yeah for fixing that!

PR #2159, opened 2019-04-08
add token everywhere maps are used and allow plans map with image tiles

MagdaN:
depends on https://github.com/liqd/adhocracy4/pull/332

token needs to be added as 'MAP_SERVICE_TOKEN' to local settings


treba123:
The workaround from the berlinonline people might be not that weird. Actually there seems to be two different formats, mapbox and openmaptiles. And AFAIU MapboxGL simply doesn't support the openmaptiles standard and by adding the additional code we can support both. So it seems to be worth the hustle to even add different config variables etc.

See e.g. https://github.com/maputnik/editor/pull/81/files

PR #2181, opened 2019-04-18
use eslint instead of pylint

fuzzylogic2000:


fuzzylogic2000:
@treba123 It uses the same stuff and settings that pylint also uses, but doesn't need pylint. I'd like to have the same linters (and more importantly linter settings) in all our projects! Are we all fine with the settings of mB? @MagdaN @phillimorland In that case, I would also make PRs for the other projects.

PR #2191, opened 2019-04-29
Ka 2019 04 another tiny email fix

fuzzylogic2000:
This current email also works because user has the email as string method, but in this version it's more robust!


fuzzylogic2000:
@treba123 Also, how do I get the commit from my other branch out of this PR?

treba123:
Nice and done

Issue #2228, opened 2019-05-17
participation starting date wrongly displayed on tiles


CarolingerSeilchenspringer:
tiles show a different participation date than the actual one. Tile is one day ahead of date in dashboard:

![Bildschirmfoto 2019-05-17 um 15 21 13](https://user-images.githubusercontent.com/35491681/57930915-76dc2d00-78b7-11e9-900b-d2b060a1c3d0.png)

![Bildschirmfoto 2019-05-17 um 15 19 27](https://user-images.githubusercontent.com/35491681/57930847-48f6e880-78b7-11e9-8d5c-ea34505b01c8.png)

rmader:
Fixed by https://github.com/liqd/a4-meinberlin/pull/2238

PR #2238, opened 2019-05-22
apps: Stop using 'date()'

rmader:
From the commit message:
```
The 'date()' function is not timezone aware, giving us the UTC date
rather than the date of the current timezone. Projects that start
at 0:00 CET will therefore counted as 22:00 of the previous day,
leading to a wrong date.
```


fuzzylogic2000:
@rmader This fixed #2228, right?

rmader:
Right :)

PR #2248, opened 2019-06-03
apps/bplan: if end_date is after now unarchive bplan

MagdaN:


meinberlin/apps/bplan/serializers.py
```diff
@@ -100,13 +101,16 @@ def update(self, instance, validated_data):
         end_date = validated_data.pop('end_date', None)
         if start_date or end_date:
             self._update_phase(instance, start_date, end_date)
+            if end_date > timezone.localtime(timezone.now()):
```
fuzzylogic2000: Maybe this should go to an extra if with '''end_date and end_date>now'''?! It should work like that, because we validate that there are always the two days, but with the extra if it also will not break if we ever decide to change that validation

MagdaN: yes thats true

PR #2310, opened 2019-06-26
 apps/plans: Remember closing the infobox on sticky map 

rmader:
and some ES6 conversion, as that made it easier.

Based on this tutorial: https://medium.com/@rossbulat/using-cookies-in-react-redux-and-react-router-4-f5f6079905dc

Hopefully doesn't break anything jquery related in other places...

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,10 @@
 {
+  "env": {
+    "browser": true,
+    "commonjs": true,
+    "es6": true,
+    "jquery": true
+  },
   "extends": [
     "standard",
     "standard-jsx",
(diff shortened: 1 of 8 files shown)


rmader:
The meat of this MR is actually the changes in `meinberlin/apps/plans/assets/PlansMap.jsx`, so that's what really needs review.

fuzzylogic2000:
I get the same error. 
And was wondering: I guess, we need to thoroughly test it in all browsers (IE11!)?! Did you already do that?

rmader:
Ah right, should work now and the solution is now even cleaner. Will test with IE11 tomorrow, but I'd assume react-cookie works there, too.

rmader:
Just deployed this branch on dev and tested with browserstack. Works on IE11 and safari :)
Any concerns left or can we merge this? :) @MagdaN @fuzzylogic2000

PR #2316, opened 2019-06-27
Mn 2019 06 add modules detail views

MagdaN:
still styling of module detail needed

### DIFF
--- a/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
+++ b/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
@@ -164,6 +164,7 @@
             role="tabpanel"
             aria-labelledby="tab-project-{{ project.pk }}-participation"
             aria-expanded="true">
+
             <div class="l-wrapper">
                 {% if participation_dates|length > 1 %}
                 <div class="timeline-carousel">
@@ -177,7 +178,7 @@
                     <h2>{{ event.name }}</h2>
                     <P>{{ event.description | safe}}</P>
                 {% endif %}
-                {% if modules %}
+                {% if modules|length > 1 or view.events %}
                     {% for module in modules %}
                         <div>
                             <h3>{{ module.name }}</h3>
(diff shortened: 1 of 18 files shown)


meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -187,30 +189,33 @@ <h2>{{ event.name }}</h2>
...
-
-                <div class="l-center-6">
-                    {% block project_action %}{% endblock %}
+                {% else %}
+                    {% if not module.active_phase %}
+                    <div>
+                        <strong>
+                            {% trans 'Participation is not possible at the moment.' %}
+                            {% if not module.past_phases %}
+                                {% html_date module.future_phases.first.start_date 'DATETIME_FORMAT' as start_date %}
```
fuzzylogic2000: I think module.module_start should already work here

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -187,30 +189,33 @@ <h2>{{ event.name }}</h2>
...
+                                {% blocktrans with date=start_date %}It starts on {{ date }}.{% endblocktrans%}
+                            {% elif module.future_phases %}
+                                {% html_date module.future_phases.first.start_date 'DATETIME_FORMAT' as start_date %}
+                                {% blocktrans with date=start_date %}It continues on {{ date }}.{% endblocktrans%}
+                            {% else %}
+                                {% html_date module.past_phases.last.end_date 'DATETIME_FORMAT' as end_date %}
```
fuzzylogic2000: And module.module_end

meinberlin/apps/projects/views.py
```diff
@@ -378,7 +378,8 @@ def project(self):
 
     @cached_property
     def module(self):
-        return self.project.last_active_module
+        if self.modules.count() == 1 and not self.events:
+            return self.modules.first()
```
fuzzylogic2000: Else it returns None, right?

fuzzylogic2000:
Hmm, I think there is a problem with past modules. They are shown as "starting on None". But I will merge nevertheless because it's mainly fine and then we will fix all the stuff as soon as we have everything together.

PR #2335, opened 2019-07-02
module detail: styling module detail with reuseable classes

phillimorland:


meinberlin/assets/scss/components/_button.scss
```diff
@@ -51,6 +51,8 @@
 }
 
 .btn--light {
+    font-weight: inherit;
```
phillimorland: @fuzzylogic2000 is this really bad? we just seem to overwrite the bold font weight or always use btn light with small that I didn't want to make a new class name for it

meinberlin/assets/scss/components/_project_header.scss
```diff
@@ -2,7 +2,7 @@
     display: flex;
     flex-direction: column;
     justify-content: center;
-    padding: 1.3em 0;
+    padding: 1.3em inherit;
```
phillimorland: this keeps project heading inline with other content, incl header

meinberlin/assets/scss/components/_button.scss
fuzzylogic2000: I think it's fine!

PR #2340, opened 2019-07-03
apps: Use is_authenticated instead of is_authenticated()

rmader:
As of Django 2.2, it's not callable anymore but a bool.

Forgot that during the update (was done in AE). Found when debugging https://github.com/liqd/a4-cookiecutter/pull/90


rmader:
This is a bug and uncontroversial, therefore merged

PR #2343, opened 2019-07-04
Adding style guide to comp lib

phillimorland:


meinberlin/assets/scss/components/_component_library.scss
```diff
@@ -2,3 +2,144 @@
...
+    font-size: $font-size-lg;
+}
+
+.fs-sm {
+    font-size: $font-size-sm;
+}
+
+.fs-xs {
+    font-size: $font-size-xs;
+}
```
phillimorland: not sure if this stuff should be added as utilities but it seems excessive?

fuzzylogic2000: Hmm, as long as we don't need it, we could keep it here. But everything here should start with cl-. Anyway, let's keep it like that and see what and if to change if we need it elsewhere!

Issue #2121, opened 2019-03-25
missing translation in comments


CarolingerSeilchenspringer:
this is only true for having just one comment on a comment, having two or more translation is correct.

![Bildschirmfoto 2019-03-25 um 09 59 32](https://user-images.githubusercontent.com/35491681/54906996-d257f300-4ee4-11e9-8e76-cce8d63a4c9c.png)

![Bildschirmfoto 2019-03-25 um 09 59 49](https://user-images.githubusercontent.com/35491681/54906997-d257f300-4ee4-11e9-9dee-bed41a141e5a.png)

phillimorland:
@MagdaN weird double entry in translation file
#| msgid "view one reply"
msgid "view one reply"
msgstr "view one reply"

CarolingerSeilchenspringer:
I reopen the issue because I found a missing translation here https://mein.berlin.de/projects/beteiligungsverfahren-namensgebung-wuhlheide/. I guess we didn't check comments in polls?

phillimorland:
@fuzzylogic2000 @rmader This is weird, we assumed the problem was last time that i had run make po without changing my sed to gsed but this problem is still there and I cannot find a reason for it, the translations are there, the file path is correct, I can't find another case of this wording being used?! i'm confused!

fuzzylogic2000:
@phillimorland 
I don't think it's a missing translation. Maybe the nget-stuff doesn't work? 
And it's everywher (not only polls) and also on dev.

phillimorland:
@fuzzylogic2000 as it's from adhocracy, its also on Opin and on a+ but only for 'hide one answer' not for 'show one answer'?!

rmader:
IIRC if there's a line saying `#, fuzzy, javascript-format` that can break the translation. Will check

PR #2491, opened 2019-09-04
Translation fixes

rmader:
The 'fuzzy' keywords got added when there were other parts using the
same wording not with ngettext. For some reason it now breaks it.

Fixes https://github.com/liqd/a4-meinberlin/issues/2121


Makefile
```diff
@@ -110,6 +110,10 @@ po:
 	msgen locale/en_GB/LC_MESSAGES/django.po -o locale/en_GB/LC_MESSAGES/django.po
 	msgen locale/en_GB/LC_MESSAGES/djangojs.po -o locale/en_GB/LC_MESSAGES/djangojs.po
 
+.PHONY: mo
+mo:
+	$(VIRTUAL_ENV)/bin/python manage.py compilemessages
+
 .PHONY: compilemessages
```
fuzzylogic2000: can't we remove this then? I don't think we use it in any deployment script?

rmader: Hm might be. Will check and remove accordingly.

PR #2492, opened 2019-09-04
apps/notifications: add notification for moderator feedback

fuzzylogic2000:
The linter is broken, but I#ve seen that @rmader checked it in some other project. Will check that!


rmader:
Well, at the point the linter failed it was already done with the python code, only having js/css files left, both not being touched here.

Without really testing: LGTM!

PR #2542, opened 2019-09-19
WIP: Ka 2019 09 fix future polls

### LINKED ISSUE #2513: 403 for poll in future
if a poll starts in the future, click on the tile tells me to sign in, if signed it I get a 403. If I put the date into past, it works. same for projects with and without timeline.

fuzzylogic2000:
fixes #2513 

and makes use of a slightly updated Mixin in A4. This still needs to be updated once (and if) that is merged. Depending on https://github.com/liqd/adhocracy4/pull/401


fuzzylogic2000:
Well, the rules made sense for the preview... I will have to have another think about this.

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -221,12 +221,14 @@ <h2>{{ event.name }}</h2>
             {% endif %}
         </div>
         <!-- these blocks are only filled when in phase view -->
+        {% if module.active_phase or module.past_phases %}
```
fuzzylogic2000: This needs to go with our new plan! :) @MagdaN

PR #2597, opened 2019-10-25
Ka 2019 10 dashboard project create

fuzzylogic2000:
depends on: https://github.com/liqd/adhocracy4/pull/411

still missing:
- only publish when all required fields are filled in
- all the module-logic to only show and take into account published modules
- styling

### DIFF
--- a/meinberlin/apps/dashboard/views.py
+++ b/meinberlin/apps/dashboard/views.py
@@ -1,8 +1,8 @@
-
-
 from django.apps import apps
+from django.contrib.messages.views import SuccessMessageMixin
 from django.http import HttpResponseRedirect
 from django.urls import reverse
+from django.utils.translation import ugettext_lazy as _
 from django.views import generic
 from django.views.generic.detail import SingleObjectMixin
 
@@ -10,6 +10,7 @@ from adhocracy4.dashboard import mixins
 from adhocracy4.dashboard import signals
 from adhocracy4.dashboard import views as a4dashboard_views
 from adhocracy4.dashboard.blueprints import get_blueprints
+from adhocracy4.dashboard.forms import ProjectCreateForm
 from adhocracy4.modules import models as module_models
 from adhocracy4.phases import models as phase_models
 from adhocracy4.projects import models as project_models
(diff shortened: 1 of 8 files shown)


meinberlin/apps/dashboard/views.py
```diff
@@ -89,6 +91,86 @@ def get_permission_object(self):
...
+        if 'referrer' in self.request.POST:
+            return self.request.POST['referrer']
+        elif 'HTTP_REFERER' in self.request.META:
+            return self.request.META['HTTP_REFERER']
+
+        return reverse('a4dashboard:project-edit', kwargs={
+            'project_slug': self.project.slug
+        })
+
+    def publish_module(self):
```
MagdaN: the publish and unpublish function are pretty similar, maybe they can be combined somehow

fuzzylogic2000: Yeah, the function could know, which it gets?! But I think, it wouldn't be more readable, when we then add three more if-statements and only have two lines out of if-statements.

MagdaN: fair enough

meinberlin/apps/dashboard/views.py
```diff
@@ -49,8 +52,9 @@ def post(self, request, *args, **kwargs):
         project = self.get_object()
         module = module_models.Module(
             name=self.blueprint.title,
-            weight=len(project.modules) + 1,
+            weight=max([module.weight for module in project.modules]) + 1,
```
MagdaN: could you also work with a value_list here? As there won't be a lot of modules, using a for loop should not be a problem, but if possible, I guess, it is a bit better to try to use DB functionalities.

meinberlin/apps/dashboard/views.py
```diff
@@ -88,6 +92,119 @@ def get_permission_object(self):
...
+        kwargs['creator'] = self.request.user
+        return kwargs
+
+    def get_success_url(self):
+        return reverse('a4dashboard:project-edit',
+                       kwargs={'project_slug': self.object.slug})
+
+    def form_valid(self, form):
+        response = super().form_valid(form)
+        signals.project_created.send(sender=None,
```
MagdaN: Can we be always sure at this point that the creation was successful?

PR #2644, opened 2019-11-19
apps/users: Update registration form code

rmader:
Use simple overwrites of the allauth forms, like we do on a+. Also,
activate the social registration form for the servicekonto, to mirror
what we do on a+ for github, twitter etc. This potentially fixes some
outstanding bugs and lowers the risk of self-introduced bugs. Follow
up commits might be needed after some testing.


rmader:
This is good to go and an improvement over the current situation. Unfortunately it does not yet solve the 'connect existing account to service-konte' issue. But that can be done in a follow up.

Issue #2030, opened 2019-02-14
project overview is slow


rmader:
To many sql queries...

fuzzylogic2000:
But @MagdaN made it a a lot faster! So, can we close this issue?

treba123:
Jup :)

CarolingerSeilchenspringer:
@rmader @MagdaN @fuzzylogic2000: the project overview is still (or again?) quite slow so I reopen this issue. You have any idea how to fix this?

rmader:
Ok so there's more to this. Apparently some projects have somewhat "broken" data, slowing us down tremendously. #2646 makes this much easier to debug, but does not solve the issue by itself. 

Here is what I got so far. With #2646 applied, there are two API calls that heavily stand out:
 - `/api/containers/`
 - `/api/extprojects/`

---

`/api/containers/` is slow because of `/projects/radverkehrsversuch-tempelhofer-damm/`, which for some reason renders a `point` several thousand chars long. Interestingly this is not heavy on the server, but on the browser. Both Firefox and Chrome spend a long time with maxed out CPU during rendering this, but django debug bar does not report it as being slow:

```
"point": "\"\\\"\\\\\\\"\\\\\\\\\\\\\\\"\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\"\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\ ....
```

Deactivating this project makes the api as fast as expected.

---

`/api/extprojects/` is much worse, but I haven't figured out what exactly causes it. It alone creates 2808 sql queries on the production database and rendering it in the browser took my mashine five minutes (just loading it on `/projekte` is faster, but still the main bottleneck. To be continued.

rmader:
So removing the faulty point from `https://mein.berlin.de/projects/radverkehrsversuch-tempelhofer-damm/` brought us back to acceptable / expected speed. Further more, google page speed does accept our requests again, it simply aborted previously (and rightly so).

Still the projekte site is pretty bad: https://developers.google.com/speed/pagespeed/insights/?hl=de&url=http%3A%2F%2Fmein.berlin.de%2Fprojekte

rmader:
[...]

PR #2646, opened 2019-11-20
Mn 2019 11 load projects asynchronly

MagdaN:
@rmader this is a first draft, just moving all code for fetching project data to an api. 

Helps with https://github.com/liqd/a4-meinberlin/issues/2030


rmader:
I just tested this on dev and it seems to reduce the time to first paint quite a bit. Also gives us some points on pagespeed. But the best part is imo the fact that we can query the api and therefore debug stuff better. So with the tests adapted and a loading animation added: awesome!

MagdaN:
ok, I fixed the tests, now still missing:

- [ ] private projects for user

- [ ] spinner until projects are loaded

PR #2648, opened 2019-11-21
settings: Disable SESSION_COOKIE_SAMESITE to allow servicekonto connections

rmader:
It was introduced in django 2 and by default blocks cookies being send on
cross origin request. Unfortunately the servicekonto account connection,
when a user already has an account and is logged in and wants to connect
her account to the servicekonto, requires us to send the cookie on login
callback.

Note: This is not necessary for oauth2 and other more sophisticated
social login techniques. Further more the allauth `draugiem` provider, on
which the `servicekonto` provider is based on, is the only provider in allauth
requiring a `@csrf_exempt`. Using OAuth2 would be much saver and easier.


rmader:
For the record: this setting is site-wide, but it should not expose any security risks on mB as we use csrf tokens basically everywhere. That is the more general and sophisticated method to defend from cross-site attacks.

PR #2651, opened 2019-11-21
templates/sign_up+registration: add click tracking to login and signu…

phillimorland:
…p with if to check if its being tracked  tracking code
This is the redo of the other tracking pull request which i fucked up with a rebase and an attempt at editing commits so decided to start clean 
old pull request #2095


phillimorland:
@MagdaN This is from a while ago but I finally made the changes you suggested but put them on a new branch as made a mess with the rebase, does this look about right to you?

PR #2667, opened 2019-11-26
apps/plans: Move displaying of image to extra component that loads im…

MagdaN:
…age only when component is rendered

I don't know of this will help.


MagdaN:
@rmader now the images are only loaded if the component is visible. Should we test it?

PR #2704, opened 2019-12-04
Pm 2019 12 a11y map switch

phillimorland:


phillimorland:
@fuzzylogic2000 that was way simpler then I was making it and I have tested with mouse and not just screen reader this time, should all be working

PR #2708, opened 2019-12-04
Story #2394 issue fixes 

phillimorland:


meinberlin/assets/scss/components/_button.scss
```diff
@@ -146,7 +146,7 @@
 .btn--full {
     display: block;
     width: 100%;
-    margin-bottom: $r-spacer;
+    // margin-bottom: $r-spacer;
```
fuzzylogic2000: Did you check, what it does to the other uses of the wide button? If so, can this line be deleted?

meinberlin/assets/scss/_layout.scss
```diff
@@ -66,7 +66,7 @@
     }
 
     .l-tiles-3 {
-        @include grid-tiles(3);
+        @include grid-tiles(2);
```
fuzzylogic2000: I guess this also helps for the other cases where we use the tiles!

meinberlin/assets/scss/components/_button.scss
phillimorland: only in component library but no buttons should have a margin, i'm not sure why this one does?! I thought testing as we go with that one

meinberlin/assets/scss/_layout.scss
phillimorland: we don't use it

meinberlin/assets/scss/components/_button.scss
fuzzylogic2000: Oh, sorry, cache!

meinberlin/assets/scss/_base.scss
```diff
@@ -107,6 +107,11 @@ img {
     }
 }
 
+//stop iphone font scaling for icons
+i {
+    -webkit-text-size-adjust: 100%;
+}
+
```
phillimorland: this might not work but browserstack local isn't working for me today and I can't reproduce anywhere else so if we can just merge and I can check it?

fuzzylogic2000: This is for the bins that are too small? Yes, let's try!

PR #2777, opened 2020-01-14
clean up js

MagdaN:


rmader:
Is this still WIP?

MagdaN:
yes

meinberlin/apps/maps/assets/map_choose_polygon_with_preset.js
```diff
@@ -3,7 +3,7 @@
 /* global django */
 import { createMap } from 'a4maps_common'
 import 'leaflet-draw'
-import '../../../assets/js/i18n-leaflet-draw'
+import '../assets/i18n-leaflet-draw'
```
rmader: Could be `'./i18n-leaflet-draw'`

meinberlin/templates/a4dashboard/base_dashboard_project.html
```diff
@@ -187,8 +187,3 @@ <h2 class="modal-title u-first-heading" id="delete-modal-label">{% trans "Are yo
         </div>
     </div>
 {% endblock %}
-
-
-{% block extra_js %}
-    <script src="{% static 'js/unload_warning.js' %}"></script>
```
rmader: While I do agree with this change here, it's a bit odd to load one big monolithic js file everywhere. But for our current webpack setup it's probably the best approach :/

rmader: Some for the ajax modal btw.

meinberlin/apps/maps/assets/map_choose_polygon_with_preset.js
MagdaN: true

meinberlin/templates/a4dashboard/base_dashboard_project.html
MagdaN: I agree, but we should come up with a generell rule (maybe we could have webpack bundles for each app) here, and the unload_warning actually makes sense for all forms I would say the ajax_modal not so much.

PR #2785, opened 2020-01-16
apps/embed: Move more js into app folder

rmader:


rmader:
I take the freedom to just merge -> trivial.

PR #2786, opened 2020-01-16
more js cleanup

rmader:
This reduces the loaded js size a bit and fixes some bugs (and hopefully does not introduce new ones), including:
 * login via model when embedded did not have a next parameter, so popup would not close
 * select2 did not work (broken container project selection)
 * some css behaving weirdly


rmader:
No hurry here :) Would be great if you could have a look. I'd be especially interested if the issues you find are pre-existing bugs that just happened to work by accident or if the changes here broke something for real.

Issue #2823, opened 2020-02-25
Export of polygon broken


CarolingerSeilchenspringer:
**URL:** https://mein.berlin.de/dashboard/modules/kiezkasse-11/area_settings/
**user:** initiator
**expected behaviour:** I can export my polygon
**behaviour:** if I hit the export button nothing happens.
**important screensize:**
**device & browser:** testet chrome and firefox on mac
**Comment/Question:**

rmader:
Should be similar to https://github.com/liqd/adhocracy-plus/pull/228, will need to port that over.

See https://github.com/liqd/adhocracy-plus/pull/228/files#diff-41cc48c025952a18eef09ca0430ab179L5

PR #2839, opened 2020-03-18
apps/maps/assets: Fix import of FileSaver

rmader:
Fixes https://github.com/liqd/a4-meinberlin/issues/2823


rmader:
Trivial typo, therefore going with it

PR #2840, opened 2020-03-19
Ks 2020 03 update emails

Rineee:


meinberlin/templates/email_base.html
```diff
@@ -56,15 +60,26 @@ <h3 style="color: #818181;">
...
-                            <a href="{{ email.get_host }}/terms-of-use">{% trans 'Terms of use' %}</a>
-                        </p>
+                    <td style="padding-top: 15px;">
+                        <table width="100%" border="0" cellpadding="0" cellspacing="0" style="border-collapse: collapse;">
+                            <tr>
+                                {% if organisation.logo %}
+                                    <td style="width: 1%; padding-right: 10px">
+                                        <img width="50" src="cid:logo" style="width: 50px;" alt="MeinBerlin" >
+                                    </td>
+                                {% endif %}
```
Rineee: this also comes from newsletter_base -> maybe do this in every email?

fuzzylogic2000: Does this show the mB-logo in the footer, when an organisation logo is displayed on top? Then I think, that's a good idea!

Rineee: Yes, thats exactly what it does!

meinberlin/templates/a4reports/emails/report_moderators.en.email
```diff
@@ -1,2 +1,31 @@
 {# Hack to ensure german mails are sent regardless of the current users browser language setting #}
 {% extends 'a4reports/emails/report_moderators.de.email' %}
+
+{% comment %}
```
fuzzylogic2000: You added it as a comment, so that we have the text, because it was translated?

Rineee: Well it was only the 2 upper lines before, which I removed and filled the email and then realized that it is actually sent instead of the german one (which is the reason for the hack obviously, haha). And then I just kept what I did as a comment, because it was already there.. should I delete it?

fuzzylogic2000: No, leave it. Good idea!

PR #2841, opened 2020-03-19
#2769 templates/progress_line: adding progress line with some styling and s…

phillimorland:
…ome if statements
@fuzzylogic2000 this is what I got so far

### DIFF
--- a/meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
+++ b/meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
@@ -57,3 +57,11 @@ def get_num_entries(module):
         + Comment.objects.filter(poll__module=module).count() \
         + Vote.objects.filter(choice__question__poll__module=module).count()
     return item_count
+
+
+@register.filter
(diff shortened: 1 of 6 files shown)


fuzzylogic2000:
Nice! Do you want to style more and then we do the other stuff or should I already have a go?
Oh, I know, I will just branch of your branch and we might have to learn more git in the end. :) I mean, we will have to tackle the rebase! Most exciting thing for that day then

phillimorland:
@fuzzylogic2000 I was waiting for a design on the published project but Caro's going to send me something but I changed the dots a bit as the first one is always completed as the creation has happened so added that and swapped the line and added the 3 dot types. (hope thats what you meant!)

fuzzylogic2000:
@phillimorland I added all the logic, but our dashboard gets more and more crazy. If this works, I still think you should merge for us to test it and talk about it!

phillimorland:
@fuzzylogic2000 i'm not sure on the active behavior but it's working :)

PR #2859, opened 2020-03-25
Pm 2020 03 dashboard styling

phillimorland:
@fuzzylogic2000 ok so most of my styling stuff i scrapped anyway so figured would just put it all in same pull request, it's working and not making loads of spans so yey?!

### DIFF
--- a/meinberlin/templates/a4dashboard/includes/progress_line_in_progress.html
+++ b/meinberlin/templates/a4dashboard/includes/progress_line_in_progress.html
@@ -13,50 +13,64 @@
         <span>{% trans 'Create project' %}</span>
         {% endif %}
     </li>
+
+
     <li class="progress-line__entry">
         <span class="progress-line__line"></span>
         {% if project_progress.project_is_complete %}
             <i class="fas fa-check-circle u-success progress-line__icon" title="{% trans 'Project ready for publication' %}" aria-label="{% trans 'Project ready for publication' %}"></i>
         {% else %}
-            <i class="fas fa-circle progress-line__icon" title="{% trans 'Project not ready for publication' %}" aria-label="{% trans 'Project not ready for publication' %}"></i>
+            <i class="fas fa-circle progress-line__icon--active" title="{% trans 'Project not ready for publication' %}" aria-label="{% trans 'Project not ready for publication' %}"></i>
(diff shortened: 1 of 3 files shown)


meinberlin/templates/a4dashboard/includes/progress_line_in_progress.html
```diff
@@ -69,7 +83,7 @@
         {% if project.is_draft and project_progress.valid != project_progress.required or project.published_modules.count == 0 %}
```
phillimorland: I tried to correct this one but my brain is sad, for projects, this doesn't become filled dot when a module is added to project, i'll have another look in morning

meinberlin/templates/a4dashboard/includes/progress_line_in_progress.html
```diff
@@ -13,50 +13,64 @@
...
-        {% elif project.project_type == 'meinberlin_projectcontainers.ProjectContainer' %}
-        <span>{% trans 'Fill in container details' %}</span>
-        {% else %}
-        <span>{% trans 'Fill in project details' %}</span>
+            <i class="fas fa-circle progress-line__icon--active" title="{% trans 'Project not ready for publication' %}" aria-label="{% trans 'Project not ready for publication' %}"></i>
         {% endif %}
+
+
+        <span class="{% for item in dashboard_menu.project %}{{ item.is_active|yesno:"progress-line__entry--active," }}{% endfor %}">
+            {% if project.project_type == 'meinberlin_bplan.Bplan' %}
```
phillimorland: this is how I dealt with it?!

fuzzylogic2000: I think that's a good idea! Will put it into a templatetag with the cleanup, but merge now!

meinberlin/templates/a4dashboard/includes/progress_line_in_progress.html
fuzzylogic2000: But this should only become filled when it's possible to publish the project? Oh, yeah, this is not working when a module is created, but not filled in, even if another module was added to the project. It's the same for the publish button, I think. Will ignore for now and have a think later!

PR #2862, opened 2020-03-26
Pm 2020 03 aria labels

phillimorland:


locale/de_DE/LC_MESSAGES/django.po
```diff
@@ -3984,13 +4117,16 @@ msgstr ""
...
-"\">register</a> first."
+msgid "If you have not created an account yet, then please "
+msgstr ""
+
+#: meinberlin/templates/account/login.html:9
+msgid "register"
+msgstr ""
+
+#: meinberlin/templates/account/login.html:9
+msgid " first."
```
rmader: The seem to be some translations missing. Mind adding them in this PR so we don't forget?

rmader: If that's planned for later - never mind.

fuzzylogic2000: Öhm, there's something else that went wrong here. The string in login and register is formatted quite strangely. Will check!

PR #2868, opened 2020-03-30
apps/users: add receiver to send email on email_cofirm

Rineee:
Adds welcoming mail that is sent once user confirmed her/his e-mail (allauth signal email_confirm)

### DIFF
--- a/meinberlin/apps/users/emails.py
+++ b/meinberlin/apps/users/emails.py
@@ -1,0 +1,17 @@
+from django.conf import settings
+
+from adhocracy4.emails.mixins import SyncEmailMixin
+from meinberlin.apps.contrib.emails import Email
+
+
+class WelcomeEmail(SyncEmailMixin, Email):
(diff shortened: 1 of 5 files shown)


meinberlin/apps/users/emails.py
```diff
@@ -0,0 +1,17 @@
+from django.conf import settings
+
+from adhocracy4.emails.mixins import SyncEmailMixin
+from meinberlin.apps.contrib.emails import Email
+
+
+class WelcomeEmail(SyncEmailMixin, Email):
```
fuzzylogic2000: @Rineee As this worked and you also already tried and found it, I will merge! It might work now, because you imported the mixin after the Email and it should be the other way round?!
https://stackoverflow.com/questions/10018757/how-does-the-order-of-mixins-affect-the-derived-class
https://www.reddit.com/r/django/comments/5hsl4n/how_the_order_of_mixin_inheritance_works/

PR #2872, opened 2020-03-30
Ks 2020 03 change date format in mails

### LINKED ISSUE #2865: testing 2770 redesign mails
**URL:** mail
**user:** moderator/ user
**expected behaviour:** date and time in the mail I get are displayed e.g. "30. März 2020, 07:34 Uhr"
**behaviour:** they are displayed "30. März 2020 07:34"
**important screensize:**
**device & browser:** 
**Comment/Question:** If the comma is not possible, please add at least "Uhr". This is what we also did in the follow-mails we re-did before. The report mail is also missing a "." behind the "Uhr".

Screenshot?

Rineee:
fixes #2865

### DIFF
--- a/meinberlin/templates/a4reports/emails/report_moderators.de.email
+++ b/meinberlin/templates/a4reports/emails/report_moderators.de.email
@@ -13,7 +13,7 @@
 in dem von Ihnen moderierten Projekt wurde ein Beitrag als unangemessen gemeldet.
 
 Gemeldeter Beitrag: <i style="color: #818181;">„{{ report.content_object }}“</i>,
-erstellt von {{ report.content_object.creator.username }} am {% language 'de' %} {{ report.content_object.created }} {% endlanguage %}
+erstellt von {{ report.content_object.creator.username }} am {% language 'de' %} {{ report.content_object.created|date:"d. F Y, H:i"}} Uhr.{% endlanguage %}
 
 Gemeldet von {{ report.creator.username }} mit folgender Begründung: <i style="color: #818181;">„{{ report.description }}“</i>
 
--- a/meinberlin/templates/a4reports/emails/report_moderators.en.email
+++ b/meinberlin/templates/a4reports/emails/report_moderators.en.email
@@ -18,7 +18,7 @@
 in the project you moderate, a contribution was reported as inappropriate.
 
 Reported contribution: <i style="color: #818181;">"{{ report.content_object }}"</i>,
-created by {{ report.content_object.creator.username }} on {% language 'de' %} {{ report.content_object.created }} {% endlanguage %}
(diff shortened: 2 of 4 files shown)


meinberlin/templates/a4reports/emails/report_moderators.de.email
```diff
@@ -13,7 +13,7 @@
 in dem von Ihnen moderierten Projekt wurde ein Beitrag als unangemessen gemeldet.
 
 Gemeldeter Beitrag: <i style="color: #818181;">„{{ report.content_object }}“</i>,
-erstellt von {{ report.content_object.creator.username }} am {% language 'de' %} {{ report.content_object.created }} {% endlanguage %}
+erstellt von {{ report.content_object.creator.username }} am {% language 'de' %} {{ report.content_object.created|date:"d. F Y, H:i"}} Uhr.{% endlanguage %}
```
fuzzylogic2000: The language de tag tells the date to display it in the German way. The date-filter, you added gives the date the specified format (you know, you added it because of that). As you don't need both, removing the language tag would be good

meinberlin/templates/a4reports/emails/report_moderators.en.email
```diff
@@ -18,7 +18,7 @@
 in the project you moderate, a contribution was reported as inappropriate.
 
 Reported contribution: <i style="color: #818181;">"{{ report.content_object }}"</i>,
-created by {{ report.content_object.creator.username }} on {% language 'de' %} {{ report.content_object.created }} {% endlanguage %}
+created by {{ report.content_object.creator.username }} on {% language 'en' %} {{ report.content_object.created }}|date:"F d Y, H:i"}} o'clock.{% endlanguage %}
```
fuzzylogic2000: In the English emails, I think it's a good idea to keep the language en tag and remove the specified format, as this would read very weird in English, I think.

meinberlin/templates/a4reports/emails/report_moderators.de.email
fuzzylogic2000: Oh no, there was a weird language-problem with some emails, so not to end up with "23. May 2020, 13:12 Uhr" or something like that, leave the language tag in! Sorry!

meinberlin/templates/a4reports/emails/report_moderators.en.email
fuzzylogic2000: In the English ones, I still think, we should remove it. :)

Rineee: But I adapted the filter to the English format here :) Ok, maybe the time would look weird.. will remove it!

fuzzylogic2000: Haha, so it was already strange before. :)

PR #2877, opened 2020-04-01
apps/bplans: save start and end date to bplan as well as correspondin…

### LINKED ISSUE #2528: bplan template dates saved but not shown in Dashboard
URL: https://mein.berlin.de/dashboard/projects/erweiterung-mauerpark-bebauungsplan-3-64-im-bezirk/bplan/
user: initiator
expected behaviour: date and time that I have entered are still shown after saving form
behaviour: dates are no longer shown after saving, no error message, I can still publish the project and date is shown correctly on project tile
device & browser: Desktop, mac, chrome Version 76.0.3809.132 (Offizieller Build) (64-Bit)
Importance: relevant bug, fix before next release

fuzzylogic2000:
…g phase - fixes #2528

Ok, this fixes it, but the whole issue was introduced when we added a start and end_date to the external project (the bplan inherits from). I don't know why we did that and if it's not a better idea to remove that again and only and always use the dates of the corresponding phases. 
Right, no, we talked about that: https://github.com/liqd/a4-meinberlin/pull/2264
I still think, we might be able to use the dashboard to show and change the phase in the corresponding module instead of in the project. But I am not supersure that's possible and would have to get into the whole concept of the dashboard and the extra projects added in mB again.
@rmader What do you think?

### DIFF
--- a/meinberlin/apps/bplan/serializers.py
+++ b/meinberlin/apps/bplan/serializers.py
@@ -66,8 +66,8 @@ class BplanSerializer(serializers.ModelSerializer):
         orga = orga_model.objects.get(pk=orga_pk)
         validated_data['organisation'] = orga
 
-        start_date = validated_data.pop('start_date')
-        end_date = validated_data.pop('end_date')
+        start_date = validated_data['start_date']
+        end_date = validated_data['end_date']
 
         image_url = validated_data.pop('image_url', None)
         if image_url:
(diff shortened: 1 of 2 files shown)


meinberlin/apps/bplan/serializers.py
```diff
@@ -66,8 +66,8 @@ def create(self, validated_data):
         orga = orga_model.objects.get(pk=orga_pk)
         validated_data['organisation'] = orga
 
-        start_date = validated_data.pop('start_date')
-        end_date = validated_data.pop('end_date')
+        start_date = validated_data['start_date']
+        end_date = validated_data['end_date']
```
rmader, line 70: nitpick: also `.get` to be consistent?

rmader: Wait, there is no `get()` on lists - does it work in the change below? Is `validated_data` a dict there?

rmader: Answer to myself: it is, otherwise `pop(foo, default)` wouldn't work - slightly confusing imho, but never mind :)

Issue #2781, opened 2020-01-14
Map only loads only once on embedded page


MagdaN:
**URL:** https://mein.berlin.de/embed/projects/burgerhaushalt-treptow-kopenick/
**user:** any user
**expected behaviour:** when I click a link in a popup of a marker, I can see more detail of the idea. If I click on 'zurück' in the top left corner, I can see the map again with all ideas
**behaviour:** when I click a link in a popup of a marker, I can see more detail of the idea. If I click on 'zurück' in the top left corner, the map does not load again
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
![Bildschirmfoto 2020-01-14 um 17 43 09](https://user-images.githubusercontent.com/2905743/72363572-5f2ea000-36f5-11ea-8f0d-876ce9d10ea7.png)

phillimorland:
when testing also can't seem to add ideas via embed https://meinberlin-dev.liqd.net/embed/projects/test-proj/

phillimorland:
@rmader any ideas on this?

rmader:
A rough one - it's about how we initialize our js, which I *think* is on the page load signal. We'd need sure to emit some signal when we change the site in the embed code and listen to it in the initializer.

rmader:
Note to myself: the `a4.embed.ready` is properly emitted (`$(document).trigger('a4.embed.ready')`) and `a4map_display_points.js` does listen to it (`$(document).on('a4.embed.ready', init)`) but for some reason does not fire. To fix the issue here we need to figure out why that is.

rmader:
Note to myself: https://github.com/liqd/a4-meinberlin/pull/2169 was also already an incomplete fix for the same issue

PR #2880, opened 2020-04-02
assets: Replace jQuery 'a4.embed.ready' listeners with vanilla js

rmader:
When using multple webpack endpoints we do not get a global jQuery
object but instead one for each endpoint. These do not see each others
signals, thus do not fire.

Use standard js events instead - they don't suffer from that problem and
we should get rid of jQuery anyway over time.

Also use the chance to clean up the init syntax.

Depends on https://github.com/liqd/adhocracy4/pull/519

Fixes https://github.com/liqd/a4-meinberlin/issues/2781

**Needs to get tested with IE11** :(


rmader:
Will check IE11 on dev then, thx :)

Issue #2813, opened 2020-02-17
MultipleObjectsReturned: get() returned more than one User -- it returned 2!


fuzzylogic2000:
https://sentry.liqd.net/sentry/meinberlin-prod/issues/970/

```
MultipleObjectsReturned: get() returned more than one User -- it returned 2!
(15 additional frame(s) were not displayed)
...
  File "allauth/account/adapter.py", line 489, in authenticate
    user = authenticate(request, **credentials)
  File "django/contrib/auth/__init__.py", line 73, in authenticate
    user = backend.authenticate(request, **credentials)
  File "allauth/account/auth_backends.py", line 24, in authenticate
    ret = self._authenticate_by_username(**credentials)
  File "allauth/account/auth_backends.py", line 40, in _authenticate_by_username
    user = filter_users_by_username(username).get()
  File "django/db/models/query.py", line 412, in get
    (self.model._meta.object_name, num)

Internal Server Error: /accounts/login/
```

rmader:
Did I get this right that this can only happen when the username is changed by our own form in the settings?

fuzzylogic2000:
@rmader Yes, I think if you try to register as an existing user using other cases in his name (for example Admin, when admin already exists), it's not possible. But when you change your name to Admin later, this happens.

fuzzylogic2000:
Oh, duplicate: #1630 ! Will close the other one, though.

PR #2890, opened 2020-04-09
apps/account/forms: Check username for case insensitive uniqueness

rmader:
Fixes https://github.com/liqd/a4-meinberlin/issues/2813 (well, at least prevents new cases to pop up)


fuzzylogic2000:
Ups, but failing test!

rmader:
Yes, I'm still on it - requires some reading from my side :(

rmader:
Went for a simple solution

PR #2905, opened 2020-04-21
Ka 2020 04 unpublish module

fuzzylogic2000:


meinberlin/templates/a4dashboard/includes/nav_modules_item.html
```diff
@@ -46,7 +46,7 @@
                 {% trans 'Show in project' %}
             </button>
         {% else %}
-            <button name="action" value="unpublish" {% if project|has_unpublishable_modules %}disabled{% endif %} class="btn btn--secondary btn--full btn--attached-top u-spacer-bottom-half">
+            <button name="action" value="unpublish" {% if not project|has_unpublishable_modules %}disabled{% endif %} class="btn btn--secondary btn--full btn--attached-top u-spacer-bottom-half">
```
rmader, line 49: He, right :)

meinberlin/apps/dashboard/templatetags/meinberlin_dashboard_tags.py
```diff
@@ -35,7 +35,7 @@ def has_unpublishable_modules(project):
     another module published for (added to) the project.
     """
     return (project.is_draft
-            and project.published_modules.count() <= 1)
+            and project.published_modules.count() > 1)
```
rmader, line 38: I wonder if we should also allow it when `== 0` - e.g. when creating a new project and then having several unpublished modules. Anyway, that's for later.

PR #2920, opened 2020-04-30
assets/js: Move onload_warning out of global assets again

### LINKED ISSUE #2910: warning in project overview after setting filter
**URL:** https://mein.berlin.de/projekte/
**user:** any
**expected behaviour:** if I set a filter via search function and click on a project tile, I am going straight to the project page
**behaviour:** I get a warning first, that my changes will not be saved
**important screensize:**
**device & browser:** I tested chrome, firefox, safari on mac
**Comment/Question:** 

Screenshot?
image

rmader:
This reverts parts of 80fe4e69d996f360025751535af06b40dfbf6d6f as it caused
us to warn e.g. when having filters set in the project overview.

Closes #2910

### DIFF
--- a/meinberlin/assets/js/app.js
+++ b/meinberlin/assets/js/app.js
@@ -32,7 +32,6 @@ require('../../apps/maps/assets/map-address.js')
 require('../../apps/moderatorremark/assets/idea_remarks.js')
 require('../../apps/newsletters/assets/dynamic_fields.js')
 require('../../apps/dashboard/assets/ajax_modal.js')
-require('../../assets/js/unload_warning.js')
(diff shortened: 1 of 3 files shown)


rmader:
> Cool! And fixes the issue, but don't we want to have it in the forms outside the dashboard as well?

I agree that it would be desirable, but that turns out to be a bit more complex :( To have it everywhere where we do want but not annoy us where we don't - we apparently need a more fine grained solution. That should go into a US IMO

fuzzylogic2000:
Yes, you're right!
@CarolingerSeilchenspringer So, if you want to have this warning outside of the dashboard in all the forms (e.g. the idea forms), please add an US.

PR #2923, opened 2020-04-30
settings/renovate: disable automerge for minor versions

fuzzylogic2000:


rmader:
Nice! Maybe we can enable automerge for `patch`, i.e. the last of the three digits.

PR #2943, opened 2020-05-14
Add platform emails app

rmader:
This adds a new add split off from the newsletter one, intended for admin use only. It allows sending to all users and, in contrast to the newsletter app, does not check if the users disabled newsletters in the settings. It's accessible from the user indicator menu and replaces the kinda broken newsletter entry there.

Todo:
- [x] Make sure we don't send to not verified emails (also newsletter)
- [ ] Add / adopt tests
- [x] Remaining wording changes / email design


rmader:
@Rineee: this is mostly finished from my side, so feel free to take over

meinberlin/apps/newsletters/views.py
```diff
@@ -107,10 +105,6 @@ def form_valid(self, form):
                 enabled=True
             ).values_list('creator', flat=True).distinct()
 
-        elif receivers == models.PLATFORM:
```
rmader, line 84: I thought about this, too, but I'm not sure if we really want to delete it. Platform mails are not newsletters, so maybe we still want a way to write a newsletter to all people?

rmader: Oh never mind, didn't see it's just unused code!

fuzzylogic2000:
But the tests are failing! @rmader @Rineee who fixes that?

Rineee:
> But the tests are failing! @rmader @Rineee who fixes that?

I can do it!

Rineee:
Will add platform tests on Friday!

tests/newsletters/test_newsletter_views.py
```diff
@@ -12,12 +13,15 @@
 
 @pytest.mark.django_db
 def test_send_organisation(admin, client, project, user_factory,
-                           follow_factory):
+                           follow_factory, email_address_factory):
     organisation = project.organisation
     user1 = user_factory(get_newsletters=True)
-    user2 = user_factory()
+    user2 = user_factory(get_newsletters=True)
```
fuzzylogic2000, line 19: This was to test if only one email is sent when one user does not want to get newsletters. In the new version it would make sense to have 3 users: one with email verified and get_newsletters, one with get_newsletters, but email not verified and one without get_newsletters, but email verified. Then there should still be one email sent.

tests/newsletters/test_newsletter_views.py
```diff
@@ -12,12 +13,15 @@
...
     user_factory()
-
+    email_address_factory(user=user1, email=user1.email,
+                          primary=True, verified=True)
+    email_address_factory(user=user2, email=user2.email,
+                          primary=True, verified=True)
     follow_models.Follow.objects.all().delete()
     follow_factory(creator=user1, project=project)
     follow_factory(creator=user2, project=project, enabled=False)
```
fuzzylogic2000, line 27: oh right, that's why it's still only one email sent out...
But I still think it's a good idea to test all options why the email is not sent.

tests/newsletters/test_newsletter_views.py
Rineee: I thought this test was to test if only users that follow an organsisation receive the newsletter. That is why I added the get_newsletter=True, bc otherwise it would not have be sent anyway and thus the follow would have not been tested. 

The setting with the 3 users you describe above is tested in test_send_newsletter_platform(). There, user1 wants newsletters and is verified, user2 wants newsletters and is not verified and user3 doesnt want newsletters, but is verified. The email is then only sent to user1 (plus the initiator, moderator and admin, since they are all verified and want newsletters).

Rineee:
> Very nice! I will merge!
> @Rineee But could you check again tomorrow if all options are tested? Also, the test was for the platform-wide newsletter. Is that still working?

I am not sure what you mean by that? There is the test_send_newsletter_platform() which is still working (see my answer above). Is that what you were asking for?

tests/newsletters/test_newsletter_views.py
fuzzylogic2000: Yeah, maybe the two different users are enough. Just thought the more different scenarious the better. :)

tests/newsletters/test_newsletter_views.py
```diff
@@ -243,23 +288,6 @@ def test_skip_inactive(admin, client, project, user_factory, follow_factory):
...
-
-    url = reverse('meinberlin_newsletters:newsletter-create')
-    client.login(username=admin.email, password='password')
-    response = client.get(url)
-    assert response.status_code == 200
-
-    client.login(username=initiator.email, password='password')
-    response = client.get(url)
-    assert response.status_code == 403
-
```
fuzzylogic2000, line 261: I meant this test. But that was only for the weird view from the dropdown...

tests/newsletters/test_newsletter_views.py
fuzzylogic2000: Haha, oh my, being awake when reviewing would certainly help. :)

tests/newsletters/test_newsletter_views.py
Rineee: Ah, I moved this to the platform email tests. Coming soon :)

PR #2954, opened 2020-05-19
apps/bplan: use new sub-headline from email_base and make content lef…

fuzzylogic2000:
…t-aligned

I am not super happy to add yet another content block, but didn't know how to make it left-aligned AND keep the newlines otherwise :woman_shrugging:

### DIFF
--- a/meinberlin/apps/bplan/templates/meinberlin_bplan/emails/submitter_confirmation.de.email
+++ b/meinberlin/apps/bplan/templates/meinberlin_bplan/emails/submitter_confirmation.de.email
@@ -1,11 +1,14 @@
 {% extends 'email_base.'|add:part_type %}
+
 {% load wagtailcore_tags %}
 
 {% block subject %}Ihre Stellungnahme zum Bebauungsplan "{{ project.name }}" von {{ module.active_phase.start_date|date:'d.m.Y' }} - {{ module.active_phase.end_date|date:'d.m.Y' }}{% endblock %}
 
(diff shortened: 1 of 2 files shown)


meinberlin/apps/bplan/templates/meinberlin_bplan/emails/submitter_confirmation.de.email
```diff
@@ -1,11 +1,14 @@
 {% extends 'email_base.'|add:part_type %}
+
 {% load wagtailcore_tags %}
 
 {% block subject %}Ihre Stellungnahme zum Bebauungsplan "{{ project.name }}" von {{ module.active_phase.start_date|date:'d.m.Y' }} - {{ module.active_phase.end_date|date:'d.m.Y' }}{% endblock %}
```
Rineee, line 5: Shouldnt this be vom instead of von?

fuzzylogic2000: Oh! I didn't read this, because it wasn't in the story... Will do it!

fuzzylogic2000: Well, actually I will remove the dates as we also did that in the headline.

Rineee: Thats makes sense, yes!

Issue #2935, opened 2020-05-11
category icons not displayed correctly


lthueer:
**URL:** https://meinberlin-stage.liqd.net/projects/test-17/
**user:** all
**expected behaviour:** sort ideas by categories with different icons (spatial idea challenge & spatial brainstorming
**behaviour:** the icons for different categories are all displayed the same
**important screensize:** /
**device & browser:** /
**Comment/Question:** the screenshot describes the problem. A test project on stage was used for that problem: https://meinberlin-stage.liqd.net/projects/test-17/

Moreover, here is a link to a project how the category icons "should" look like: https://mein.berlin.de/projects/vorbereitende-untersuchungen-nach-ss141-baugesetzb/

![Bildschirmfoto 2020-05-11 um 16 29 11](https://user-images.githubusercontent.com/36957474/81575688-97927200-93a7-11ea-805f-b01e88f6e601.png)

fuzzylogic2000:
When I look at the link it looks like the icons can't be added to the categories in the dashboard. Is that the problem? Can they be changed for projects that were there before?
Is it the same on dev?
@CarolingerSeilchenspringer

CarolingerSeilchenspringer:
@fuzzylogic2000 the icon-dropdown is not shown in dashboard next to the categories, neither on dev nor on prod and neither for new or older project.

fuzzylogic2000:
But on the map they are still shown for older projects? @CarolingerSeilchenspringer ?
Then the problem is in the dashboard! (where we changed a lot, so that makes sense. :smile: )

CarolingerSeilchenspringer:
correctly @fuzzylogic2000 in frontend they are shown for old projects but not for new ones, in Dashboard they can't be set in any project.

fuzzylogic2000:
I suspect that the problem in the dahboard is an unknown webpack endpoint: https://github.com/liqd/adhocracy4/blob/master/adhocracy4/categories/assets/category_formset.js

fuzzylogic2000:
So, this and the select are already in there: https://github.com/liqd/a4-meinberlin/blob/fbf106d0e432c9a57e288bb942cbbe67f1f23747/meinberlin/templates/a4forms/includes/inline_form.html#L28
The select is not shown because of d-none and also shouldn't because it needs to be replaced by a dropdown also showing the icons. That should be done by https://github.com/liqd/adhocracy4/blob/master/adhocracy4/static/select_dropdown.js, but I can't get it to work.
[...]

PR #2957, opened 2020-05-19
Dropdown select fixes

rmader:
Cleanups for our `app.js` and use select2 in more places.

Depends on: https://github.com/liqd/adhocracy4/pull/550
Related: https://github.com/liqd/a4-meinberlin/issues/2935


rmader:
This also means this:
image

becomes this:
image

Or even graphic:
image

PR #2963, opened 2020-05-22
WIP: apps/cms: add captcha to EmailForm

fuzzylogic2000:
I added the captcha like here https://github.com/springload/wagtail-django-recaptcha
and more info https://docs.wagtail.io/en/v2.1.1/reference/contrib/forms/customisation.html

Now it's displayed, but not verified. The mixin didn't have any effect. 
I also (shortly) tried to do the verification in the field, because I couldn't find any other way to get it into the form. But I only had value and not the data there, so didn't succeed.


rmader:
This was merged as part of https://github.com/liqd/a4-meinberlin/pull/2990 (and https://github.com/liqd/a4-meinberlin/pull/3003)

PR #3026, opened 2020-07-06
apps/notifications/send_test_emails: add new email to test emails

phillimorland:
@fuzzylogic2000 this was my just copy over test but I think it currently just checks if there is a bplan, i need to add the parts where it checks if it updates the bplan, also how the does this reduce test coverage?! its a test?! errr


fuzzylogic2000:
@phillimorland This is so cool! It's not for the automated python tests, but can be used for manual testing of the design of (nearly) all emails! So, you found quite a treasure there! Maybe we can check which new emails are missing (like wellcome and offline evnet notification) and/or also do something like this for a+?! @Rineee what do you think?
To use it, you'll have to add a user with your email address to your local DB and then start it with
`python manage.py send_test_emails <your email address>`
It might tell you that stuff is missing from the DB, so you just have to add or do one of these (like one Idea, one Project, one Bplan, one ParticipantInvite, ...)

PR #3033, opened 2020-07-07
WIP bplan/signals: add in update if statement so triggered on draft save

phillimorland:
@fuzzylogic2000 I just added the update if statement back as that was working with the save in dashboard but apparently not with the publication.
I was wondering if we could use anything from the serializer but I don't understand enough to know, I was just reading what python signals are so thats where i'm at


fuzzylogic2000:
The update_fields are a bit weird. One would think, they should be there in the save-signal, but you actually have to manually set them on save.
I tried to explain, where we set the update fields here: https://github.com/liqd/a4-meinberlin/pull/3019/commits/f273c2eca2c53f58e679dfdfc3c7a47019152911

PR #3097, opened 2020-09-03
Ks 2020 09 add interactive event module

Rineee:
The permission issue is fixed now. Because of [this code](https://github.com/liqd/adhocracy4/blob/3eb15d67f4de91aa004671a6813d5a847755d471/adhocracy4/api/permissions.py#L54) it was always asking for view_livequestions, but the permission was called view_questions, so renaming all permission was the solution.

So now you can add an interactive event and view it and post questions, but the whole react part looks very messy and I am not even sure if everything works there.. but maybe this is something for @phillimorland to go on with? :)


Rineee:
The permission issue is fixed now. Because of [this code](https://github.com/liqd/adhocracy4/blob/3eb15d67f4de91aa004671a6813d5a847755d471/adhocracy4/api/permissions.py#L54) it was always asking for view_livequestions, but the permission was called view_questions, so renaming all permission was the solution.

The whole react part looks very messy and I am not even sure if everything works there.. but maybe this is something for @phillimorland? :)

fuzzylogic2000 (changes requested):
The project tile and the view from the dashboard doesn't link to the view: 
```
django.urls.exceptions.NoReverseMatch: Reverse for 'module-detail' with arguments '('liqd', 'interactive-event')' not found. 1 pattern(s) tried: ['projekte/module/(?P<module_slug>[-\\w_]+)/$']
```

fuzzylogic2000 (approved):

fuzzylogic2000:
@phillimorland Alright, do you want to go on in this PR or do we just merge (while it still looks pretty broken)?

phillimorland:
@fuzzylogic2000 @Rineee I have started doing a clean up and it now doesn't look so broken, so we can maybe merge now and then keep working on different parts?

fuzzylogic2000 (approved):
Yes! I like that, let's do that!

PR #3098, opened 2020-09-03
settings/base.py: adding additional plugins to ckeditor and updating …

phillimorland:
…unload filename so doesn't clash with plugins

* The preview is broken after save but the iframe remains so it's 'working' this might be a long standing problem though according to some issues I found
* It also has the capability added to all ckeditors for testing purposes but I think we only want it on the collapsible element one but until we are happy with whats happening it's good to have the tester options.
* The nocookie URL from youtube may or may not solve our problem, I didn't find much on how to add the alert though?!
* And apparently travis is failing?! I have only just notice that

### DIFF
--- a/meinberlin/assets/js/unload_warning.js
+++ b/meinberlin/assets/js/unload_warning.js
@@ -1,7 +1,8 @@
 /* This code checks if something has been changed in a form but not submitted.
    If the user wants to leave the the page there will be warning. */
 
-/* global CKEDITOR django */
+// /* global CKEDITOR */
+/* global django */
 
 $(function () {
   var submitted = false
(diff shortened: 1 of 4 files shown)


fuzzylogic2000:
Cool! But when I fix travis (by also chaning unload_warning to warning_unload in webpack), I get the target not defined and not-stopping loading again. :( Don't know what to do about that! Or maybe, we don't use the warning_unload on the dashboard form pages with ck editor?

fuzzylogic2000:
Right, but we don't use different templates for the different dashboard forms. :(

fuzzylogic2000:
I just updated the ck editor. So before we go on, let's rebase! :)

fuzzylogic2000 (approved):
Very nice! It works!

PR #3107, opened 2020-09-08
WIP update cofig to fix browserlist warning -not working

phillimorland:


fuzzylogic2000:
@phillimorland Do we need it for this release or can we put it in the next one (and do it some time inbetween)?

phillimorland:
@fuzzylogic2000 this is on my list for questions for Robert, it's only a deprecation warning (it seems) so I think it's fine to wait, or I can uninstall the updates that caused it, what do you think?

rmader:
Played a bit with this and concluded it's an upstream bug - I'd say lets just ignore it for now and hope it will get fixed by a future update :shrug:

PR #3114, opened 2020-09-14
apps/livequestion; adding tool tips and updating some styles

phillimorland:
some styling i forgot on thursdays and adding tool tips to btns


fuzzylogic2000 (approved):
Looks good!

PR #3123, opened 2020-09-15
Pm 2020 09 tooltips

phillimorland:


fuzzylogic2000 (approved):
Reads very well again!

PR #3126, opened 2020-09-16
app.js: add class to the iframe when added via ckeditor, update styling

phillimorland:
this will sort the weird nested div stuff, its a bit messy, sorry!


fuzzylogic2000 (approved):
:partying_face:

PR #3128, opened 2020-09-16
offlineevents/models: update form field and templates to use collapsi…

phillimorland:
…ble ckeditor


fuzzylogic2000 (approved):
Whoop whoop!

PR #3133, opened 2020-09-21
dsgvo: add package via webpack and style and only load on pages that …

phillimorland:
…need it


fuzzylogic2000 (approved):
Looks great and works! Let's wait for travis, though.

PR #3150, opened 2020-09-23
livequestions styling issue

phillimorland:
ok there was an unexpected js hole that i fell into but this should be everything :)
Also still want to do a code tidy but feel that would be better done monday morning


fuzzylogic2000 (approved):
Cool!

PR #3152, opened 2020-09-24
apps/bplans: don't trigger get_location or send_update on automatic b…

### LINKED ISSUE #3061: when bplan is archived, office worker update email is sent
**URL:** 
**user:** 
**expected behaviour:** Office workers only get an update email when bplan is "published, changed or created as draft"
**behaviour:** The automatic archiving also changes the bplan and by that triggers the email. Thinking about it, it should also trigger the get-point thing (and might result in failure errors being thrown).
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

fuzzylogic2000:
…plan archiving - fixes #3061

Can be tested by adding a bplan that is already finished and then triggering the auto archive by
`python manage.py bplan_auto_archive`
In this branch there is nothing in the tasks, while on master there are still the ones for get_location and send_update.


Rineee (approved):
Awesome!

PR #3156, opened 2020-09-28
apps/projects: make sure serializers only take phases of published mo…

fuzzylogic2000:
…dules into account

fixes https://sentry.liqd.net/sentry/meinberlin-prod/issues/1207/
This occurs when a published project has a published (added to the project) past module and an unpublished (not added to the project) running module. That project ended up in the wrong serializer then because of the filters not filtering for published modules.

### DIFF
--- a/meinberlin/apps/projects/filters.py
+++ b/meinberlin/apps/projects/filters.py
@@ -12,11 +12,13 @@ class StatusFilter(filters.BaseFilterBackend):
             active_projects = queryset \
                 .filter(
                     module__phase__start_date__lte=now,
-                    module__phase__end_date__gt=now) \
+                    module__phase__end_date__gt=now,
+                    module__is_draft=False) \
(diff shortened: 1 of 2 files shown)


phillimorland (changes requested):
List still isn't loading? error in terminal (i think)
I can also scrap my database and see if it's a different issue causing it this time but will have to be after lunch
[28/Sep/2020 12:33:50] "GET /api/projects/?status=pastParticipation HTTP/1.1" 500 36411

fuzzylogic2000:
@phillimorland Yes, that looks like your pastParticipation list broke now. :( Let's look at that together when you're back!

fuzzylogic2000:
Oh no, and for me the project in question also vanished from the list while it is still there in the API. How weird. :cry:

phillimorland (approved):
working with new database :)

PR #3169, opened 2020-09-30
apps/contrib: Add option to disable Matomo cookies

rmader:
In order to avoid having to show a cookie banner.

---

This is actually for a+, but as meinberlin-dev is already set up on matomo, lets test the code here first.


fuzzylogic2000 (approved):
Looks good! And do you want to put cookie_disabled into the pillar settings then?

fuzzylogic2000 (commented):
Well, could you actually add a test? If this is not [easily] testable for something else. Maybe we manage > 84%.

rmader:
TBH, it doesn't make much sense to me just for the sake of pushing the number up :/
If I write tests, I would rather write them for something we can actually break by accident.

PR #3181, opened 2020-10-12
Pm 2020 10 video format issues

phillimorland:
I don't know is loading the video embed with adhocracy4 is correct but it works as it's loading before any of the map css, I wonder if it would be best to have it own entry point? @fuzzylogic2000 @rmader @Rineee what do you think?


fuzzylogic2000 (approved):
Looks good!
I think it's own entrypoint would be better. But if it doesn't work like that, we have to do it differently. :woman_shrugging:

PR #3182, opened 2020-10-12
editorconfigs: update editor configs to be same as a+ and remove obso…

phillimorland:
…lete eslintignore
i think isort already skips these but this will make it same as a+

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,12 @@
 VIRTUAL_ENV ?= venv
 NODE_BIN = node_modules/.bin
 SOURCE_DIRS = meinberlin tests
+ARGUMENTS=$(filter-out $(firstword $(MAKECMDGOALS)), $(MAKECMDGOALS))
+
+SED = sed
+ifneq (, $(shell command -v gsed))
+	SED = gsed
+endif
 
 .PHONY: all
 all: help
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -88,12 +89,24 @@
...
+  },
+  "lint-staged": {
+    "*.{js,jsx}": [
+      "eslint"
+    ],
+    "*.scss": [
+      "stylelint"
+    ],
+    "*.py": [
+      "make lint-python-files"
```
fuzzylogic2000, line 109: This make-target is missing, so it doesn't work!

phillimorland: oupps missed that, should I also add it to the help section in make file? it's not added in a+

Makefile
```diff
@@ -94,6 +95,20 @@ lint:
...
+lint-quick:
+	EXIT_STATUS=0; \
+	npm run lint-staged ||  EXIT_STATUS=$$?; \
+	$(VIRTUAL_ENV)/bin/python manage.py makemigrations --dry-run --check --noinput || EXIT_STATUS=$$?; \
+	exit $${EXIT_STATUS}
+
+.PHONY: lint-python-files
+lint-python-files:
+	EXIT_STATUS=0; \
+	$(VIRTUAL_ENV)/bin/isort --df -c $(ARGUMENTS) || EXIT_STATUS=$$?; \
```
fuzzylogic2000, line 108: I think the ARGUMENTS are still missing? And yes, I think it makes sense to also keep the help section updated?! Or what do you think?

phillimorland: I didn't add as the make check-migrations wasn't added so assumed only things that needed to be there are commands that might be run by someone, I figured this probably wouldn't be run by someone but will add it anyway. Also thinking about it the make check-migrations doesn't need to be there anymore as that is called in the make lint-quick so will rm.
As for the ARGUMENTS thing, I don't know, I think this is something @rmader added, i'll ask in morning as stuck in style linting in Opin now.

fuzzylogic2000: The Arguments should just be a variable added to the beginning of the file.

fuzzylogic2000 (approved):
Very nice!

PR #3199, opened 2020-10-15
templates: add superblock to all extra js and css blocks to ensure th…

### LINKED ISSUE #3178: video cookie info: layout broken
**URL:** https://meinberlin-dev.liqd.net/projekte/ihre-vision-fur-berlin-welche-themen-sind-ihnen-wi/?initialSlide=4 (click: über das Projekt)
**user:** unregistered user
**expected behaviour:** should look like this: https://meinberlin-dev.liqd.net/projekte/ihre-vision-fur-berlin-welche-themen-sind-ihnen-wi/?initialSlide=2 (click: über das Projekt)
**behaviour:** layout is broken if 1) there is a module with map that is not part of a cluster 2) URL has ending of the slide of that module (if you have a timeline)
**important screensize:**
**device & browser:** 
**Comment/Question:** the only difference of the two above URL is the initialSlide. if it leads to a map, the view is broken in both information and result tab. if it leads to an event or module without map, view of video is fine. It does not have to do with state of publication nor with active/non-active phase. it happens on both chrome and firefox and an dev/stage/prod.

Screenshot?
Bildschirmfoto 2020-10-07 um 16 25 18

phillimorland:
…ey are not overitten by another fixes #3178

### DIFF
--- a/meinberlin/apps/bplan/templates/meinberlin_bplan/bplan_create_dashboard.html
+++ b/meinberlin/apps/bplan/templates/meinberlin_bplan/bplan_create_dashboard.html
@@ -46,8 +46,10 @@
 {% endblock %}
 
 {% block extra_js %}
+    {{ block.super }}
     {{ form.media.js }}
 {% endblock %}
 {% block extra_css %}
+    {{ block.super }}
     {{ form.media.css }}
 {% endblock %}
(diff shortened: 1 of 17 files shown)


fuzzylogic2000 (approved):
Ahahahaha! Awesome!

rmader:
I'm slightly worried that this will trigger styling bugs somewhere now - but not too much :)

It should behave way more predictable now I guess.

phillimorland:
we have very few css extra blocks except forms which is same so should be pretty solid 🤞

PR #3225, opened 2020-11-02
Pm 2020 11 header video embed

phillimorland:


fuzzylogic2000:
@phillimorland It's in the dashboard now. I left it as an extra commit, so that you can see the changes. We can squash the commits later. And I also still need to work on the template (but I think, that shouldn't be too bad. we'll see).

phillimorland:
@fuzzylogic2000 I decided to add it into the module_detail as I was very unhappy with it's placement when added under everything and I think we may get live_stream in at least the debate module, let me know if you disagree and I will talk to Marcelo tomorrow to se how we can get it to work. I will mention in the story that we will leave in the phases until we have decided what we are doing with them.

meinberlin/templates/a4modules/module_detail.html
```diff
@@ -48,6 +48,15 @@
                     </li>
                 </ul>
             </nav>
+            {% for phase in module.phases %}
+              {% if phase == module.active_phase %}
```
fuzzylogic2000, line 52: Yes, that's alright. But then it's also shown above the info on the livequestions page with the cta, isn't it? Or only in timeline view? If it's really only shown where we want it to be, you can make one if from the two: `if phase == module.active_phase and live_stream`

phillimorland: oh cool! thanks! 
I checked on my current project and it wasn't shown but i'll check timeline view

fuzzylogic2000 (approved):
Whoop whoop!

PR #3235, opened 2020-11-09
Pm 2020 11 cookie banner add

phillimorland:


fuzzylogic2000 (approved):
Nice! But could you add some comment, what this is used for? (and maybe also a comment where the the actual livestream is added to say what is added and when?)

phillimorland:
It says in the file name, you think more documentation is needed? also should that be in all templates where it's added?

fuzzylogic2000:
No, I just thought it is a good idea to say in this very long and overly complicated template, what the js is for. For the result or info tab, I think (or hope) it is more straightforward and easier to understand.

phillimorland:
cool, will do! i'll also try add the test in this pull request :D

fuzzylogic2000 (approved):
Yeah!

PR #3245, opened 2020-11-10
tests: add test for livequestion detail view and add livestream facto…

phillimorland:
…ry and register


fuzzylogic2000 (approved):
Supercool!

PR #3250, opened 2020-11-16
Ka 2020 11 livestream helptext

### LINKED ISSUE #3248: testing 3491 (livestream interactive event): can write into livestream ck-editor
**URL:** 
**user:** initiator
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** It is not by any chance possible to not be able to write into the CK-editor?  Initiators can write long texts in there and what can be done probably will be. The questions is, if we want to allow it or not? At least I think we should add a helptext saying that they SHOULD NOT write here. this could be also the place to explain that the video is only visible during active phase. Let's discuss this in comments and depending on what we decide I can think about a helptext.

Screenshot?

fuzzylogic2000:
fixes #3248


Rineee (approved):
Excellent!!

PR #3258, opened 2020-11-18
deps: update a4 and adapt to project access enum

fuzzylogic2000:
What I didn't do here is in the story now. 
So, when this is fine, please merge!

- [x] update a4
- [x] replace is_public in forms
- [x] cheange serialzer
- [x] fix project list
- [x] add new icon for semi-public (to projects list and projecttiles)
- [x] display "join in" or "read" on project tiles
- [x] fix in dashboard
- [ ] check external projects, bplans and containers
- [x] check tests
- [ ] copy tests over
- [x] think about interactive events
- [x] fix sitemap(s)
- [x] add info sentence to project detail when private or semi-private

### DIFF
--- a/meinberlin/apps/cms/blocks.py
+++ b/meinberlin/apps/cms/blocks.py
@@ -3,6 +3,7 @@ from django.utils.functional import cached_property
 from wagtail.core import blocks
 from wagtail.images.blocks import ImageChooserBlock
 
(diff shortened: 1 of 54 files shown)


rmader (approved):
Nice, thanks for structuring the commits so nicely, makes it very good to read. LGTM!

PR #3268, opened 2020-11-25
Ks 2020 11 use react to add livequestions

Rineee:
Todo:

- [x] do api call here

- [x] fix/add tests

Also there are two little styling issues left, maybe @phillimorland can look at it? The style of the select dropdown and there is space missing next to the checkbox.


.eslintrc
```diff
@@ -18,7 +18,8 @@
     "plugin:jsx-a11y/recommended"
   ],
   "rules": {
-    "jsx-quotes": [2, "prefer-double"]
+    "jsx-quotes": [2, "prefer-double"],
+    "jsx-a11y/no-onchange": "off",
```
phillimorland, line 22: I added that as I looked into the onChange thing and it's 3 use cases that it shouldn't be used for, non of which we are likely to do and with onblur it only changed when I clicked into next form field, not after select so not very useful to us

meinberlin/apps/livequestions/assets/QuestionForm.jsx
```diff
@@ -80,8 +80,8 @@ export default class QuestionForm extends React.Component {
             value={this.state.question}
           />
 
-          <div class="form-check">
-            <label class="form-check__label">
+          <div className="form-check">
+            <label className="form-check__label">
```
Rineee, line 84: :see_no_evil:

meinberlin/apps/livequestions/api.py
```diff
@@ -36,3 +39,9 @@ def get_queryset(self):
         ):
             live_questions = live_questions.filter(is_hidden=False)
         return live_questions
+
+    def dispatch(self, request, *args, **kwargs):
```
Rineee, line 43: I am bit unhappy with this, because I have to get the category manually, but didnt know how else to solve it..

meinberlin/apps/livequestions/api.py
```diff
@@ -36,3 +39,9 @@ def get_queryset(self):
...
+        if request.method == 'POST':
+            body = json.loads(request.body.decode("utf-8"))
```
Rineee, line 45: I am bit unhappy with this, because I have to get the category manually, but didnt know how else to solve it..

.eslintrc
Rineee: Cool, thank you for looking into that!

fuzzylogic2000 (approved):
Yeah! Supercool!

meinberlin/apps/livequestions/api.py
fuzzylogic2000: Maybe @rmader has an idea?

rmader: Hm, given that it's not part of the livequestion model...do we by any chance have a mixin for the categories? Other than that I don't have an idea :/

Rineee: No, we don't have a category mixin. I could write one, but maybe thats overkill and we just leave it like that?

fuzzylogic2000: Oh righ, it's ther from the weird a4-categorizable: https://github.com/liqd/a4-meinberlin/blob/166e44d2e947db6caf290afd187c44954f2602cb/meinberlin/config/settings/base.py#L474
So, I would say, leave it like that!

PR #3292, opened 2020-12-02
Pm 2020 12 styling issues

phillimorland:
Ok the last commit adds a scroll to the question however the question is added a bit too slowly, not sure if there's anything we can do about that? 

@Rineee found the js package that skins select elements which I had totally forgotten about :D so should be less stress about that! :)


rmader (approved):
LGTM!

Rineee (approved):
Excellent!!!

PR #3296, opened 2020-12-03
contrib/management: Add command to clean up unverified users

rmader:
Strictly speaking we should check for the email verified status here,
but if a user never logged in that pretty much implies the same.

Prints a list of deleted users, which we'll recieve as mail from
a cron job.

See also https://github.com/liqd/adhocracy-plus/pull/991


fuzzylogic2000 (approved):
Yeah!

PR #3297, opened 2020-12-04
Ks 2020 12 interactive event wording issues

Rineee:


meinberlin/apps/livequestions/assets/QuestionForm.jsx
```diff
@@ -21,17 +21,20 @@ export default class QuestionForm extends React.Component {
     this.setState({ comment: e.target.value, questionCharCount: e.target.value.length })
   }
 
-  getPrivacyPolicyLabel () {
+  getPrivacyPolicyLabelWithLinks () {
+    const splittedLabel = this.props.privatePolicyLabel.split('{}')
```
fuzzylogic2000, line 25: I will merge it like that, but (and maybe for a+?), we don't have these urls in the settings? Then we could also put that into the templatetag.

fuzzylogic2000 (approved):
Yeah!

meinberlin/apps/livequestions/assets/QuestionForm.jsx
Rineee: In Aplus I am taking the urls from the Important Pages in the templatetag, so there it is fine.. but in mB I couldnt really figure out how to get the urls (because here we dont have the Important Pages and they are just some wagtail page, right?) and thus just did like its done in the registrations form, where it is also hard coded. If you have an idea how to get them, please share, I also dont like it this way!

fuzzylogic2000: Oh, then we just don't have them in the wagtail settings in mB and rely on them having always the same url. :(

PR #3313, opened 2020-12-09
livequestions/assets: Restart polling interval when adding question

rmader:
Directly fire a update request and reset the timer, so the own question
is shown faster.

Port of https://github.com/liqd/adhocracy-plus/pull/1017


phillimorland (approved):
super cool! :)

PR #3314, opened 2020-12-09
livequestions/assets: make own component for category select to make …

Rineee:
…it work with select2 package

So the problem I had was when first clicking on the dropdown of the category select, it would immediately select the first option and close again.. it is gone now after restart. This is maybe some js race codition? Do you have the same problem?


phillimorland (approved):
working like a charm on ff chrome and safari on mac, will wait to ensure linux isn't a problem though

rmader (approved):
LGTM and work great here on Linux

PR #3315, opened 2020-12-09
Pm 2020 12 poll issues

phillimorland:


Rineee (approved):
Nice! The tests were broken, but already before and we didnt notice bc we always merged before build was ready since Travis is taking so long .. I opened a PR with test fix, so will merge now anyway!

PR #3318, opened 2020-12-11
templates/a4dashboard/nav_modules_item: use unique id for every delet…

Rineee:
…e modal to actually delete the right module.


fuzzylogic2000 (approved):
Supercool! I guess, we also need to fix that in A+. Before we release... :)

PR #3364, opened 2021-02-04
apps/plans: reorder and add new field to form and detail page and add…

phillimorland:
… to export
haven't updated the export order yet but asked about desired order in the story


fuzzylogic2000 (approved):
Works like a charm! 
I will try to change the order of the export, before I merge, though.

PR #3366, opened 2021-02-08
apps/budgeting//dashboard: add buergerhaushalt with 2 phases to bluep…

Rineee:
…rints and add respective phases to budgeting phases

Since I am always confused with what is used where from phase definitions and blueprint definitions, I made a list with explanations.. but not sure where to put it? Phase docs are in A4, maybe somewhere there? But the usage might be project specific, so maybe write down per project? Or do you think it is not necessary at all?

[phases_blueprints_what_is_what.txt](https://github.com/liqd/a4-meinberlin/files/5943383/phases_blueprints_what_is_what.txt)

Oh and please feel free to add stuff :)

### DIFF
--- a/meinberlin/apps/budgeting/phases.py
+++ b/meinberlin/apps/budgeting/phases.py
@@ -13,8 +13,7 @@ class RequestPhase(phases.PhaseContent):
     view = views.ProposalListView
 
     name = _('Request phase')
-    description = _('Request budgeting and get feedback through rates and '
-                    'comments.')
+    description = _('Participants can submit ideas, comment on and rate them.')
     module_name = _('participatory budgeting')
 
     features = {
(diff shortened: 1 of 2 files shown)


meinberlin/apps/budgeting/phases.py
```diff
@@ -13,8 +13,7 @@ class RequestPhase(phases.PhaseContent):
     view = views.ProposalListView
 
     name = _('Request phase')
-    description = _('Request budgeting and get feedback through rates and '
-                    'comments.')
+    description = _('Participants can submit ideas, comment on and rate them.')
```
fuzzylogic2000, line 16: Are these the new descriptions? 
Maybe in English we should stick to proposal instead of idea, so it might be less confusing?

fuzzylogic2000 (approved):
Nice! I will merge and maybe we can think about the English phase descriptions again?

fuzzylogic2000:
Right! And I forgot about the file. Just put it in the docs as blueprints.md?! I think we will find it there and it might help!

meinberlin/apps/budgeting/phases.py
Rineee: Yes, they are in the story.. @CarolingerSeilchenspringer ok to change ideas to proposasl?

Rineee:
> Right! And I forgot about the file. Just put it in the docs as blueprints.md?! I think we will find it there and it might help!

You mean here in mB, right?

PR #3375, opened 2021-02-09
Pm 2021 02 linting update

phillimorland:
not sure if this should all go in together?
Also haven't tested it all that well as no projects locally yet


fuzzylogic2000:
Isn't this the same as https://github.com/liqd/a4-meinberlin/pull/3222 ?

phillimorland:
@fuzzylogic2000 err the reason for that is long and stupid but short answer is yes, i'll close this and will look at the other one, I have transferred my database now so will be able to test properly now

PR #3378, opened 2021-02-09
renovate.json: allow minor updates to auto merge

phillimorland:
This is ok to do now right?


fuzzylogic2000 (approved):
I think it's fine! Otherwise, we will merge it without really testing (I do that at least.). :)

PR #3380, opened 2021-02-10
assets: js cleanups

rmader:
Instead of relying on webpack to provide the common map file, properly
import it (requires new a4). Also rely less on global jQuery.

Depends on https://github.com/liqd/adhocracy4/pull/670


phillimorland (approved):
Awesome!

PR #3382, opened 2021-02-11
package.json; rm webpack-terser as now included in webpack, webpack.prod: update …

phillimorland:
…schema test
I think we could maybe update this further: 
https://stackoverflow.com/questions/57360588/how-to-use-terser-with-webpack
@rmader what do you think?


webpack.prod.js
```diff
@@ -3,7 +3,6 @@ const { merge } = require('webpack-merge')
 const TerserPlugin = require('terser-webpack-plugin')
 
 module.exports = merge(common, {
-  devtool: 'eval',
```
rmader, line 6: maybe change to `devtool: 'source-map'` instead, see https://webpack.js.org/configuration/devtool/

rmader: Did this now

rmader (approved):
LGTM!

PR #3383, opened 2021-02-11
Ks 2021 02 participatory budgeting 2 phase adjustments

Rineee:


meinberlin/templates/a4modules/module_detail.html
```diff
@@ -91,7 +91,6 @@ <h1 class="project-header__title">{{ module.name }}</h1>
         {% with phase_count=module.phases.count %}
         <div class="l-center-6">
             {% for phase in module.phases %}
-                {% if phase.is_over or phase == module.active_phase %}
```
fuzzylogic2000, line 94: everything we do in the module detail, we have to also do in the project detail in many cases. https://github.com/liqd/a4-meinberlin/blob/b079c853aa08e033fd882b718159966391262161/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html#L214 
I guess putting that into includes would be a good idea... Will add an issue!

meinberlin/apps/budgeting/phases.py
```diff
@@ -13,7 +13,8 @@ class RequestPhase(phases.PhaseContent):
     view = views.ProposalListView
 
     name = _('Request phase')
-    description = _('Participants can submit ideas, comment on and rate them.')
+    description = _('Participants can submit proposals, comment on and '
```
fuzzylogic2000, line 16: :) Thank you!

fuzzylogic2000 (commented):
Awesome! Just the module / project detail thing...

meinberlin/templates/a4modules/module_detail.html
Rineee: Oh yes, I forgot that the phases are shown there for single module!

fuzzylogic2000 (approved):
Yeah!

PR #3387, opened 2021-02-11
maps/assets: Fix select2 for polygon areasetting

rmader:
`select2` is a jQuery extension, thus we need to use the window-global
jQuery instance instead of the webpack provided one.

Depends on https://github.com/liqd/adhocracy4/pull/673

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "@babel/preset-react": "7.12.13",
     "@fortawesome/fontawesome-free": "5.15.2",
     "acorn": "8.0.5",
-    "adhocracy4": "liqd/adhocracy4#0a066f3be27d61b641c6260a8eb248e4d12b61d3",
+    "adhocracy4": "liqd/adhocracy4#4f8efd0822a424787138ac8e51d8740d4c2849a8",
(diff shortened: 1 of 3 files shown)


rmader:
@phillimorland got a much better solution!

package.json
```diff
@@ -20,7 +20,7 @@
     "@babel/preset-react": "7.12.13",
     "@fortawesome/fontawesome-free": "5.15.2",
     "acorn": "8.0.5",
-    "adhocracy4": "liqd/adhocracy4#0a066f3be27d61b641c6260a8eb248e4d12b61d3",
+    "adhocracy4": "liqd/adhocracy4#cf583fde42ac6aae4eaafdab7b338b4d83cea9bd",
```
fuzzylogic2000, line 23: Ah, that's where the wrong versions came from yesterday. :) We left the hash out of depending PRs and only added it after the a4 PRs are merged. Then, people had to locally try it by either adding the a4 hash or using the local a4 and there are no wrong hashes introduced by accident.

fuzzylogic2000 (approved):
Nice! Will merge after another a4-Pr is merged and update a4 in here.

PR #3388, opened 2021-02-12
apps/plans: fix plan detail when there are only topics not existing i…

fuzzylogic2000:
…n the settings

fixes https://sentry.liqd.net/organizations/sentry/issues/343 
It's a bit odd, because there should be only topics that are set in the settings and the field is required, so every plan should have it. But if there is a wrong topic, it should just not be shown instead of causing a 500.

### DIFF
--- a/meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
+++ b/meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
@@ -46,9 +46,11 @@
                         </dl>
                         {% endif %}
 
+                        {% if object.topic_names %}
                         <dl>
                             <dt>{% trans 'Topic' %}</dt><dd>{{ object.topic_names.0 }}{% if object.topic_names.1 %}, {{ object.topic_names.1 }}{% endif %}</dd>
                         </dl>
+                        {% endif %}
 
                         {% if object.duration %}
                         <dl>


Rineee:
I had to research why this is happening, because its so strange :grimacing: There are several old plans on stage that have topics that dont come from A4_PROJECT_TOPICS, like 'Platzmanagement' or 'Umgestaltung', but fortunately on prod it looks fine. 
The topics field used to be called theme and was a textfield before, see https://github.com/liqd/a4-meinberlin/commit/894e9d19bf208dcf23305dae51d841ea01e7c9c1. I guess this is where the strange values come from? 

But doesnt the above solution still throw an error because it still calls topic_names property of the plan model and thats where it goes wrong? 

What do you think of this in addition? https://github.com/liqd/a4-meinberlin/pull/3391/files

fuzzylogic2000:
@Rineee Yes, it used to be a textfield and not required. And also, if we just use topics on an object directly it's fine, while if I go though all objects in a QS and print out topics there, it's None. All a bit weird. But still a good idea if it doesn't break when there is weird stuff in the DB, I'd say. So yes, let's also do yours!

Rineee:
> And also, if we just use topics on an object directly it's fine, while if I go though all objects in a QS and print out topics there, it's None. 

I think this is just the printing, if you do print(object.topics) it also returns None. And this comes from (today is looking things up day) here https://github.com/goinnn/django-multiselectfield/blob/9c43f195bac628b612800958e841f0a6499ccd72/multiselectfield/db/fields.py#L190, because TopicField inherits from MultiSelectField.

PR #3390, opened 2021-02-12
styling/drowpdowns: give dropdown-menu higher z than dropdown toggle …

### LINKED ISSUE #3365: frontend: category filter broken
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-spandau/
**user:** all
**expected behaviour:** 
**behaviour:** filter broken having more than 11 categories.
**important screensize:**
**device & browser:** chrome, firefox and safari
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2021-02-05 um 10 24 16

fuzzylogic2000:
…- fixes #3365

I am not sure if it destroys something else, though. I guess, the dropdown menu should always be above the button, though?!

### DIFF
--- a/meinberlin/assets/scss/components/_dropdown.scss
+++ b/meinberlin/assets/scss/components/_dropdown.scss
@@ -42,7 +42,7 @@
     max-width: none;
     padding: 0;
     margin: 0;
-    z-index: 2;
+    z-index: 3;
     background-color: $body-bg;
     color: contrast-color($body-bg);
     border: 1px solid $text-color;


meinberlin/assets/scss/components/_dropdown.scss
```diff
@@ -42,7 +42,7 @@
     max-width: none;
     padding: 0;
     margin: 0;
-    z-index: 2;
+    z-index: 3;
```
fuzzylogic2000, line 45: @phillimorland I was not allowed to comment on the lines above, so here: At the top of this file (line 16-19), there is another z-index, that might be there to be higher than the one here?! Maybe instead of uping this one, we can remove the other one?

fuzzylogic2000: Hmm: https://github.com/liqd/a4-meinberlin/commit/10d2ca628cdf365c4eae6f31aa6aef6799c39aad I don't know, why it's there, maybe it can go?

phillimorland: this makes the boxshadow overlap the toggle, hence the z-index on line 19, making them the same also causes overlap so I would avoid if possible. 
i'm not sure if this fully solves the issue either. I don't think hiding the dropdown toggle under the dropdown is great for UX and may cause further issues if the list is very long. We would be better trying to make it drop down and extend beyond the bottom of the map, what do you think?

phillimorland (changes requested):

fuzzylogic2000:
@phillimorland Yes, true! I also thought it's not great for UX, but as there were other dropdowns in the dashboard that behave the same, I thought it's alright.
Could you have a go at this or could we together? I have no idea how to make it drop down all the way.

fuzzylogic2000:
Will close, this is not the way we want it fixed in!

phillimorland:
@fuzzylogic2000 oupps which are the other dropdowns?

fuzzylogic2000:
@phillimorland In the category dropdowns: https://meinberlin-dev.liqd.net/dashboard/modules/burgerinnenhaushalt-2-phase-6/categories/
Screenshot from 2021-02-17 11-03-06

phillimorland:
@fuzzylogic2000 balls! ok we can add the overflow-visible to the parent div but in dashboard i'm a little wary of that, I'll have a look more closely and see where it can be added.

PR #3391, opened 2021-02-12
plans/models: avoid accessing topic choices with wrong keys

Rineee:


meinberlin/apps/plans/models.py
```diff
@@ -129,7 +129,14 @@ def administrative_district(self):
     def topic_names(self):
         if hasattr(settings, 'A4_PROJECT_TOPICS'):
             choices = dict(settings.A4_PROJECT_TOPICS)
-            return [choices[topic] for topic in self.topics]
+            names = []
+            for topic in self.topics:
+                try:
+                    name = choices[topic]
+                except:
+                    name = topic
```
Rineee, line 137: or pass?

fuzzylogic2000: Hmm. I like this one!

fuzzylogic2000 (approved):

meinberlin/apps/plans/models.py
fuzzylogic2000: Oh right! The linter complains and is absolutely right! You shouldn't just use try without the actual error. So, could you please add that?

Rineee:
In the multiselectfield code I saw, that there actually is this get function for dictionaries that handles key errors :)

fuzzylogic2000 (approved):
Nice! Good, that you had a look!

PR #3411, opened 2021-02-16
Ks 2021 02 remove livestream object when its deleted

### LINKED ISSUE #3408: testing 3781 participatory budgeting - template wording change
**URL:** https://meinberlin-dev.liqd.net/dashboard/projects/testing-burgerhaushalt-2-phases/basic/
**user:** initiator
**expected behaviour:** template wording: 2 Phasen
**behaviour:** template wording: 1 Phase (missing n)
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2021-02-16 um 14 19 52

Rineee:
Live stream deletion is still a bit odd, but maybe a little better like this?

### DIFF
--- a/meinberlin/apps/livequestions/models.py
+++ b/meinberlin/apps/livequestions/models.py
@@ -78,6 +78,10 @@ class LiveStream(module_models.Item):
     )
 
     def save(self, *args, **kwargs):
-        self.live_stream = transforms.clean_html_field(
-            self.live_stream, 'video-editor')
-        super().save(*args, **kwargs)
+        if self.live_stream and '<iframe' in self.live_stream:
+            self.live_stream = transforms.clean_html_field(
+                self.live_stream, 'video-editor')
+            super().save(*args, **kwargs)
+        elif self.__class__.objects.filter(module=self.module).exists():
+            for stream in self.__class__.objects.filter(module=self.module):
+                stream.delete()
--- a/meinberlin/templates/a4modules/module_detail.html
+++ b/meinberlin/templates/a4modules/module_detail.html
@@ -62,7 +62,7 @@
             <!-- Live stream -->
             {% for phase in module.phases %}
               {% if phase == module.active_phase and live_stream %}
-                  {{ live_stream.live_stream | transform_collapsibles | richtext }}
+                  {{ live_stream.live_stream | richtext }}
                   <span class="label label--active">•{% trans ' Live now' %}</span>
               {% endif %}
             {% endfor %}


meinberlin/apps/livequestions/models.py
```diff
@@ -78,6 +78,10 @@ class LiveStream(module_models.Item):
...
-        self.live_stream = transforms.clean_html_field(
-            self.live_stream, 'video-editor')
-        super().save(*args, **kwargs)
+        if self.live_stream and '<iframe' in self.live_stream:
+            self.live_stream = transforms.clean_html_field(
+                self.live_stream, 'video-editor')
+            super().save(*args, **kwargs)
+        elif self.__class__.objects.filter(module=self.module).exists():
+            for live_stream in self.__class__.objects.filter(module=self.module):
+                live_stream.delete()
```
fuzzylogic2000, line 87: I don't quite get what it does. Maybe we can talk tomorrow?

fuzzylogic2000:
And the linter complains... Is our pre-commit-hook currently not working?

meinberlin/apps/livequestions/models.py
Rineee: I should have added some explanation, sorry .. so yes, lets talk about it!

Rineee:
> And the linter complains... Is our pre-commit-hook currently not working?
Apparently not :/

fuzzylogic2000:
@Rineee You still need to explain this to me. :)

meinberlin/apps/livequestions/models.py
Rineee: @fuzzylogic2000 Ok, I now remeber what I did :) So: the problem was that if the livestream is deleted from the ck editor in dashboard, the object was still there in the db and thus when asking in template for {% if livestream%}, it was true and the 'Live now' was showing. That is why I thought to manually delete the object when the livestream is saved without content (and thus deleted). This is what the `if self.live_stream` part is for.
The `'<iframe' in self.live_stream` part is bc when deleting the livestream in the ckeditor, you actually have to press delete twice, because when only once, there still is an empty div (you can see that in the lower left of ckeditor frame). So if you only press delete once, the empty div will still be saved as livestream object. 
I know this is not a very pretty solution and it would have been better to tell ckeditor to not accept empty divs, but Philli and me tried and didnt succeed..

Rineee:
I rebased on master and now this stupid revert commit, reverts the (now right) wording from master.. how would you solve that @fuzzylogic2000 ? Revert the revert?

fuzzylogic2000:
I am not sure, I get what you mean. Can you just remove the second commit from this PR and all is fine? (`git reset HEAD~1`)

Rineee:
> I am not sure, I get what you mean. Can you just remove the second commit from this PR and all is fine? (`git reset HEAD~1`)

When I opened the PR, I accidentally commited the tranlsation fix and then reverted it again. No the translation is fixed on master and thus rebasing led to this commit.. but ok, I will try to remove the revert.

meinberlin/apps/livequestions/models.py
fuzzylogic2000: Cool! Thank you!

fuzzylogic2000 (approved):
I do get it now and you can merge once the git-thing is fixed. :)

Rineee:
> I do get it now and you can merge once the git-thing is fixed. :)

Cool! Yes, resetting head worked! :)

Rineee:
Ok and now there is another problem.. will fix it after lunch

Rineee:
Oh god, it was a linting error and I couldnt solve it, no matter what.. I now renamed the variable to have a shorter line :shrug:

PR #3415, opened 2021-02-17
webpack: Explicitly declare chunk dependencies

rmader:
We implicitly assume that adhocracy4.[js|css] and vendor.[js|css]
are always present. Make this dependency explicit in Webpack,
so entry point that import the same dependencies don't need to include
the same code.

This appears to slim down our overall js load quite a bit.


phillimorland (approved):
super cool! :)

PR #3417, opened 2021-02-18
apps/cms: remove projects block from home page

### LINKED ISSUE #2255: Project block homepage
We no longer use project block on homepage and have deleted the template but you can still add block. We should consider moving it to the CC and recovering template as it is a useful block to have.

rmader:
Rebase of #2262. Can't reproduce any migration issues here, maybe that was fixed upstream?

fixes #2255


fuzzylogic2000 (approved):
Yes, it works!

fuzzylogic2000:
@phillimorland With this working, should we remove more blocks? Like the actions block (and there might be more). Currently it's not looking too bad, but everything that is not needed on the new landing page might break quite badly at some point and it's probably safer to remove what we don't want to be used anyway?! Maybe also talking to @CarolingerSeilchenspringer to make sure we only delete what is not needed anymore. :p

PR #3420, opened 2021-02-18

rmader:
client cpu load. Skip rendering until the tile becomes visible.

This will need some testing. Also it might make sense to set the interval number slightly higher as otherwise we rist to not initially show enough tiles on big screens.

### DIFF
--- a/meinberlin/apps/plans/assets/PlansList.jsx
+++ b/meinberlin/apps/plans/assets/PlansList.jsx
@@ -2,6 +2,9 @@
 const React = require('react')
 const Moment = require('moment')
 
+const autoScrollThreshold = 500
+const itemListInterval = 10
+
 class LazyBackground extends React.Component {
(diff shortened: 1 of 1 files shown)


rmader:
@fuzzylogic2000 @Rineee @phillimorland: this appears to help with the project list performance on slow clients - maybe you can have a look if this looks reasonable to you.

phillimorland:
I would refactor this if we want to keep it but maybe we can merge and test on dev first? locally it's not making any difference but I wouldn't expect it to, what do you think @fuzzylogic2000 @Rineee

fuzzylogic2000:
@phillimorland Yes, I think this is a goof idea! Let's try it!

phillimorland:
This unfortunatly causes the filtering to not work unless it's in list view, I will revert merge and work on fixing and refactoring as needed to be done anyway.

Issue #3422, opened 2021-02-19
safari: cannot vote in participatory budgeting 2. phase/ whole comment/voting section not shown in other modules


CarolingerSeilchenspringer:
**URL:** https://meinberlin-dev.liqd.net/budgeting/2021-00520/
**user:** registered
**expected behaviour:** I can vote and 2 phase participatory budgeting and vote and comment in other modules
**behaviour:** I cannot vote in project linked above. In other projects (https://meinberlin-dev.liqd.net/budgeting/2021-00514/) the whole comment/voting section isn't shown.
**important screensize:**
**device & browser:** safari (both desktop and mobile for problem 1 and desktop only for not showing comment section)
**Comment/Question:** these are two problems but I suppose they are related?

<img width="603" alt="Bildschirmfoto 2021-02-19 um 16 03 21" src="https://user-images.githubusercontent.com/35491681/108522535-35b74f00-72cd-11eb-9df9-17060bf463df.png">
<img width="570" alt="Bildschirmfoto 2021-02-19 um 16 03 32" src="https://user-images.githubusercontent.com/35491681/108522542-36e87c00-72cd-11eb-9d43-88ae2d5e552f.png">

Screenshot?

fuzzylogic2000:
This happened on all browsers on idea pages with maps, once logged out. It couldn't find adhocracy.js, then. 
That happenend with this commit https://github.com/liqd/a4-meinberlin/commit/23fd841f69138fe15e298a47b31636cabae33fea Could you have another go at this @rmader ?

fuzzylogic2000:
@carolin This is solved, but I will keep it open for @rmader  to check out and see how to redo his PR, we reverted.

rmader:
Should be properly fixed by https://github.com/liqd/a4-meinberlin/pull/3445 now

PR #3440, opened 2021-02-24
webpack: Explicitly declare chunk dependencies

rmader:
This reverts https://github.com/liqd/a4-meinberlin/pull/3423 again. It appears to work now, at least I can't reproduce https://github.com/liqd/a4-meinberlin/issues/3422


phillimorland:
will test on mac and check safari

phillimorland (approved):
nice!

PR #3462, opened 2021-03-03
webpack: Remove vendor chunk

rmader:
Using the vendor chunk is, IIUC, a leftover from Webpack <= 3 times.
Webpack >= 4 promote a different kind of chunk splitting. This is
a step into that direction.

The probably biggest benefit is that there's now a single big `adhocracy4`
entry point which includes all base dependencies, avoid odd ordering issues
and making it clearer where to add stuff to.

Port of some improvements found in https://github.com/liqd/adhocracy-plus/pull/1128, which again was a port of https://github.com/liqd/a4-meinberlin/pull/3440

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -5,31 +5,18 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 
 module.exports = {
   entry: {
-    vendor: [
+    adhocracy4: [
       '@fortawesome/fontawesome-free/scss/fontawesome.scss',
       '@fortawesome/fontawesome-free/scss/brands.scss',
       '@fortawesome/fontawesome-free/scss/regular.scss',
       '@fortawesome/fontawesome-free/scss/solid.scss',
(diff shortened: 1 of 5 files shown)


webpack.common.js
```diff
@@ -5,31 +5,18 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin')
...
-      'react-dom',
-      'react-flip-move',
-      'react-sticky-box'
+      'shariff/dist/shariff.min.css',
+      'select2/dist/css/select2.min.css',
+      'slick-carousel/slick/slick.css',
+      './meinberlin/assets/extra_css/_slick-theme.css',
+      './meinberlin/assets/scss/style.scss',
+      './meinberlin/assets/js/app.js'
     ],
```
phillimorland, line 19: where do these go?

phillimorland (approved):
looking good and working localy, will wait for travis

PR #3463, opened 2021-03-03
apps/embed/assets: Work around login popup not closing

rmader:
This affects cases when people login from the login-link shown
e.g. above comments.


phillimorland (approved):
awesome!

PR #3466, opened 2021-03-05
Ks 2021 03 change phase form titles

Rineee:
Hm, I thought there was a way to add the helptext to the form, but I couldnt find one as its all done via components and I think they dont allow helptexts? So now I just added it in the template (as done for affiliations).


fuzzylogic2000 (approved):
Nice!
And yes, bacuse the helptext is for the whole form and not one of the fields it can't be added differently afaik.

PR #3467, opened 2021-03-05
Ks 2021 02 add contact fields to participatory budgeting

Rineee:
There are still some styling issues, but I thought we could merge and then I'll look at them together with @phillimorland in the afternoon?


webpack.common.js
```diff
@@ -112,7 +112,13 @@ module.exports = {
         'adhocracy4/adhocracy4/categories/assets/select_dropdown_init.js'
       ],
       dependOn: 'adhocracy4'
-    }
+    },
+    budgeting_disable_contact: {
+      import: [
+        './meinberlin/apps/budgeting/assets/disable_contact.js'
+      ],
+      dependOn: 'adhocracy4'
```
Rineee, line 120: Do I need this here?

Rineee:
Hm, when I do `git reabse master` it says current branch is up to date.

Rineee:
> Hm, when I do `git reabse master` it says current branch is up to date.

Ok, ignore me, didnt have the newest master.. will rebase translation files now, wohoo

webpack.common.js
rmader: This tells webpack that adhocracy4.js will always be present when this endpoint is used, allowing to avoid duplicates in imports.

Locking at `disable_contact.js` there appears to be only a implicit import: jquery (`$(document).ready`). It's likely a good idea to avoid jquery in all new code so if that's done - then this `dependOn` is not necessary.

meinberlin/apps/budgeting/assets/disable_contact.js
```diff
@@ -0,0 +1,43 @@
+function disableContact (disable, disableTextfield) {
+  document.getElementById('id_contact_email_0_0').disabled = disable
+  document.getElementById('id_contact_email_0_1').disabled = disable
+  document.getElementById('id_contact_email_1').disabled = disableTextfield
+  document.getElementById('id_contact_phone').disabled = disable
+}
+
+$(document).ready(function () {
```
rmader, line 8: In order to avoid jQuery and to make sure it works in embedded cases:
```
function foo () {
...
}
document.addEventListener('DOMContentLoaded', foo, false)
document.addEventListener('a4.embed.ready', foo, false)
```

fuzzylogic2000:
We want to add the changes of the tabbing order and then merge, right?

Rineee:
> We want to add the changes of the tabbing order and then merge, right?

@phillimorland @fuzzylogic2000 It is changed now with the checkbox on top, what do you think?

fuzzylogic2000 (approved):
Very nice!
Only thing to think about before merge: old proposals now have the checkbox checked, but nothing is exported because the fields are empty, right? Should we add a migration with this PR to set the checkbox to false in the already existing propsals?

Rineee:
> Very nice!
> Only thing to think about before merge: old proposals now have the checkbox checked, but nothing is exported because the fields are empty, right? Should we add a migration with this PR to set the checkbox to false in the already existing propsals?

Yes, you are right! New migration is there now..

fuzzylogic2000 (approved):
Whoop! Nice!

PR #3481, opened 2021-03-11
WIP: webpack: IE11 fixes and workarounds

### LINKED ISSUE #3441: bplan statement form blank on IE 11
**URL:** https://www.stadtentwicklung.berlin.de/planen/b-planverfahren/de/fruehbb/6-47/index.shtml
**user:** any
**expected behaviour:** statement form can be used
**behaviour:** statement form is empty
**important screensize:** 
**device & browser:** Windows, IE 11
**Comment/Question:** can you look into it soon and then we decide how to handle/prioritize it?

Screenshot?
Bildschirmfoto 2021-02-24 um 14 17 48

rmader:
From the first commit:
```
Webpack 5 dropped ES5 polyfills, required for IE11 support. Explicitely
specify ES5 support (`target`), making Webpack complain about everything
that's missing. Install all required polyfills and make Webpack fall
back to them.
```

Sad thing is that we loose chunk dependencies again, resulting in bigger js files (like we had before https://github.com/liqd/a4-meinberlin/pull/3440) and, because we now transpile almost all js (but mapbox because of a bug in babel), in longer build times, even during development.

The last point, longer dev build times, can probably be circumvented by moving the babel-loader config to production settings only.

Closes #3441


phillimorland:
@rmader is this still WIP? Also once we do not need to support IE11 anymore, would you say we could just rebase and implement https://github.com/liqd/a4-meinberlin/pull/3462 ? obviously after extensive testing

rmader:
Yeah I'm sorry, looks like this is blocked by https://github.com/parcel-bundler/parcel/issues/1128 (there are a few similar bugs in different projects). Short story is that if we transpile all js dependencies (so it's compatible with IE11) we run into some babel error. It might be possible to somehow exclude all relevant packages, but I still haven't managed to pick the right ones. Another option is to wait and hope this gets fixed in babel.

Either way I don't quite find a solution that makes IE11 work properly in embeds (and other areas) while not breaking mapbox for all browsers :(

phillimorland:
We no longer have to support IE11 so will close :tada:

PR #3484, opened 2021-03-15
remove project type function, add some tests

fuzzylogic2000:
With the last commit the problem was a test (which I didn't touch) broken, because the plans were in a different order.

And coveralls is wrong! If I compare the coverage to the last build of master, it increases by 0.03. Not a lot, but it doesn't decrase the coverage!


phillimorland (approved):
Nice!

PR #3502, opened 2021-03-19
Ks 2021 03 issues

Rineee:
The last commit fixes that when the contact form has been disabled (and thus no contact info is sent) and there is a form validation error somewhere else in the form, the radio button is still checked.


meinberlin/apps/contrib/mixins.py
```diff
@@ -58,6 +58,7 @@ def clean(self):
                            _("You want to upload an image. "
                              "Please check that you have the "
                              "right of use for the image."))
+        return cleaned_data
```
fuzzylogic2000, line 61: good catch!

fuzzylogic2000 (approved):
Looks good!

PR #3516, opened 2021-03-25
Pm 2021 03 add break word

phillimorland:


package.json
```diff
@@ -81,7 +81,7 @@
     "lint-staged": "10.5.4",
     "markdownlint-cli": "0.27.1",
     "stylelint": "13.12.0",
-    "stylelint-config-standard": "20.0.0",
+    "stylelint-config-standard": "21.0.0",
```
phillimorland, line 84: I branched off wrong one but this is already merged so it shouldn't be showing up

fuzzylogic2000 (approved):
Nice!

PR #3550, opened 2021-04-19
translations: make po - replace all dsgvo-video-embed paths - fixes #…

fuzzylogic2000:
…3548

There was `dsgvo-video-embed/dist/...` as well as `dsgvo-video-embed/js/...`


phillimorland (approved):
LGTM

PR #3551, opened 2021-04-19
Kl 2021 04 disallow container creation

fuzzylogic2000:


meinberlin/apps/projectcontainers/templates/meinberlin_projectcontainers/container_list_dashboard.html
```diff
@@ -8,8 +8,9 @@
         <h1 class="lr-bar__left">
             {% trans 'Containers' %}
         </h1>
+        {% has_perm 'meinberlin_projectcontainers.add_container' request.user as can_add_project %}
         <div class="lr-bar__sm-left-lg-right">
-            <a href="{% url 'a4dashboard:container-create' organisation_slug=view.organisation.slug %}" class="btn">
+            <a href="{% url 'a4dashboard:container-create' organisation_slug=view.organisation.slug %}" class="btn {% if not can_add_project %}disabled{% endif %}" {% if not can_add_project %}aria-disabled="true"{% endif %}>
                 {% trans 'New Container' %}
```
phillimorland, line 14: disabled comes from disabled attribute so can just have `class="btn" {% if not can_add_project %}disabled aria-disabled="true"{% endif %}`

phillimorland (changes requested):
super cool, just one little addition :)

meinberlin/apps/projectcontainers/templates/meinberlin_projectcontainers/container_list_dashboard.html
fuzzylogic2000: Ah, I tried this because I hoped for bootstrap and their docs said to add it like that for links. https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state But as we don't have that bit of bootstrap and the styling comes later anyway, yes, I will do that!

phillimorland: yeah we do that in a+ but not here as we don't use much boootstrp, I added issue to add example to comp library so we have a record of how we do it so it's at least somewhere

fuzzylogic2000: Yes, I did change it! Can we merge or should we wait for the styles to be added?

phillimorland: sorry I got lost in some database weirdness merging :p

PR #3552, opened 2021-04-19
apps/organisations: allow organistaion change only to admins

fuzzylogic2000:


phillimorland (approved):
Nice! also checked and coverage seems to be confused

PR #3558, opened 2021-04-20
apps/organsiations: remove unused predicate - we use the one from a4

fuzzylogic2000:
We already use the predicate from a4 everywhere anyway, the rule for the organisation update was the only one using it.


Rineee (approved):
Yes, cool, lets get rid of it ;)

PR #3560, opened 2021-04-20
apps/dashboard: filter dashboard project, bplan, extproject, containe…

### LINKED ISSUE #2616: dashboard project list: pagination doesn't know if list filtered for group members
Screenshot from 2019-10-31 12-07-24

fuzzylogic2000:
…r lists for group members - fixes #2616

It would have been much better to filter by permissions instead of doing
this dance, but as explained in issue #2616 it's not that easy. With
this we have to take care, when we add or remove project update
permissions, as this (instead of working with permissions) works with
knowing the roles and showing respective projects only to people allowed
to update them.

PS to the reviewer: If I want to test sth. with querysets, I just decrease its size instead of adding that many projects. :)

### DIFF
--- a/meinberlin/apps/bplan/views.py
+++ b/meinberlin/apps/bplan/views.py
@@ -7,6 +7,7 @@ from adhocracy4.dashboard.components.forms.views import \
     ProjectComponentFormView
 from adhocracy4.dashboard.mixins import DashboardBaseMixin
 from meinberlin.apps.bplan import phases as bplan_phases
+from meinberlin.apps.dashboard.mixins import DashboardProjectListGroupMixin
 from meinberlin.apps.extprojects.views import ExternalProjectCreateView
 
 from . import forms
@@ -55,7 +56,8 @@ class BplanProjectUpdateView(ProjectComponentFormView):
         return self.project
 
 
-class BplanProjectListView(DashboardBaseMixin,
+class BplanProjectListView(DashboardProjectListGroupMixin,
+                           DashboardBaseMixin,
                            generic.ListView):
     model = models.Bplan
     paginate_by = 12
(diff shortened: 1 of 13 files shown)


fuzzylogic2000:
Right! Tests! Are failing and I still need to add some as well...

fuzzylogic2000:
Now with tests. :)

fuzzylogic2000 (commented):
@Rineee pointed out, that moderators should also be able to see their projects...

fuzzylogic2000:
@Rineee We decided, that the edge case is so edgy and that we can also manually fix it if need be, by adding the moderation project to the group, that this can stay like it is now. 
So, if there wasn't anything else, please merge. :)

meinberlin/apps/bplan/views.py
```diff
@@ -64,9 +66,17 @@ class BplanProjectListView(DashboardBaseMixin,
...
             organisation=self.organisation
         )
+        if (is_initiator(self.request.user, self.organisation)
+                or is_superuser(self.request.user)):
+            return qs
+        else:
+            return qs.filter(
+                group__id__in=[self.request.user.groups.values_list(
+                    'id', flat=True)]
+            )
```
Rineee, line 79: Im a bit unhappy with repeating this everywhere.. maybe we could have a method (sth like get_projects_for_user) and put it into a mixin?

Rineee:
> @Rineee We decided, that the edge case is so edgy and that we can also manually fix it if need be, by adding the moderation project to the group, that this can stay like it is now.
> So, if there wasn't anything else, please merge. :)

Alright, yes, it is really edgy, so makes sense.
I made a comment above, but apart from that it works perfect. If you dont like the idea and it needs to be merged quickly, feel free to then just merge it yourself :)

meinberlin/apps/bplan/views.py
fuzzylogic2000: Hmmmmm, but as a mixin defining get_queryset, I could only use it for two of the four cases, because the others have other extra-stuff in it. Ah, right, or I could then add/filter for the other things afterwards.

Rineee (approved):
Supercool!

PR #3561, opened 2021-04-21
Make container feature inactive

khamui:
Using flag can_add_project to show enabled or disabled button
Along with hint/note on the left side (incl. exclamation icon)
Taiga task 4082

### DIFF
--- a/meinberlin/apps/projectcontainers/templates/meinberlin_projectcontainers/container_list_dashboard.html
+++ b/meinberlin/apps/projectcontainers/templates/meinberlin_projectcontainers/container_list_dashboard.html
@@ -9,11 +9,23 @@
             {% trans 'Containers' %}
         </h1>
         {% has_perm 'meinberlin_projectcontainers.add_container' request.user as can_add_project %}
-        <div class="lr-bar__sm-left-lg-right">
-            <a href="{% url 'a4dashboard:container-create' organisation_slug=view.organisation.slug %}" class="btn" {% if not can_add_project %}disabled aria-disabled="true"{% endif %}>
-                {% trans 'New Container' %}
-            </a>
-        </div>
+        {% if not can_add_project %}
+            <div class="lr-bar__flex--wrapper u-spacer-top">
+                <div class="lr-bar__flex--wrapper nowrap">
+                    <i class="fas fa-exclamation-circle u-danger u-icon-spacing" aria-hidden="true"></i>
+                    <p class="u-danger u-font-italic u-spacer-right">{% trans 'No new project containers can be created. Please use the multimodule feature. Existing project containers can still be edited.' %}</p>
+                </div>
+                <a class="btn--flex bottom is-disabled" aria-disabled="true">
+                    {% trans 'New Container' %}
+                </a>
+            </div>
+        {% else %}
+            <div class="lr-bar__sm-left-lg-right">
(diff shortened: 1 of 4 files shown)


meinberlin/assets/scss/components/_button.scss
```diff
@@ -29,6 +29,7 @@
         background-color: mix($color, $color-contrast, $weight-bg);
         color: mix($color, $color-contrast, $weight-fg);
         cursor: not-allowed;
+        pointer-events: none;
     }
```
phillimorland, line 33: nice!

meinberlin/assets/scss/utility.scss
```diff
@@ -193,3 +201,7 @@
         display: none !important;
     }
 }
+
+.u-font-italic {
+    font-style: italic;
+}
```
phillimorland, line 207: when adding utilities, we try to only add them if it is garenteed we will reuse them so as not to have a huge long list, as 2 of these are for positioning an icon, I would be more inclined to make an icon position class to cover the position with one class name and as it's not a spacer size we are likley to use for any other purpose.

phillimorland (changes requested):
Don't forget mobile :p design can be found: https://app.zeplin.io/project/5b051eef7aaa417d2227c55c/screen/6012dea2f36a2ea83f461f19
Otherwise nice and clean!

meinberlin/assets/scss/utility.scss
fuzzylogic2000: Where would you put that? Do we have something for that or is that also a utility?

meinberlin/apps/projectcontainers/templates/meinberlin_projectcontainers/container_list_dashboard.html
```diff
@@ -11,8 +11,10 @@ <h1 class="lr-bar__left">
         {% has_perm 'meinberlin_projectcontainers.add_container' request.user as can_add_project %}
         {% if not can_add_project %}
             <div class="lr-bar__flex--wrapper u-spacer-top">
-                <i class="fas fa-exclamation-circle u-danger u-spacer-right-half u-spacer-top-quarter"></i>
-                <p class="u-danger u-font-italic u-spacer-right">{% trans 'No new project containers can be created. Please use the multimodule feature. Existing project containers can still be edited.' %}</p>
+                <div class="lr-bar__flex--wrapper nowrap">
+                    <i class="fas fa-exclamation-circle u-danger u-icon-spacing"></i>
```
fuzzylogic2000, line 15: The font awesome icons always need something for a11y: either `aria-hidden=true` when they don't have any extra meaning that's not covered by the text. Or an aria-label when you need it, for example for the magnifying glass in a search box.

fuzzylogic2000 (commented):
Cool! Just one small thing.
And here you can also put everything in one commit. :)

meinberlin/assets/scss/utility.scss
khamui: i think for this case it will be re-used in future so can be there as utility class right @phillimorland ?

phillimorland: yeah, margin in just one style, if we have to use more then 2 or 3 then it's probably not a utility

fuzzylogic2000 (approved):
Cool!

PR #3563, opened 2021-04-22
Center title and wording text

### LINKED ISSUE #3547: title not centered on landing page
**URL:** https://mein.berlin.de/
**user:** any
**expected behaviour:** title and wording centered
**behaviour:** they are left alighned
**important screensize:** 
**device & browser:** testet on chrome and firefox
**Comment/Question:** is that something I can change via wagtail (couldn't find it) or an issue?

Screenshot?
Bildschirmfoto 2021-04-16 um 10 33 25

khamui:
Adding css selector h2 and p for .block--map-teaser_background
fixes #3547

Note:
I have to move the entire class, because of css specificity (what a word!)

stylelint complaint about it.
https://stylelint.io/user-guide/rules/no-descending-specificity

### DIFF
--- a/meinberlin/assets/scss/components/_blocks.scss
+++ b/meinberlin/assets/scss/components/_blocks.scss
@@ -213,6 +213,7 @@
 .block--map-teaser_background {
     background-image: radial-gradient($white-shadow, rgba(0, 0, 0, 0));
     border-radius: 100px;
+    text-align: center;
 }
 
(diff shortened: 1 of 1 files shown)


fuzzylogic2000 (changes requested):
Nice!
But could you please squash the commits? It will be easier to find out what we did and why if we ever need to.

fuzzylogic2000 (approved):

PR #3564, opened 2021-04-23
apps/notifications: only send schedule notifications for a4 projects …

### LINKED ISSUE #1901: moderators for b-plans
if an initiator starts a b-plan via meinBerlin (as e.g. SenWohn does, they don't have imperia) or an external project he/she is automatically added as moderator and gets mails as the one below. This is confusing because:
a) you don't see moderators in dashboard
b) you can't follow a b-plan/external project
c) the link does not go to the external page (in this case it goes here: https://mein.berlin.de/projects/bebauungsplan-8-66-buckower-felder/)

Should we take out this rule for these two templates or stop sending mails?

bildschirmfoto 2019-02-04 um 13 32 08

fuzzylogic2000:
…- fixes #1901

To test, add phases to an a4 project, a bplan and an external project. They all need to end soon (>24 hours). Now create the schedule actions by `python manage.py create_system_actions`. Now there should be three new actions, but only one new task.

I tried to add tests, but then didn't want to invest too much time. :grimacing:

### DIFF
--- a/meinberlin/apps/notifications/signals.py
+++ b/meinberlin/apps/notifications/signals.py
@@ -25,7 +25,8 @@ def send_notifications(instance, created, **kwargs):
         if action.project:
             emails.NotifyModeratorsEmail.send(action)
 
-    elif action.type == 'phase':
+    elif (action.type == 'phase' and
+          action.project.project_type == 'a4projects.Project'):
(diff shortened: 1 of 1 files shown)


Rineee (approved):
Yeah, cool!

PR #3578, opened 2021-04-27
translations: adjusting 'make po' command in Makefile to work with...

khamui:
…mac os and ubuntu.

To be tested in macOS and linux (ubuntu).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ NODE_BIN = node_modules/.bin
 SOURCE_DIRS = meinberlin tests
 ARGUMENTS=$(filter-out $(firstword $(MAKECMDGOALS)), $(MAKECMDGOALS))
 
+# for mac os gsed is needed (brew install gnu-sed)
 SED = sed
 ifneq (, $(shell command -v gsed))
 	SED = gsed


khamui:
@phillimorland can you please test this if the `make po` command works for both macOs and ubuntu (you are the chosen one because you have the power of both OS right)?

phillimorland:
I am! so mac is good btu for ubuntu I get 
`sed -i '' 's%#: .*/adhocracy4%#: adhocracy4%' locale/*/LC_MESSAGES/django*.po
sed: can't read s%#: .*/adhocracy4%#: adhocracy4%: No such file or directory
make: *** [Makefile:130: po] Error 2`

phillimorland (changes requested):
see comment :)

fuzzylogic2000:
So, no, this does not work in most of our shells. Maybe @goapunk has an idea? It's about the sed command (see Files changed), which in @khamui zshell only works with the empty string added at front. Is there a way for the Makefile to test in which shell it is used?
Otherwise @khamui you could change it in your Makefile every time and not commit it or use another shell for the make po command?

khamui:
okay i pushed another version which adds this workaround (empty string) for macOS, but keeps it without for all other operating systems. I could remove the linux case and keep * for all others. @phillimorland if this would work for macOs and LInux for you we could go with that?

Otherwise i will just change that everytime for me as @fuzzylogic2000 proposed.

goapunk:
this shouldn't be necessary if you install gsed on mac (brew install gnu-sed), maybe we should document this somewhere.

khamui:
as @goapunk stated above, i removed that fix. And instead mac users should install gsed and all works fine! I added a comment to mention this. Not sure if this is enough.

fuzzylogic2000 (approved):
Nice! And I think we will remember this. :)

PR #3579, opened 2021-04-28
module templates: rm background colour and add spacer where required for filter reposition

phillimorland:


phillimorland:
This is awaiting approval from Daniel

fuzzylogic2000 (approved):
Looks good!
Can we completely remove `control-bar__top-overlap`, though?

PR #3580, opened 2021-04-28
templates/map_filter_and_sort//Toggles: rm switch--btn class when no …

### LINKED ISSUE #3575: #4064 [mB] change filters above maps for mobile version UI
**URL:** https://meinberlin-dev.liqd.net/projekte/test-burgerhaushalt/?mode=list
**important screensize:**Both*
**device & browser:** *e.g. Desktop, firefox 68.0.2 (64-Bit)*
**Comment/Question:** *If possible, the only thing I would change here is that the inactive state should have the text in black and the hover should be light-gray instead of dark blue like the other buttons* 

[200427-d-change-filters-above-maps.pdf](https://github.com/liqd/a4-meinberlin/files/6384130/200427-d-change-filters-above-maps.pdf)

phillimorland:
…active fixes #3575


fuzzylogic2000 (approved):
Nice!

PR #3587, opened 2021-05-03

phillimorland:
client cpu load. Skip rendering until the tile becomes visible.

Need: 
- [ ] refactoring
- [ ] to work with filters on both map and list view


phillimorland:
@fuzzylogic2000 @Rineee @goapunk been thinking about this and the actual most important load is the initial one where we have a default filter for 'laufend', however we fetch the whole list can anyone remember why we died that instead of requesting only the laufend list from backend? I feel there probably was a good reason we do the filtering in the front but I don't know it?!

PR #3597, opened 2021-05-06
tests/projects: add tests for moderator and participant invite accept…

fuzzylogic2000:
… and reject


phillimorland (approved):
Passes all the tests :)

PR #3646, opened 2021-06-15
modules/poll: refactoring of poll management

khamui:
- popper for submenu (create predefined/open questions), if out of view, it will automatically flip on other side
- adjust form for multiple choice question (legacy) and adding new form for open question
- additional other checkbox with disabled other input for multiple choice questions


khamui:
Just to mention about the naming, i tested throughout developing a lot of namings like SurveyEdit, PollEdit etc. At the end I find it more confusing. Although the current name of PollManagement and QuestionForm/OpenQuestionForm is also confusing. If you have a suggestion for a good naming, I appreciate suggestions. For now its okay like that, but if we shift it at some point to adhocracy4 and use it as central component, it should have a more clear set of names, no?

meinberlin/apps/polls/assets/PollManagement.jsx
```diff
@@ -173,20 +172,23 @@ export const PollManagement = (props) => {
     },
     popperMenuItems: [
       {
-        styleClass: 'btn btn--light btn--small',
+        styleClass: 'btn btn--light btn--small submenu-item__first',
         text: 'vorgegebene Antworten',
```
Rineee, line 176: better use English and we then translate by using django.gettext.. you can just use your favourite english wording as mB is only in German :)

meinberlin/apps/polls/assets/PollManagement.jsx
```diff
@@ -37,13 +38,21 @@ export const PollManagement = (props) => {
...
         getNewChoice(),
         getNewChoice()
       ]
     }
   }
 
+  const getNewOpenQuestion = (label = '') => {
+    const newQuestion = getNewQuestion(label)
+    newQuestion.is_open = true
+    return newQuestion
```
Rineee, line 53: maybe add `newQuestion.choices = []`?

Rineee (approved):
Super awesome cool!

PR #3647, opened 2021-06-18
Kl 2021 06 use polls fom a4

fuzzylogic2000:
Don't forget to save your DB before trying this out! (and maybe add some "old" polls before you check this out and save it)


fuzzylogic2000:
Heureka! here is the migration! @Rineee Maybe you can already have a look? Now we only have to use the stuff from a4 to be able to see the data from outside of the shell or DB as well...

fuzzylogic2000:
Oh, and I think, we should try the migration on prod-data as well. :grimacing: Will need some help for that. @goapunk @Rineee :smile_cat:

goapunk:
> Oh, and I think, we should try the migration on prod-data as well. grimacing Will need some help for that. @goapunk @Rineee smile_cat
sure :)

Rineee (approved):
Looking good! I migrated my polls and everything is fine.. should we dare to merge? :scream:

fuzzylogic2000:
> Looking good! I migrated my polls and everything is fine.. should we dare to merge? scream

@goapunk Did you save the database? In that case, I would say: go for it!

fuzzylogic2000:
Oh, and @CarolingerSeilchenspringer Do you already have old-poll-test-projects? We shouldn't merge before...

CarolingerSeilchenspringer:
@fuzzylogic2000 didn't know it was so urgent, will do it right away.

PR #3659, opened 2021-06-23
apps/polls: do not overwrite poll templates from a4

fuzzylogic2000:
It does not make sense, that it was there as we were not using the polls from a4.
restores the timeline/cluster logic depending on https://github.com/liqd/adhocracy4/pull/745

So, the PR in a4 also needs to be merged and then updated in mB.

To keep the current permissions logic https://github.com/liqd/adhocracy4/pull/747  also needs to be merged first and then updated


fuzzylogic2000:
It will fail until a4 is updated.

phillimorland (approved):
template working but cluster still odd, will merge and see if's my install

PR #3660, opened 2021-06-24
Pm 2021 06 bootstrap fixes

phillimorland:


webpack.common.js
```diff
@@ -17,6 +17,12 @@ module.exports = {
...
     ],
+    blueprint_picker: {
+      import: [
+        './meinberlin/assets/js/blueprint-picker.js'
+      ],
+      dependOn: 'adhocracy4'
+    },
     captcheck: {
       import: [
         './meinberlin/apps/captcha/assets/captcheck.js'
```
phillimorland, line 28: @goapunk quick explanation: when you add a seperate endpoint in webpack you can load it individually. I should have done this originally especially as this is only used on 1 page, app.js should only be used for multipage use code, otherwise you run into issues like this :)

goapunk: nice, sounds like a good solution. Thanks for explaining!

Rineee (approved):
Yaaaaay

PR #3662, opened 2021-06-28
apps/polls: use react from a4

fuzzylogic2000:
Depending! Only merge when the react for the polls is all in a4!


fuzzylogic2000:
@phillimorland Here it is!

Rineee (approved):
Coool!

PR #3663, opened 2021-06-28
modules/poll: GET poll with REST API

khamui:


khamui:
linking the actual api method here:
https://github.com/liqd/adhocracy4/pull/748

in order for this to work we would need the pull request (or something similar) merged in a4

phillimorland (approved):
Cool!

PR #3664, opened 2021-06-28
apps/poll: fetching list via REST API for userview

khamui:


khamui:
@phillimorland this is the super WIP state, but we can merge this in order to migrate all things into a4 :) thanks!!

meinberlin/apps/polls/assets/PollQuestions.jsx
```diff
@@ -0,0 +1,19 @@
+import React, { useState, useEffect } from 'react'
+import PollQuestion from './PollQuestion'
+
+const api = require('adhocracy4').api
+
+export const PollQuestions = (props) => {
+  const [questions, setQuestions] = useState([])
+
+  useEffect(() => {
+    api.poll.get(props.pollId)
```
Rineee, line 10: Its telling me props.poll is not defined when saving a question, so I think you also need to adjust the api.poll.change call above?

Rineee:
Its looking good, but now all the counts are missing, because the questions are not annotated .. should I look into that?

meinberlin/apps/polls/assets/PollQuestions.jsx
khamui: oh thanks, good catch!

You mean the one in `EditPollQuestions.jsx` right? This part (userview) can be buggy, because its very WIP.

Rineee: Oh yes, commented on the wrong file, sorry..

khamui:
> Its looking good, but now all the counts are missing, because the questions are not annotated .. should I look into that?

@Rineee do you mean the results page? the userview is a very WIP state, i just ask for this PR to move all stuff into a4 and continue from there

Rineee:
> > Its looking good, but now all the counts are missing, because the questions are not annotated .. should I look into that?
> 
> @Rineee do you mean the results page? the userview is a very WIP state, i just ask for this PR to move all stuff into a4 and continue from there

Yes, I mean in the userview, but its when you look at the fetched data, the questions all dont have a count (or a count -1), this has to fixed in Django..

phillimorland:
closed in favour of https://github.com/liqd/adhocracy4/pull/749

PR #3691, opened 2021-07-05
apps/polls: add poll export

Rineee:


meinberlin/apps/polls/exports.py
```diff
@@ -40,3 +42,93 @@ def get_virtual_fields(self, virtual):
...
+
+class PollExportView(export_views.BaseItemExportView):
+
+    def get_queryset(self):
+        creators_vote = poll_models.Vote.objects.filter(
+            choice__question__poll=self.poll).values_list('creator', flat=True)
+        creators_answer = poll_models.Answer.objects.filter(
+            question__poll=self.poll).values_list('creator', flat=True)
+        creator_ids = list(set(creators_vote).union(set(creators_answer)))
+        return User.objects.filter(pk__in=creator_ids)
```
Rineee, line 55: @fuzzylogic2000 Because of the User model, we would need to keep this here, right? But all of the rest, I could move into a mixin and add to A4 instead, what do you think?

fuzzylogic2000 (approved):
Yeah! It reads very well! I can't test the actual export, because I do not have polls that have any answers locally... But the virtual fields all get there perfectly!

Do we merge and test on dev before moving to a4?

meinberlin/apps/polls/exports.py
```diff
@@ -40,3 +42,93 @@ def get_virtual_fields(self, virtual):
...
+
+
+class PollExportView(export_views.BaseItemExportView):
+
+    def get_queryset(self):
+        creators_vote = poll_models.Vote.objects.filter(
+            choice__question__poll=self.poll).values_list('creator', flat=True)
+        creators_answer = poll_models.Answer.objects.filter(
+            question__poll=self.poll).values_list('creator', flat=True)
+        creator_ids = list(set(creators_vote).union(set(creators_answer)))
```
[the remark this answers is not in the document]
fuzzylogic2000: And I also like how you build all the dynamic virtual fields!

meinberlin/apps/polls/exports.py
Rineee: Ah, cool! So we would then just leave an empty view in mB and the rest can go to A4?

Rineee:
> Yeah! It reads very well! I can't test the actual export, because I do not have polls that have any answers locally... But the virtual fields all get there perfectly!
> 
> Do we merge and test on dev before moving to a4?

Yes, we can do that! Or move it directly, as you prefer..

PR #3711, opened 2021-07-08
requirements/base: psycopg2 >= 2.9 is incompatible with django 2.2.x

goapunk:


renovate.json
```diff
@@ -24,6 +24,10 @@
       {
         "packagePatterns": ["^eslint"],
         "groupName": "eslint packages"
+      },
+      {
+       "matchPackageNames": ["psycopg2-binary"],
+       "allowedVersions": "<2.9"
```
fuzzylogic2000, line 30: Do we need this? Is it also js and shouldn't be updated by renovate?

goapunk: I'm not 100% sure to be honest, but on bplan renovate had a MR for it, renovate does python dependencies ? or at least sometimes it does, there doesn't seem to be a clear line

goapunk: hmm, I guess you are right, maybe it was a bug in bplan, all the other repos only get js updates via renovate. I'll remove this.

fuzzylogic2000 (approved):

PR #3712, opened 2021-07-08
webpack: use new inbuilt asset modules instead of file-loader

goapunk:
recommended with the css-mini-extract-plugin, one external dependency less


goapunk:
This seems to change the behavior a bit: with the loader the font `svg` would end up in both `/static/fonts/` and `/static`images`, for some reason it now will only be the latter. This should be fine though?

fuzzylogic2000 (approved):
Looks fine locally, let's try on dev!

PR #3715, opened 2021-07-08
Kl 2021 07 cleanup poll

fuzzylogic2000:
I also wanted to add tests, but I think that doesn't make a lot of sense.


fuzzylogic2000:
@Rineee Can we merge this? I fear our latest changes are not deployed on dev, because the linter is failing.

PR #3731, opened 2021-07-13
apps/poll/pollresults: styling of user answer highlighting

khamui:


meinberlin/assets/scss/components/_timeline-carousel.scss
```diff
@@ -45,7 +45,6 @@ $timeline-carousel__top-space: 11px;
 
 //slick overwrites
 .slick-slide {
-    margin: 0 3*$em-spacer;
```
phillimorland, line 48: we use this in the project timeline so we can't remove it and not add it back, one way to do it would be to nest these styles within a timeline parent class, I think .timeline_carousel would work

meinberlin/assets/scss/components/_poll.scss
```diff
@@ -311,3 +311,8 @@ $checkbox-size: 20px;
         right: 4*$spacer;
     }
 }
+
+.poll-slider__item--highlight {
+    padding: $spacer;
+    background-color: $brand-primary-tint;
+}
```
phillimorland, line 318: cool that it's working but the design was changed due to a11y, can be seen here: https://app.zeplin.io/project/5b051eef7aaa417d2227c55c/screen/5f466bb2f5b0f43658eac1a3

phillimorland (changes requested):
nice that it's working! just couple comments, also the slide numbers are now in wrong position, not sure where that came from but probably due to padding? maybe you a4 wasn't rebased?

khamui:
> nice that it's working! just couple comments, also the slide numbers are now in wrong position, not sure where that came from but probably due to padding? maybe you a4 wasn't rebased?

hm..i double checked and it was rebased. could it be that on dev the numbers are also in wrong position?

meinberlin/assets/scss/components/_poll.scss
```diff
@@ -296,10 +296,6 @@ $checkbox-size: 20px;
     padding: $spacer;
 }
 
-.poll-slider__answer {
-    padding-bottom: $spacer;
-}
-
```
phillimorland, line 302: this is to stop longer answers from overlapping with the arrowes so should be retained or alternate solution added, currently they do overlap

phillimorland (changes requested):
I checked on dev and the slide count is correctly positioned but it's still positioned behind the arrows here, do you see it differently in your browser?

meinberlin/assets/scss/components/_poll.scss
khamui: oops, i removed the wrong one, sorry! should have removed what you found in the other comment `..--highlight`

meinberlin/assets/scss/components/_timeline-carousel.scss
khamui: hmm..that is strange, let me doublecheck

phillimorland:
slide counter still behind arrows?! and as thats the only thing this pull request seems to be doing now will close

PR #3746, opened 2021-07-19
Pm 2021 07 poll issues

phillimorland:


khamui:
Nice, i checked the styles as good as i could i think. 

Slick slides may be tested better on dev (having only one answer on local, so arrows don't appear here for me).

And also the tabs on mobile should be verified on dev.

Rest is looking good!

PR #3755, opened 2021-07-22
Ks 2021 07 translations

Rineee:
Wasnt sure about translation of error message, but left a note for Caro inside story..


phillimorland (approved):
LGTM

PR #3767, opened 2021-08-05
plans: cost field is required

sabinammm:
US #379 plans: make cost field required
- cost field is obligatory
- empty fields of old plans are filled with "Keine Angabe"


fuzzylogic2000 (commented):
Cool! I think, we can already merge this first before doing the other wording changes for the model fields. The only very small thing left, though: please rename the migration file. (Let me know if you want git-help to get that into this same commit.)

fuzzylogic2000 (approved):
Nice!

PR #3768, opened 2021-08-09
created and modified fields in exported plans

sabinammm:
US #4523 plans: add new info in export

### DIFF
--- a/meinberlin/apps/plans/exports.py
+++ b/meinberlin/apps/plans/exports.py
@@ -22,7 +22,8 @@ class DashboardPlanExportView(a4dashboard_mixins.DashboardBaseMixin,
     permission_required = 'meinberlin_plans.export_plan'
     model = models.Plan
     fields = ['title', 'description', 'contact', 'district', 'topics',
-              'cost', 'duration', 'status', 'participation', 'organisation']
+              'cost', 'duration', 'status', 'participation', 'organisation',
+              'created', 'modified']
     html_fields = ['description']
 
     def get_object_list(self):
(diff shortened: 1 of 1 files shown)


meinberlin/apps/plans/exports.py
```diff
@@ -22,7 +22,8 @@ class DashboardPlanExportView(a4dashboard_mixins.DashboardBaseMixin,
     permission_required = 'meinberlin_plans.export_plan'
     model = models.Plan
     fields = ['title', 'description', 'contact', 'district', 'topics',
-              'cost', 'duration', 'status', 'participation', 'organisation']
+              'cost', 'duration', 'status', 'participation', 'organisation',
+              'created', 'modified']
```
fuzzylogic2000, line 26: I think, we do not need the fields in here, because we do add them to the virtual fields in the `get_virtual_fields`. Also, if we only add them to the virtual fields later, it should be added to the end, which was a NTH (nice to have).

meinberlin/apps/plans/exports.py
```diff
@@ -48,11 +49,19 @@ def get_virtual_fields(self, virtual):
...
         virtual['projects_links'] = ugettext('Project Links')
+        virtual['created'] = ugettext('Created')
+        virtual['modified'] = ugettext('Modified')
         return virtual
 
     def get_organisation_data(self, item):
         return item.organisation.name
 
+    def get_created_data(self, item):
+        return item.created.strftime('%X %x')
```
fuzzylogic2000, line 60: I like the formatting of the time. It's much more readable then the isoformat.

meinberlin/apps/plans/exports.py
```diff
@@ -48,11 +49,19 @@ def get_virtual_fields(self, virtual):
...
         virtual['projects'] = ugettext('Projects')
         virtual['projects_links'] = ugettext('Project Links')
+        virtual['created'] = ugettext('Created')
+        virtual['modified'] = ugettext('Modified')
         return virtual
 
     def get_organisation_data(self, item):
         return item.organisation.name
 
+    def get_created_data(self, item):
```
fuzzylogic2000, line 59: If we add the fields to the end of the virtual fields, it also makes more sense to me to add these methods to the end of the method list. Even though, it doesn't change anything about the function of it. So, if you think, they make more sense here, leave them here.

fuzzylogic2000 (changes requested):
I think the small change would add the fields to the end of the list.

fuzzylogic2000 (approved):
Nice!

PR #3785, opened 2021-08-17
Kl 2021 08 plan export test

fuzzylogic2000:
Maybe there are too many tests in the test, because some of it should be tested in a4. But with these we do also notice if something breaks there!


sabinammm (approved):
nice! We needed this!

PR #3786, opened 2021-08-18
apps/plans adapt contact fields

sabinammm:
Though the contact fields have been successfully adapted and the plan detail page now displays the information, almost as depicted in the mock-ups, there are a couple of things that still need to be done:

- [ ] the template for the detail view has to be adjusted to look symmetrical, please help @phillimorland 
- [x] the export now fails because the data fields have changed. This needs to be fixed. 
- [ ] apparently, the designs will change a little bit, so this should probably not be merged before these are final and the story is officially picked up.


meinberlin/apps/plans/models.py
```diff
@@ -16,9 +16,10 @@
 from adhocracy4.projects import models as project_models
 from adhocracy4.projects.enums import Access
 from adhocracy4.projects.fields import TopicField
+from adhocracy4.projects.models import ProjectContactDetailMixin
 
 
-class Plan(UserGeneratedContentModel):
+class Plan(ProjectContactDetailMixin, UserGeneratedContentModel):
```
fuzzylogic2000, line 22: when using the mixin, the old field needs to be removed from the model (and will be replaced by the new field, I think). Somehow the migration is right and only adds fields, though.

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -97,23 +96,66 @@ <h1 class="item-detail__title u-break-word">
                     </div>
                 </div>
 
-                <h3 class="item-detail-2__smalltitle">{% trans 'Contact' %}</h3>
-                {{ object.contact|linebreaks }}
+                <div style="float: left; width: 50%;">
```
fuzzylogic2000, line 99: Ah yes, you were right, we don't put styles into the html. We have classes in the scss components you can either already use or if we need new styles they have to be added there. Let's do that together tomorrow!

meinberlin/apps/plans/views.py
```diff
@@ -35,7 +35,7 @@ class PlanListView(rules_mixins.PermissionRequiredMixin,
     permission_required = 'meinberlin_plans.list_plan'
 
     def get_queryset(self):
-        return super().get_queryset()\
+        return super().get_queryset() \
```
fuzzylogic2000, line 38: There shouldn't be a blank inbetween, because it all goes together `super().get_queryset().select_related()`
Same below in lines 50-52

fuzzylogic2000 (commented):
I don't know if it's git or the commits. Something is weird here. The first three are shown two times and then when I go through the code commit for commit, the tests are missing. Hmm.

fuzzylogic2000:
Ok, locally the commits are fine! It's just github being a bit weird

sabinammm:
Since the commits are in disarray, this is replaced by two new branches that will be merged instead

PR #3791, opened 2021-08-18
apps/polls: also fill weight field for choices

fuzzylogic2000:
The order of the choices stays the same like that.
We have to change it in this migration, because we cannot make sure, that this is run before
the migration in a4 adding the weight to the poll. And for projects where this migration is already run, it is no problem that it changed.

I also got the error locally when I used an older database. This fixed it and the order of the choices stayed the same.


goapunk (approved):
nice fix!

Rineee (approved):
Yeah, cool!

PR #3794, opened 2021-08-19
Sm 2021 08 plans contact fields

fuzzylogic2000:
adds the new contact fields for model and forms

PR #3795, opened 2021-08-19
Sm 2021 08 style plan detail contact

fuzzylogic2000:
depends on https://github.com/liqd/a4-meinberlin/pull/3794 and has to be rebased after that is merged


sabinammm (approved):
Looks perfect and tested great. Thank you :)

PR #3796, opened 2021-08-19
apps/plans: add free text filter for dashboard list

fuzzylogic2000:
I put the small change into the second commit, so that @sabinammm can look at the first commit to also add the filter to the Bplans and external projects.


sabinammm (approved):
Thanks for the how-to:) Works perfectly, will take care of the bplans and external projects.

fuzzylogic2000:
@sabinammm Do you want to use this branch? Maybe it's better to merge this first and then open a new branch, so that I can review what you did?!

PR #3797, opened 2021-08-19
apps/plans style form

sabinammm:
Unclear and need to continue on the following

- [x] the handling of translations
- [x] spacing under and styling of new descriptions
- [x] can't seem to get rid of the nav-bar 
-@fuzzylogic2000 can you help?


fuzzylogic2000:
@sabinammm You don't need to worry about the images at all for the story you are working on, so you can remove the first and third task here.
For 2.: You already marked all the strings a translatable with the "trans"-tag, so that's already fine! The actual translations can be done later together with the ones from the other stories. (I will add a task to the story, though.)

PR #3800, opened 2021-08-23
apps/plans: additional contact fields in export

sabinammm:


fuzzylogic2000 (commented):
Looks good! You'll also have to add to the test, what it should be exported as, though. Like `assert plan.contact_name == export get_field_data(plan, 'contact_name')`. Does that work? And maybe test it for one plan with (some) empty fields and for the other one with something filled in into all of them. The `<i>...` was for richtextfields, I think. To test that all the tags are not in there anymore.

fuzzylogic2000 (approved):
Cool! I'll merge for the export and we will add to the tests later!

PR #3801, opened 2021-08-23
Ks 2021 08 add tile image to plans

Rineee:
I also added the image copyright, because I thought we need that for all images, right?


fuzzylogic2000 (commented):
Yes, does it! But there was also wording changes for the other image... :( Can we do that in the same migration? And maybe in the model put the new image field directly below the old one?

fuzzylogic2000 (approved):
Nice!

PR #3802, opened 2021-08-23
plans/image: new position below profile info

khamui:


meinberlin/assets/scss/components/_item_detail.scss
```diff
@@ -56,8 +55,12 @@
     background-repeat: no-repeat;
     background-size: cover;
     background-position: center;
-    width: 330px; //size is same as anticipated ratio
-    height: 220px;
+    width: 100%;
+    padding-top: 60%; // this creates an aspect ratio in css, 100% = 1:1 ratio
```
fuzzylogic2000, line 59: Cool!

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -76,19 +76,17 @@ <h1 class="item-detail__title u-break-word">
...
+                    {% if object.description_image %}
                         <div class="item-detail__image--background" style="background-image: url({% thumbnail object.description_image 'item_image' %})">
                         {% if object.description_image_copyright %}
                             <span class="item-detail__copyright copyright">© {{ object.description_image_copyright }}</span>
                         {% endif %}
                         </div>
-                        {% endif %}
-
+                    {% endif %}
+                </div>
```
fuzzylogic2000, line 86: Should this be here? You didn't add an opening div.

meinberlin/assets/scss/components/_item_detail.scss
```diff
@@ -39,8 +39,7 @@
     margin-bottom: 1.5em;
 }
 
-.item-detail__image,
-.item-detail__image--background {
+.item-detail__image {
```
fuzzylogic2000, line 42: I know, you didn't add that, but when we are on it, can we change the naming? `item-detail__image--background` is not a variant, so maybe we call it `item-detail__background-image` and the other one `item-detail__hero-image`? Or only the first, because less changes.

fuzzylogic2000 (commented):
It looks good! Just two comments/questions.

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
khamui: i doublechecked this and saw that this was already there and i just shifted the `tile image` part  right above it.
Don't know why its marked as new/added in the diff.

fuzzylogic2000 (approved):
Nice!

PR #3803, opened 2021-08-23
Sm 2021 08 issue export time difference

### LINKED ISSUE #3793: export for ideas, comments and plans: time difference
**URL:** https://meinberlin-dev.liqd.net/dashboard/organisations/caros-testorga/plans/
**user:** initiators
**expected behaviour:** 
**behaviour:** the time of creation or last change for ideas, comments and plans in the export is 2 hours ahead of the actual time. e.g the plan linked above was created at 16.15, the export says it has has been created at 14.15.
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

sabinammm:
Closes part of #3793, the other one needs to be fixed in a4.

### DIFF
--- a/meinberlin/apps/plans/exports.py
+++ b/meinberlin/apps/plans/exports.py
@@ -88,7 +88,7 @@ class DashboardPlanExportView(a4dashboard_mixins.DashboardBaseMixin,
         return ''
 
     def get_created_data(self, item):
-        return item.created.strftime('%X %x')
+        return item.created.astimezone().isoformat()
 
     def get_modified_data(self, item):
-        return item.created.strftime('%X %x')
+        return item.modified.astimezone().isoformat()
--- a/tests/plans/test_export.py
+++ b/tests/plans/test_export.py
@@ -74,8 +74,10 @@ def test_reply_to_mixin(plan_factory, project_factory,
         )
     assert projects == export.get_projects_data(plan)
 
-    assert plan.created.strftime('%X %x') == export.get_created_data(plan)
-    assert plan.modified.strftime('%X %x') == export.get_modified_data(plan)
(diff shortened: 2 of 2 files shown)


sabinammm:
The first commit in this PR is wrong and needs to be deleted. I pushed it by accident.

fuzzylogic2000 (commented):
> The first commit in this PR is wrong and needs to be deleted. I pushed it by accident. I attempted to revert it, but now I have an extra commit (number 4) instead....

Yes, I was wondering about that. Try removing the last commit with `git reset HEAD~1` and then deleting the changes. You could also do `git reset HEAD~1 --hard` to also get rid of the changes in one step, but I find `--hard` too scary.
And then you can do the `git rebase main -i` thing and drop the first commit instead of reverting it. And then only the two middle ones should be left.

fuzzylogic2000:
Ok, actually doing the rebase-thing and just dropping both commits should also be fine...

fuzzylogic2000 (commented):
Okeyyyyy, thinking about it again, I think we cannot export a time and not say which timezone it is in. On the website the time changes depend on the browser settings, but exporting it to an excel without saying the timezone is just wrong. I will discuss in the issue and check if there is a way to display the isformat with the offset, so that it looks like "our time", but still has the info attached.

fuzzylogic2000:
Well, and that also wasn't quite true. While we use the user's timezone in OPIN, in mB we use the one set in the settings for everyone, I think...

fuzzylogic2000:
Alright, I think we should use this:
```
>>> Plan.objects.first().created.astimezone().isoformat()
'2021-08-23T17:29:21.248264+02:00'
```

fuzzylogic2000 (approved):
yeah!

PR #3805, opened 2021-08-23
apps/plans: add tile image to API

fuzzylogic2000:
This adds the tile-image and its copyright if it exists, otherwise it adds the image used on the detail page.
This is the same as for the projects.


khamui:
LGTM! nice

PR #3817, opened 2021-08-26
dev: add option to use postgresql in dev environment

goapunk:


fuzzylogic2000 (commented):
I didn't get it to work. Maybe because I already had too much of a postgresql DB running? Can we have a look together (next week)?

PR #3818, opened 2021-08-27
templates/a4dashboard: deleting project info form with current a4 era…

sabinammm:
…ses titels from form

depending on https://github.com/liqd/adhocracy4/pull/821, merge with a4 update, which replaces the deleted file.


fuzzylogic2000 (approved):
Nice!

PR #3824, opened 2021-08-30
apps/plans: check if item has been modified added

sabinammm:


fuzzylogic2000 (commented):
Nice! 
Could you also write a test for that? (Put `modified=None` into one of the plans in the test.) 
And did you check the exports from a4, that they also check if it's there?

sabinammm:
> Nice!
> Could you also write a test for that? (Put `modified=None` into one of the plans in the test.)
> And did you check the exports from a4, that they also check if it's there?

Can we talk about the tests when you get the chance? I'm not getting along with these in general. 

None of the other exports use "modified", only "created" and since this can never be empty, is any check necessary or should we also add "modified"?

PR #3825, opened 2021-08-30
changed contact address form field

sabinammm:
@fuzzylogic2000 thanks for the help. I'm still confused about the dashboard vs other forms. Maybe we can talk about it when you get the chance.


fuzzylogic2000 (approved):
Nice! Yes, let's talk after standup!

PR #3827, opened 2021-08-30
Sm 2021 08 test plan export contacts

sabinammm:


tests/plans/test_export.py
```diff
@@ -87,11 +91,11 @@ def test_reply_to_mixin(plan_factory, project_factory,
     project_1 = project_factory()
     project_2 = project_factory()
     plan = plan_factory(
-        contact_name='',
+        contact_name='Joe',
```
fuzzylogic2000, line 94: Jane! :)

fuzzylogic2000: Or actually this is the place to put in your name to leave some marks on the software. :)

fuzzylogic2000 (approved):
Nice!

tests/plans/test_export.py
sabinammm: Shoot! Next time :)

PR #3829, opened 2021-08-30
maps/menu: three dots menu above maps

### LINKED ISSUE #3819: mapideas: menu dropdown behind map
**URL:** https://meinberlin-dev.liqd.net/budgeting/2021-00556/
**user:** 
**expected behaviour:** the menu is visible
**behaviour:** menu is behind the map
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
image

khamui:
close #3819

### DIFF
--- a/meinberlin/assets/scss/components/_item_detail.scss
+++ b/meinberlin/assets/scss/components/_item_detail.scss
@@ -30,6 +30,8 @@
 }
 
 .item-detail__content {
+    position: relative;
+    z-index: 0;
     margin-bottom: 1em;
 }
 


fuzzylogic2000 (commented):
It looks good to me. But @phillimorland has to approve as she knows more of what we do with the maps and the z-indexes.

phillimorland:
eeeesh, I kind of know :p So I think this is fine, the issues that causes the behaviour is from all our dropdown overwrites but that would have too many repercussions so I think this is all good. 
For other cases I would try to strip away the code causing the issue rather then adding a z-index as they can cause unexpected issues but only if it doesn't mean implementing lots of fixes :)

PR #3835, opened 2021-08-31
apps/plans/templates: use more of project progess line for plans dash…

fuzzylogic2000:
…board create and update

After talking to @mcastro-lqd , we came up with this solution, which is a bit closer to the project dashboard.

### DIFF
--- a/meinberlin/apps/plans/templates/meinberlin_plans/plan_create_dashboard.html
+++ b/meinberlin/apps/plans/templates/meinberlin_plans/plan_create_dashboard.html
@@ -30,9 +30,18 @@
 
     <div class="l-menu__aside">
         <ul class="progress-line u-spacer-bottom">
-            <li class="progress-line__entry--published">
-                <i class="fas fa-circle progress-line__icon" title="{% trans 'The preview is activated after the first save.' %}" aria-label="{% trans 'The preview is activated after the first save.' %}"></i>
-                <span>{% trans 'The preview is activated after the first save.' %}</span>
+            <li class="progress-line__entry">
+                <span class="progress-line__line--bottom"></span>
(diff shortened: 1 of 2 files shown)


meinberlin/apps/plans/templates/meinberlin_plans/plan_create_dashboard.html
```diff
@@ -30,9 +30,18 @@ <h1 class="u-first-heading">{% trans 'Add a new plan' %}</h1>
 
     <div class="l-menu__aside">
         <ul class="progress-line u-spacer-bottom">
-            <li class="progress-line__entry--published">
-                <i class="fas fa-circle progress-line__icon" title="{% trans 'The preview is activated after the first save.' %}" aria-label="{% trans 'The preview is activated after the first save.' %}"></i>
-                <span>{% trans 'The preview is activated after the first save.' %}</span>
+            <li class="progress-line__entry">
+                <span class="progress-line__line--bottom"></span>
+                <i class="fas fa-circle progress-line__icon--active" title="{% trans 'Create plan' %}" aria-label="{% trans 'Create plan' %}"></i>
```
phillimorland, line 35: Just from a UX point of view I would maybe use `fa-check-circle u-success` here and have `progress-line__icon--active` on the 'Fill in plan details' step as the plan has already been created but I think we merge and check with @mcastro-lqd

phillimorland (approved):
looking good!

meinberlin/apps/plans/templates/meinberlin_plans/plan_create_dashboard.html
fuzzylogic2000: Yeah, the problem here is that it isn't created. If you leave the page without saving the plan is gone.

fuzzylogic2000: @phillimorland That answer :point_up:  was for you, not sure you got it without me tagging you?

PR #3839, opened 2021-09-01
apps/polls: delete models and tables

### LINKED ISSUE #3686: delete old poll tables (one deployment after release v2107)
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** We copied the content of the mB poll tables to the a4 poll tables and are now using the tables from a4. If everything went well with the deployment, we should drop the old mB poll tables.

Screenshot?

fuzzylogic2000:
The poll models and tables were migrated to a4 in the last release.
This deletes the models and tables that are not used anymore.

fixes #3686 

It would be great if you could all review and try it out on your local db (after copying it to make sure it's not lost). But if it works, let me merge and deploy on dev, so I can look at what happens and make sure all is fine.


sabinammm (approved):
I tested this locally and everything was fine and the same as before! So, lovely:)

fuzzylogic2000:
I will try it on dev!

PR #3841, opened 2021-09-01
apps/module dashboards: no export in nav before module is added to pr…

### LINKED ISSUE #3837: dashboard: export nav is shown when project is published, but module isn't
**URL:** https://meinberlin-dev.liqd.net/dashboard/projects/this-tests-two-polls-in-one-project/basic/
**user:** 
**expected behaviour:** The export is only shown when project and module are published.
**behaviour:** For every module the export nav is shown as soon as the project is published. That makes sense when a module is published (added to the project), but is a bit weird when it isn't. 
**important screensize:**
**device & browser:** 
**Comment/Question:** This is set in the `is_effective` of the dashboard module components. For example here: https://github.com/liqd/a4-meinberlin/blob/ed1b93cfb218a2b5900750f16d43fb38da02e753/meinberlin/apps/maptopicprio/dashboard.py#L61 But needs to be changes in all the export components when decided upon.

Screenshot?

sabinammm:
…oject
Closes #3837

### DIFF
--- a/meinberlin/apps/budgeting/dashboard.py
+++ b/meinberlin/apps/budgeting/dashboard.py
@@ -16,7 +16,7 @@ class ExportBudgetingComponent(DashboardComponent):
     def is_effective(self, module):
         module_app = module.phases[0].content().app
         return (module_app == 'meinberlin_budgeting' and
-                not module.project.is_draft)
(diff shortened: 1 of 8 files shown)


fuzzylogic2000 (approved):
very nice!

PR #3842, opened 2021-09-01
apps/plans/plan_update_form: move modal within content tag so it can …

phillimorland:
…show, add additional dialouge for screen reader

### DIFF
--- a/meinberlin/apps/plans/templates/meinberlin_plans/plan_update_form.html
+++ b/meinberlin/apps/plans/templates/meinberlin_plans/plan_update_form.html
@@ -34,7 +34,7 @@
             <ul class="progress-line u-spacer-bottom">
                 <li class="progress-line__entry">
                     <span class="progress-line__line--bottom"></span>
-                    <i class="fas fa-check-circle u-success progress-line__icon" title="{% trans 'Create plan' %}" aria-label="{% trans 'Create plan' %}"></i>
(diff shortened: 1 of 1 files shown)


meinberlin/apps/plans/templates/meinberlin_plans/plan_update_form.html
```diff
@@ -84,9 +80,8 @@ <h1 class="u-first-heading">{% trans 'Edit plan' %}</h1>
         {% endif %}
     </div>
 </div>
-{% endblock %}
 
-<div class="modal" tabindex="-1" role="dialog" id="plan-publish-modal" aria-labelledby="plan-publish-label">
+<div class="modal" tabindex="-1" role="dialog" id="plan-publish-modal" aria-labelledby="plan-publish-label" aria-hidden="true">
```
phillimorland, line 84: this is got from the docs, I assume it's incase someones screen reader is just reading everything weather it's open or not? it works fine anyway

fuzzylogic2000 (approved):
Awesome! 
And :rofl: that I didn't see that and spent about half an hour checking if the js is there. :roll_eyes:

meinberlin/apps/plans/templates/meinberlin_plans/plan_update_form.html
fuzzylogic2000: Sorry, only registered with me when rereading in my mails. Should we check the publish button for the projects for a11y as well?

PR #3857, opened 2021-09-07
dev: add option to use postgresql in dev environment

phillimorland:


phillimorland:
@goapunk @fuzzylogic2000 sorry guys, was on auto pilor this morning and was on wrong pull request when merged, have reverted but it when restored this didn't open so just opened a new one, here is link https://github.com/liqd/a4-meinberlin/pull/3817

goapunk:
@fuzzylogic2000 ping me if you want to test this again :)

goapunk:
I updated this so it's using a different port for each project so it should work with multiple instances now

README.md
```diff
@@ -22,3 +22,15 @@ based on [adhocracy 4](https://github.com/liqd/adhocracy4).
...
+### Use postgresql database for testing:
+run the following command once:
+```
+make create-postgres
+```
+to start the testserver with postgresql, run:
+```
+export DATABASE=postgresql
+make start-postgres
+make watch
```
sabinammm, line 35: I had to also do 
`python manage.py migrate`
Without it, I get `ProgrammingError: relation "django_session" does not exist`. 
Maybe add this here?

goapunk: that should be solved if you do `make install`, that shouldn't have to go here

sabinammm: Oh, sorry, my mistake - I did it in the order in the readme.

fuzzylogic2000 (approved):
So cool!

PR #3860, opened 2021-09-08
Kl 2021 09 more publish tests

fuzzylogic2000:
How to do the delete stuff, I found here: https://stackoverflow.com/questions/54467081/how-to-send-error-message-from-django-deleteview
To test the delete locally on the site instead of in the tests, change the [if around the module delete button](https://github.com/liqd/a4-meinberlin/blob/488edf5ecb1bee5be7067eb49211bf8a6277ef52/meinberlin/templates/a4dashboard/includes/nav_modules_item.html#L54) to `{% if True %}`

### DIFF
--- a/meinberlin/apps/dashboard/views.py
+++ b/meinberlin/apps/dashboard/views.py
@@ -205,6 +205,23 @@ class ModuleDeleteView(PermissionRequiredMixin,
     success_message = _('The module has been deleted')
 
     def delete(self, request, *args, **kwargs):
+        module = self.get_object()
+        # only unpublished modules can be deleted
+        if not module.is_draft:
+            messages.error(request,
+                           _('Module added to a project cannot be '
+                             'deleted.'))
+            # try to stay on same page
+            if 'referrer' in request.POST:
+                redirect = request.POST['referrer']
+            elif 'HTTP_REFERER' in self.request.META:
+                redirect = request.META['HTTP_REFERER']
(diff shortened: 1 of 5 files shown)


meinberlin/apps/dashboard/views.py
```diff
@@ -205,6 +205,23 @@ class ModuleDeleteView(PermissionRequiredMixin,
...
+                           _('Module added to a project cannot be '
+                             'deleted.'))
+            # try to stay on same page
+            if 'referrer' in request.POST:
+                redirect = request.POST['referrer']
+            elif 'HTTP_REFERER' in self.request.META:
+                redirect = request.META['HTTP_REFERER']
+            else:
+                redirect = reverse('a4dashboard:project-edit', kwargs={
+                    'project_slug': module.project.slug
```
fuzzylogic2000, line 221: I am a bit unhappy about putting all of that here. But the referrer stuff in the success_url is a bit different, because we cannot refer back to the module forms, when the module was deleted. And as long as we only have one thing to test in the delete, maybe it's fine to keep it here?

fuzzylogic2000:
@sabinammm When you tried this and it all still works fine, you can merge, please! And maybe @Rineee can also have a look next week. :)

sabinammm (approved):
I tested the tests and the error messages. Work perfectly!! I will merge.

PR #3870, opened 2021-09-16
apps/polls: make sure actions have the new a4 poll as content type

fuzzylogic2000:
We only noticed that the actions still have the wrong content type of the old polls when Caro tried to delete multiple projects at once: https://sentry.liqd.net/organizations/liqd/issues/1609/?project=2
The migration is like we migrated the content type of the comments: https://github.com/liqd/a4-meinberlin/blob/a0718a4d16fdf3c250feb1f8d805d073e7460f52/meinberlin/apps/polls/migrations/0006_copy_poll_data_to_a4_polls.py#L43


Rineee (approved):
Yes, really nice!

PR #3871, opened 2021-09-20
plans/models: update save function for richtext added content to allo…

### LINKED ISSUE #3821: testing 4543: changing picture size in CKeditor doesn't effect frontend
**URL:** https://meinberlin-dev.liqd.net/vorhaben/2021-00074/
**user:** any
**expected behaviour:** if I change the size of my picture in CKeditor the size is also changed in frontend
**behaviour:** picture always has the same size, changes are not saved
**important screensize:**
**device & browser:** chrome, big sur
**Comment/Question:** Is that on purpose because we want the plan to look nice and not have screwed up pictures? I think we should allow it as we also do that for projects.

Screenshot?

phillimorland:
…w style configs tobe passed fixes #3821

### DIFF
--- a/meinberlin/apps/plans/models.py
+++ b/meinberlin/apps/plans/models.py
@@ -214,7 +214,8 @@ class Plan(ProjectContactDetailMixin, UserGeneratedContentModel):
                                    year=self.created.year))
 
     def save(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


fuzzylogic2000 (approved):
Yeah!

PR #3879, opened 2021-09-21
config/settings/base.py: rm embed from config of default ckeditor

### LINKED ISSUE #3878: users should not be able to embed video in their idea
**URL:** https://meinberlin-stage.liqd.net/mapideas/create/module/brainstorming-mit-kartenfunktion-36/
**user:** registered user
**expected behaviour:** should not be able to embed video
**behaviour:** is able to embed video in idea form
**important screensize:**
**device & browser:** 
**Comment/Question:** we should not allow this also because it may look crap in frontend if also a picture has been uploaded. Don't know where this came from but it is not on prod.

Screenshot?
Bildschirmfoto 2021-09-21 um 16 03 08

Bildschirmfoto 2021-09-21 um 16 05 47

phillimorland:
fixes #3878

### DIFF
--- a/meinberlin/config/settings/base.py
+++ b/meinberlin/config/settings/base.py
@@ -312,7 +312,6 @@ CKEDITOR_CONFIGS = {
             ['Bold', 'Italic', 'Underline'],
             ['NumberedList', 'BulletedList'],
             ['Link', 'Unlink'],
-            ['Embed', 'EmbedBase']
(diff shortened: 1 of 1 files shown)


phillimorland:
@Rineee this is working locally

fuzzylogic2000 (approved):
Yes, works! Cool!

PR #3882, opened 2021-09-23
Ks 2021 09 move poll export to a4

Rineee:
This only works with https://github.com/liqd/adhocracy4/pull/845

Do not merge until https://github.com/liqd/adhocracy4/pull/845 is merged and A4 dependency updated here.


fuzzylogic2000 (approved):
Yeah! Works! (But I know, wait for a4. :)

PR #3883, opened 2021-09-24
apps/cms/storefront: limit choices of projects to active and future p…

fuzzylogic2000:
…rojects

@sabinammm If it also works for you, you can merge!


sabinammm:
I tested it using a DB that worked on the main branch. It showed me all my projects in wagtail. As soon as I checked out this branch, I only saw 1 project 3 times:
image

When I select it once and save, I get the error page "get() returned more than one Project -- it returned 3".

fuzzylogic2000:
> I tested it using a DB that worked on the main branch. It showed me all my projects in wagtail. As soon as I checked out this branch, I only saw 1 project 3 times:
> image
> 
> When I select it once and save, I get the error page "get() returned more than one Project -- it returned 3".
> 
> I think someone else needs to test it, though.

No, I think it needs to be distinct and I have to go back and fix it! It's a bit weird that I saw my projects only once, though. Maybe they all had only one phase? I'll check it!

fuzzylogic2000:
@sabinammm So, my projects all only had one active module, not multiple ones running at the same time... Very good that you found that!
To make it distinct, I used what you already had found: https://stackoverflow.com/questions/42572095/using-or-q-objects-in-limit-choices-to-in-wagtailadmin
Could you check it again?

sabinammm:
Yes! It works now!! It still needs testing, right? Should I merge now anyway?

fuzzylogic2000:
> Yes! It works now!! It still needs testing, right? Should I merge now anyway?

Yes, I will add a a task to add tests. :) So, you can merge! @sabinammm

PR #3890, opened 2021-09-29
Ks 2021 09 update poll tests

Rineee:


fuzzylogic2000 (approved):
So, this doesn't test the actual export anymore, but the changes export permissions that are different here, right? Nice!

Rineee:
> So, this doesn't test the actual export anymore, but the changes export permissions that are different here, right? Nice!

Yes, exactly!

PR #3945, opened 2021-11-08
update django to 3.2

fuzzylogic2000:


phillimorland (approved):
looking good and passing all the things :)

PR #3953, opened 2021-11-10
meinberlin/apps/dashboard: added new participatory budget to dashboard

sabinammm:
@Rineee 
This now contains the new module in the blueprints, but is still missing the right icon, which I think we can add later, right?

### DIFF
--- a/meinberlin/apps/budgeting/phases.py
+++ b/meinberlin/apps/budgeting/phases.py
@@ -55,6 +55,22 @@ class RatingPhase(phases.PhaseContent):
     }
 
 
+class VotingPhase(phases.PhaseContent):
+    app = apps.Config.label
+    phase = 'voting'
+    view = views.ProposalListView
+
+    name = _('How do you like the submitted proposals?')
+    description = _('Vote for up to 5 proposals for the participatory '
+                    'budget.')
+    module_name = _('participatory budgeting 3 phases')
+
+    features = {
+        'vote': (models.Proposal,)
+    }
+
+
 phases.content.register(RequestPhase())
 phases.content.register(CollectPhase())
 phases.content.register(RatingPhase())
+phases.content.register(VotingPhase())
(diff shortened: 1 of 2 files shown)


meinberlin/apps/budgeting/predicates.py
```diff
@@ -0,0 +1,10 @@
+import rules
+
+from adhocracy4.phases.predicates import has_feature_active
+
+
+@rules.predicate
+def phase_allows_vote(user, item):
+    if item:
+        return has_feature_active(item.module, item.__class__, 'vote')
+    return False
```
Rineee, line 10: Maybe move this to A4 to all other phase predicates? But on the other hand we only use that here .. @fuzzylogic2000 ?

Rineee:
Wording and icoon is requested, but I'd say we can also change that afterwards..

meinberlin/apps/budgeting/phases.py
```diff
@@ -55,6 +55,21 @@ class RatingPhase(phases.PhaseContent):
...
+    module_name = _('participatory budgeting 3 phases')
+
+    features = {
+        'vote': (models.Proposal,)
+    }
+
+
 phases.content.register(RequestPhase())
 phases.content.register(CollectPhase())
 phases.content.register(RatingPhase())
```
sabinammm, line 75: @Rineee I checked it out and it looks wonderful in the dashboard. So I'm wondering, why all the other phases are registered here and the VotingPhase isn't. What shouldn't be working then?

meinberlin/apps/budgeting/predicates.py
fuzzylogic2000: Yes, it should be moved, I think.

meinberlin/apps/budgeting/phases.py
fuzzylogic2000: Good catch. Saving the phases in the dashboard didn't work without it.

fuzzylogic2000 (approved):
Nice! I will add the phase.content.register.

fuzzylogic2000 (approved):
Looking good!

PR #3954, opened 2021-11-10
[#4987]budgeting: add proposal serializer and api

Rineee:
The url to get the proposals is `reverse('proposals-list', kwargs={'module_pk': $module_pk})` or `/api/modules/$module_pk/proposals/` which we then probably need to add to context of ProposalListView

### DIFF
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -1,0 +1,29 @@
+from rest_framework import mixins
+from rest_framework import viewsets
+
+from adhocracy4.api.mixins import ModuleMixin
+from adhocracy4.api.permissions import ViewSetRulesPermission
+
+from .models import Proposal
+from .serializers import ProposalSerializer
+
+
+class ProposalViewSet(ModuleMixin,
+                      mixins.ListModelMixin,
+                      viewsets.GenericViewSet,
+                      ):
+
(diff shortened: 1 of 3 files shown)


phillimorland:
>  `/api/modules/$module_pk/proposals/`

dumb question: how to find a module_pk?

phillimorland (commented):
I am not getting a list of my ideas, the api does have an endpoint (think thats the right word) but the list is empty despite having 3 ideas added, I feel this is not what was intended?

Rineee:
> I am not getting a list of my ideas, the api does have an endpoint (think thats the right word) but the list is empty despite having 3 ideas added, I feel this is not what was intended?

Ok weird, it does work for me, I just checked again.. but you are looking at a Bürgerhaushalt, right? Because it is only showing proposals, not ideas..

phillimorland:
> > I am not getting a list of my ideas, the api does have an endpoint (think thats the right word) but the list is empty despite having 3 ideas added, I feel this is not what was intended?
> 
> @phillimorland Ok weird, it does work for me, I just checked again.. but you are looking at a Bürgerhaushalt, right? Because it is only showing proposals, not ideas..

phillimorland (approved):
sorry closed by mistake.
Yep works when I have the correct PK, sorry! and yey perfect!

PR #3956, opened 2021-11-11
budgeting/proposals/list: add react list

khamui:
To mount the react list, please add the following lines to `meinberlin_budgeting/proposal_list.html`:

`{% load react_proposals %}`

`{% react_proposals view.module %}`

### DIFF
--- a/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
+++ b/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
@@ -1,0 +1,36 @@
+import React from 'react'
+import { toDate } from './helpers'
+
+export const BudgetingProposalListItem = (props) => {
+  const { proposal } = props
+  return (
+    <li className="list-item no-hover">
+      <h3 className="list-item__title">
+        <a href={proposal.url}>
+          {proposal.name}
+        </a>
+      </h3>
+      <div className="list-item__labels">
+        {proposal.category &&
+          <span className="label label--big">
+            {proposal.category.name}
+          </span>}
+        {proposal.budget > 0 &&
+          <span className="label label--big">
+            {proposal.budget} €
+          </span>}
+      </div>
+      <div className="list-item__vote">
+        <div>
+          <span className="list-item__author">
+            {proposal.creator}
+          </span>
+          {toDate(proposal.created)}
+        </div>
+        <button className="btn btn--light">
+          meine Stimme abgeben
+        </button>
+      </div>
+    </li>
+  )
+}
--- a/meinberlin/apps/budgeting/assets/helpers.js
+++ b/meinberlin/apps/budgeting/assets/helpers.js
@@ -1,0 +1,20 @@
+import django from 'django'
+
+const months = [
+  django.gettext('January'),
+  django.gettext('February'),
+  django.gettext('March'),
+  django.gettext('April'),
+  django.gettext('May'),
+  django.gettext('June'),
+  django.gettext('July'),
+  django.gettext('August'),
+  django.gettext('September'),
+  django.gettext('October'),
+  django.gettext('November'),
+  django.gettext('December')
+]
+export const toDate = (isodate) => {
+  const d = new Date(isodate)
+  return `${d.getDate()}. ${months[d.getMonth()]} ${d.getFullYear()}`
+}
--- a/meinberlin/apps/budgeting/templatetags/react_proposals.py
+++ b/meinberlin/apps/budgeting/templatetags/react_proposals.py
@@ -1,0 +1,23 @@
+import json
+
+from django import template
+from django.urls import reverse
+from django.utils.html import format_html
+
+# from django.urls import reverse
+# from django.utils.translation import gettext_lazy as _
+# from adhocracy4.rules.discovery import NormalUser
+
+register = template.Library()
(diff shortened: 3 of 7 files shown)


meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
```diff
@@ -0,0 +1,36 @@
...
+      </div>
+      <div className="list-item__vote">
+        <div>
+          <span className="list-item__author">
+            {proposal.creator}
+          </span>
+          {toDate(proposal.created)}
+        </div>
+        <button className="btn btn--light">
+          meine Stimme abgeben
```
fuzzylogic2000, line 31: We do also translate the stuff in mB. It might get ported to a4 or other platforms.

meinberlin/apps/budgeting/assets/helpers.js
```diff
@@ -0,0 +1,20 @@
...
+  django.gettext('May'),
+  django.gettext('June'),
+  django.gettext('July'),
+  django.gettext('August'),
+  django.gettext('September'),
+  django.gettext('October'),
+  django.gettext('November'),
+  django.gettext('December')
+]
+export const toDate = (isodate) => {
```
fuzzylogic2000, line 17: What is the isodate you use? It should have a timezone/date offset. We should be very careful with that in our js.

meinberlin/assets/scss/components/_list_item.scss
```diff
@@ -15,7 +15,7 @@
         outline: none;
     }
 
-    &:hover {
+    &:hover:not(.no-hover) {
```
fuzzylogic2000, line 18: what does this do?

fuzzylogic2000 (commented):
Looks good! We just also have to think about if we want it for all three "Bürgerhaushalts" and need to only show the button when the phase is active.
The rating and comment count is still missing.

meinberlin/apps/budgeting/assets/helpers.js
fuzzylogic2000: I think we should use get_format? https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#internationalization-in-javascript-code

meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
khamui: ah yes, this is just a "dummy" and should be replaced by the voting component (which @phillimorland started working on?)

meinberlin/apps/budgeting/assets/helpers.js
khamui: we receive a string that looks like this `YYYY-MM-DDTHH:mm:ss.sssZ`, which i think is the ISO notation?
https://262.ecma-international.org/5.1/#sec-15.9.1.15

do you know by heart or have some information to the timezone offset, which is problematic inJS?

Anyways I will try out the get_format from Django.

meinberlin/assets/scss/components/_list_item.scss
khamui: it prevents the tile from having a hover state, which is weird/confusing because if there is a button. we have a doubled hovering effect. E.g. if the mouse is over the tile, it gets darker (the button too).  @mcastro-lqd maybe you have an idea on this?

khamui:
> Looks good! We just also have to think about if we want it for all three "Bürgerhaushalts" and need to only show the button when the phase is active.
> The rating and comment count is still missing.

Yes, i was not sure about that and thought we would just use this for the third phase for now. Lets discuss in standup.

meinberlin/apps/budgeting/templatetags/react_proposals.py
```diff
@@ -0,0 +1,23 @@
+import json
+
+from django import template
+from django.urls import reverse
+from django.utils.html import format_html
+
+# from django.urls import reverse
```
khamui, line 7: tb removed, if unused

meinberlin/apps/budgeting/assets/helpers.js
fuzzylogic2000: Ah, then the timezone-stuff is already taken over by the serializer. :) 
But the get_format should still help to show the date in an internationalized format. (and we should also check that out for the comments...)

meinberlin/assets/scss/components/_list_item.scss
khamui: just checked with @mcastro-lqd and from a design's perspective a hover state without action does not really make sense. so preferably to be removed everywhere. i will do it when coming back to it for "making it possible for all buergerhaushalt variants".

phillimorland: We should make the whole item clickable when it has a hover effect, will add an issue

meinberlin/apps/budgeting/assets/helpers.js
phillimorland: @khamui @fuzzylogic2000 should this befor we merge?

phillimorland (approved):
Looking good! can be merged once time issue resolved

phillimorland (approved):
Time to be sorted in seperate pull request

PR #3973, opened 2021-11-17
Kt 2021 11 pagination

khamui:
Some notes, that might help to understand what was intended :) 
This react pagination should behave like the old one (rather than what is in the current design)
I am not sure if i utilized the pagination django part correctly, any suggestions appreciated.

You might expect it a bit different, but:
page_size = 0 won't show all, but 0 --> edge case (we should use rather a large number)

### DIFF
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -1,8 +1,6 @@
 from rest_framework import mixins
-from rest_framework import status
 from rest_framework import viewsets
 from rest_framework.pagination import PageNumberPagination
-from rest_framework.response import Response
 
 from adhocracy4.api.mixins import ModuleMixin
 from adhocracy4.api.permissions import ViewSetRulesPermission
@@ -16,32 +14,11 @@ from .serializers import ProposalSerializer
 class BudgetPagination(PageNumberPagination):
     page_size = 15
 
-    def get_next_page(self):
-        if self.page.has_next():
-            return self.page.next_page_number()
-        else:
-            return None
-
-    def get_previous_page(self):
-        if self.page.has_previous():
-            return self.page.previous_page_number()
-        else:
-            return None
-
     def get_paginated_response(self, data):
-        return Response({
-            'status': True,
-            'results': data,
-            'meta': {
-                'results_count': self.page.paginator.count,
-                'current_page': self.page.number,
-                'previous_page': self.get_previous_page(),
-                'next_page': self.get_next_page(),
-                'page_count': self.page.paginator.num_pages,
-                'page_size': self.page_size,
-                'is_paginated': self.page.paginator.num_pages > 1
-            }},
-            status=status.HTTP_200_OK)
+        response = super(BudgetPagination, self).get_paginated_response(data)
+        response.data['page_size'] = self.page_size
+        response.data['page_count'] = self.page.paginator.num_pages
+        return response
 
 
 class ProposalViewSet(ModuleMixin,
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -1,5 +1,8 @@
 from rest_framework import mixins
+from rest_framework import status
 from rest_framework import viewsets
+from rest_framework.pagination import PageNumberPagination
+from rest_framework.response import Response
 
 from adhocracy4.api.mixins import ModuleMixin
 from adhocracy4.api.permissions import ViewSetRulesPermission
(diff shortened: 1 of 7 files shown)


sabinammm (approved):
nice!!

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +9,17 @@
 from .serializers import ProposalSerializer
 
 
+class BudgetPagination(PageNumberPagination):
+    page_size_query_param = 'page_size'
```
fuzzylogic2000, line 13: we already talked about it, but I like to set the pagination size here. :)
just `page_size = 15` should work
The current one is set in the AbstractIdeaListView and is 15 https://github.com/liqd/a4-meinberlin/blob/73a9d01832939580d3574ca037df1327eacba235/meinberlin/apps/ideas/views.py#L60
Maybe we can add a comment here to add a more general IdeaPagination once we use this for more Idea lists?

meinberlin/apps/budgeting/assets/Pagination.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import django from 'django'
+
+export const Pagination = (props) => {
+  const {
+    pageSize,
+    count,
+    currPageIndex
+  } = props
+  // FIXME: this is beyond any readability. It just calculates how many
```
fuzzylogic2000, line 10: There should be a way to get all this from the API?!

meinberlin/config/settings/base.py
```diff
@@ -538,3 +538,8 @@
     'BLUEPRINTS': 'meinberlin.apps.dashboard.blueprints.blueprints'}
 
 A4_ACTIONS_PHASE_ENDS_HOURS = 48
+
+REST_FRAMEWORK = {
+    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.PageNumberPagination',
+    'PAGE_SIZE': 100
+}
```
fuzzylogic2000, line 545: I think, we can remove this again.

fuzzylogic2000 (commented):
It looks good and works well! (And I think for testing and for now it would be good, to keep the firts state of the template?)

meinberlin/apps/budgeting/api.py
khamui: alright, will make it settable in backend and set it to 15.
shall i add the comment to `budgeting/api.py` or rather to /ideas/views.py`?

meinberlin/apps/budgeting/assets/Pagination.jsx
khamui: will try if i can provide this via api, if not i will leave for now.

meinberlin/apps/budgeting/api.py
fuzzylogic2000: I would add the comment directly here to beudgeting/api. I hope that we check and read there when we want to use React for another idea list.

meinberlin/apps/budgeting/assets/Pagination.jsx
fuzzylogic2000: I think, it should be count, next and previous. https://www.django-rest-framework.org/api-guide/pagination/#pagenumberpagination

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +11,45 @@
...
+    page_size = 15
+
+    def get_next_page(self):
+        if self.page.has_next():
+            return self.page.next_page_number()
+        else:
+            return None
+
+    def get_previous_page(self):
+        if self.page.has_previous():
```
fuzzylogic2000, line 26: Why don't you use the `next` and `previous`, the serializer just gives you for free?

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +11,45 @@
...
+            return self.page.previous_page_number()
+        else:
+            return None
+
+    def get_paginated_response(self, data):
+        return Response({
+            'status': True,
+            'results': data,
+            'meta': {
```
fuzzylogic2000, line 35: Same here: do you need all of that? Could you use the stuff you have?

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +11,45 @@
...
+                'results_count': self.page.paginator.count,
```
fuzzylogic2000, line 36: This also comes with the paginator without overwriting get_paginated_response

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +11,45 @@
...
+                'current_page': self.page.number,
```
fuzzylogic2000, line 37: You should know this in react, because that is also in the url,  you get the data from?

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +11,45 @@
...
+                'previous_page': self.get_previous_page(),
+                'next_page': self.get_next_page(),
+                'page_count': self.page.paginator.num_pages,
```
fuzzylogic2000, line 40: yes, if you need this, you'll have to add it. :)

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +11,45 @@
...
+                'page_size': self.page_size,
```
fuzzylogic2000, line 41: same here! That's something to add.

meinberlin/apps/budgeting/api.py
```diff
@@ -8,11 +11,45 @@
...
+                'is_paginated': self.page.paginator.num_pages > 1
```
fuzzylogic2000, line 42: For this, you could also ask `if next`, when you are on the first page.

fuzzylogic2000 (commented):
Yeah, I like that all the logic is in the API now. But I think, that can be done with less code by using what the paginator already gives you. 
https://www.django-rest-framework.org/api-guide/pagination/#custom-pagination-styles

fuzzylogic2000 (approved):
Nice!

PR #3974, opened 2021-11-18
meinberlin/apps/budgeting & dashboard: added text for third voting ph…

sabinammm:
…ase in budgeting

make po was not working properly so translations are still missing:

- [x] add translations

Notes for reviewers:
The **first commit** also contains the addition of the new icon for the 3-phase voting module.
The **second commit** only contains changes to the Makefile, since it appears that the makemessages command has changed, and now requires a new flag in order to work. I tested the effect only in the context of this task.


fuzzylogic2000 (approved):
Looks good!
For next time: please add the translations in two commits: the first after doing `make po` and the second one only with the translations you added.

fuzzylogic2000:
Can you also fix the translations in a+, then? I guess we also have the same problem there.

fuzzylogic2000:
And supercool that you found how to fix the makemessages!

PR #3975, opened 2021-11-18
apps/budgeting/proposal-list-item: adding features as moderator feedback e…

khamui:
…tc. and structure into smaller components

### DIFF
--- a/meinberlin/apps/budgeting/assets/ListItemBadges.jsx
+++ b/meinberlin/apps/budgeting/assets/ListItemBadges.jsx
@@ -15,8 +15,8 @@ export const ListItemBadges = (props) => {
           {props.category.name}
         </span>}
       {props.pointLabel &&
-        <span class="label label--big">
-          <i class="fas fa-map-marker-alt" aria-hidden="true" />
+        <span className="label label--big">
+          <i className="fas fa-map-marker-alt" aria-hidden="true" />
           {props.pointLabel}
         </span>}
       {props.budget > 0 &&
--- a/meinberlin/apps/budgeting/assets/ListItemBadges.jsx
+++ b/meinberlin/apps/budgeting/assets/ListItemBadges.jsx
@@ -1,0 +1,39 @@
+import React from 'react'
+import { intComma } from './helpers'
+
+export const ListItemBadges = (props) => {
+  const getModeratorChoice = () => {
+    return props.moderatorChoices.find(choice => {
+      return choice[0] === props.moderatorFeedback
+    })
+  }
+
+  return (
+    <div className="list-item__labels">
+      {props.category &&
+        <span className="label label--big">
+          {props.category.name}
+        </span>}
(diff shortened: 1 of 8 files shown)


meinberlin/apps/budgeting/assets/ListItemBadges.jsx
```diff
@@ -0,0 +1,39 @@
...
+      {props.budget > 0 &&
+        <span className="label label--big">
+          {intComma(props.budget)}€
+        </span>}
+      {props.moderatorFeedback &&
+        <span
+          className={
+          'label' +
+          ' label--big' +
+          ' label--' + props.moderatorFeedback.toLowerCase() +
```
phillimorland, line 31: nice! :) also just relaised we need to update this on the detail page oupps

meinberlin/apps/budgeting/assets/helpers.js
```diff
@@ -14,7 +14,12 @@ const months = [
...
   django.gettext('December')
 ]
+
 export const toDate = (isodate) => {
   const d = new Date(isodate)
   return `${d.getDate()}. ${months[d.getMonth()]} ${d.getFullYear()}`
 }
+
+export const intComma = (number) => {
+  return Number(number).toLocaleString()
```
phillimorland, line 24: is this adding a comma for cents? as we can only have whole numbers do we need it?

phillimorland (changes requested):
Super cool! only thing is, when checking locally the margins between badges is off when more then 1 also the spacing between icons and numbers, once thats sorted it's good to go!

meinberlin/apps/budgeting/assets/helpers.js
phillimorland: maybe in js files we should consider having comments? as can't always tell, specially with no fuile name to help

khamui: @phillimorland actually it is for the thousand separator (20.000 or 20,000). Works as the filter in django. I admit the name could be better but i named it how the template filter is named in django.

khamui:
> Super cool! only thing is, when checking locally the margins between badges is off when more then 1 also the spacing between icons and numbers, once thats sorted it's good to go!

alright will check it!

khamui:
@phillimorland i didn't know that span elements are that complicated. but it seems that if multiple spans are separated by line breaking in react, the whitespaces in between are getting removed. i found a solution which i implemented into a new span element because it really looks weird and screws the readability of our JSX. hit me up if you have a better idea.

meinberlin/apps/budgeting/assets/helpers.js
phillimorland: fab!

phillimorland (approved):
Really nice solution, I think we had this issue in comments too, will have to port the solution over at some point :)

PR #3977, opened 2021-11-19
apps/contrib: make sure ordering filter also works on annotated field…

### LINKED ISSUE #3976: FieldError: Cannot resolve keyword 'positive_rating_count' into field. Choices are: allow_contact, budget, ca...
Sentry Issue: [MEINBERLIN-DEV-49](https://sentry.liqd.net/organizations/liqd/issues/1822/?referrer=github_integration)

```
FieldError: Cannot resolve keyword 'positive_rating_count' into field. Choices are: allow_contact, budget, category, category_id, comments, contact_email, contact_phone, created, creator, creator_id, description, id, image, is_archived, item_ptr, item_ptr_id, labels, moderator_feedback, moderator_statement, moderator_statement_id, modified, module, module_id, name, point, point_label, ratings, slug
(12 additional frame(s) were not displayed)
...
  File "django_filters/filterset.py", line 230, in filter_queryset
    queryset = self.filters[name].filter(queryset, value)
  File "adhocracy4/filters/filters.py", line 75, in filter
    return qs.order_by(*ordering)
  File "django/db/models/query.py", line 1149, in order_by
    obj.query.add_ordering(*field_names)
  File "django/db/models/sql/query.py", line 1993, in add_ordering
    self.names_to_path(item.split(LOOKUP_SEP), self.model._meta)
  File "django/db/models/sql/query.py", line 1540, in names_to_path
    "Choices are: %s" % (name, ", ".join(available)))
```

fuzzylogic2000:
…s comment_count and positive_rating_count

fixes #3976
I checked all views using the contrib OrderingFilter: the positive rating and the comment count are the only annotated fields used for ordering

Maybe this should go into a4? But then we would need to check all projects for the use of annotated fields for ordering and add these.
And I am not sure, if we should open a Django bug. It used to work and does not seem like a very off-usecase. (On the other hand the distinct ordering is also not an off usecase and did not make it into django-filters.)

### DIFF
--- a/meinberlin/apps/budgeting/views.py
+++ b/meinberlin/apps/budgeting/views.py
@@ -6,7 +6,6 @@ from adhocracy4.categories import filters as category_filters
 from adhocracy4.exports.views import DashboardExportView
 from adhocracy4.filters import filters as a4_filters
 from adhocracy4.projects.mixins import DisplayProjectOrModuleMixin
-from meinberlin.apps.contrib import filters
 from meinberlin.apps.ideas import views as idea_views
 from meinberlin.apps.projects.views import ArchivedWidget
 from meinberlin.apps.votes.forms import TokenForm
@@ -29,7 +28,7 @@ class ProposalFilterSet(a4_filters.DefaultsFilterSet):
         'is_archived': 'false'
     }
     category = category_filters.CategoryFilter()
-    ordering = filters.OrderingFilter(
+    ordering = a4_filters.DynamicChoicesOrderingFilter(
         choices=get_ordering_choices
     )
     is_archived = django_filters.BooleanFilter(
@@ -54,10 +53,7 @@ class ProposalListView(idea_views.AbstractIdeaListView,
 
     def get_queryset(self):
         return super().get_queryset()\
-            .filter(module=self.module) \
-            .annotate_positive_rating_count() \
-            .annotate_negative_rating_count() \
-            .annotate_comment_count()
+            .filter(module=self.module)
 
     def get_context_data(self, **kwargs):
         if 'token_form' not in kwargs:
--- a/meinberlin/apps/contrib/filters.py
+++ b/meinberlin/apps/contrib/filters.py
@@ -1,25 +1,4 @@
-from django.utils.translation import gettext_lazy as _
 from rest_framework.filters import BaseFilterBackend
-
-from adhocracy4.filters.filters import DistinctOrderingFilter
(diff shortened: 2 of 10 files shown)


sabinammm (approved):
I checked this in Kosmo where it still seems to work, so maybe not add it to a4 yet?

Rineee:
Supercool! Although I dont really understand, why it is needed/was not working before? Maybe you could explain that to me later?

It is also not working in Aplus and Kosmo, so I think it would make sense to add it to A4, what do you think?

fuzzylogic2000:
> Supercool! Although I dont really understand, why it is needed/was not working before? Maybe you could explain that to me later?
> 
> It is also not working in Aplus and Kosmo, so I think it would make sense to add it to A4, what do you think?

Yes, you are right, I will add it to a4!

goapunk:
If we apply the annotations in the filter method, do we still need to apply them in `mapideas/views.py` for example ? And what about `negative_rating_count` ?

fuzzylogic2000:
> @Rineee @goapunk Sooooo, when we do the annotations in the filter, it does not add more SQL queries to the many ones that are there anyway. I am still unsure about putting it into a4, though. We could do that as here [liqd/adhocracy4#912](https://github.com/liqd/adhocracy4/pull/912) and then use the `DistinctOrderingAnnotatedFilter` instead of the `DistinctOrderingFilter`. But that would move the annotations even further away from the view.

Ok, and more thoughts about a4: the only project we also do sth. like that is a+. Al other projects use the DistinctOrderingFilter and add the ordering choices in a static way. 
Instead of just moving the overwrite of the filter, we should either move the whole OrderingFilter and the DynamicChoicesMixin to a4 or keep it where it is. As we can get rid of doubled code, I am in favour of moving to a4 now!

goapunk:
@fuzzylogic2000 Yeah, I think in a4 is fine, less code and it's somewhat hidden away anyway with both solutions

Rineee:
@fuzzylogic2000 Yes, I also think its a good idea to move the OrderingFilter and Mixin to A4 and then use it from there :+1: And then do the annotations there, right?

fuzzylogic2000:
Alright! Still depending on https://github.com/liqd/adhocracy4/pull/912 (so the build will fail until we update a4) But done and moved all the code! :partying_face: 
@Rineee @goapunk

Rineee (approved):
Aaaweesome, everything working perfectly again! Can be merged once A4 is updated!

@fuzzylogic2000 Should we also dirctly do the same in A+?

PR #3981, opened 2021-11-24
[refactor] make icon switch component in react and use in toggles for project overview

phillimorland:
wrote this for the proposal list but didn't need it in the end so re-purposed for project overview


meinberlin/apps/contrib/assets/IconSwitch.jsx
```diff
@@ -0,0 +1,60 @@
...
+    endIconClass,
+    startText,
+    endText,
+    startID,
+    endID,
+    startAria,
+    endAria,
+    showStartObject,
+    showEndObject,
+    displayStartObject
```
phillimorland, line 17: start and end used due to bootstrap adopting it

khamui:
nice one! works fine

PR #3982, opened 2021-11-24
[issues] Cosmetic fixes for detail and list view of proposals and ideas

phillimorland:


khamui:
nice! pixel perfect and colorful again :)

PR #3983, opened 2021-11-24
WIP: apps/budgeting/proposals/filter: adjusting template

khamui:
to switch from map to list (in django)
to switch from list to map (in react)

### DIFF
--- a/meinberlin/apps/budgeting/assets/FilterBar.jsx
+++ b/meinberlin/apps/budgeting/assets/FilterBar.jsx
@@ -1,0 +1,51 @@
+import React, { useState } from 'react'
+import django from 'django'
+
+const getItemByFilter = (items, filter) => {
+  return items.find(item => item.filter === filter)
+}
+
+export const FilterBar = (props) => {
+  const filterItems = [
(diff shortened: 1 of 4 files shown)


meinberlin/apps/budgeting/assets/FilterBar.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react'
```
khamui, line 1: this file was just copied as base for the Filters, is going to be changed entirely.

phillimorland: @khamui i did a reuseable switch component to replace the switch in react but not sure if it'll work with how we want to do it, but we should talk about it before we go further to not double up work https://github.com/liqd/a4-meinberlin/pull/3981/files

khamui: @phillimorland true! lets use the switch component if possible :) you want to take a look if its possible?
I will focus on the filters then and leave the button untouched. does that make sense?

khamui:
will implement feature on a new branch.

PR #3985, opened 2021-11-25
meinberlin/templates: added general error message on top of all forms

sabinammm:
A general error message at the top of the form was added to all the forms **except** the Bplans, because this form doesn't use the errorlist box on fields:
image

, but instead the 'Missing field for publication'-icons: 
image

This solution is based on the a+ use of the top error fields with the exception of the ideas form, which has a missing top error message:

2021-11-25Missing_upper_error_message_a+

The mB idea form has this missing message.

**QUESTION**: does this implementation make sense? It seems there are about 3 different ways that error messages are displayed for the user in mB.

This PR still needs translation once everything else is reviewed.

### DIFF
--- a/meinberlin/apps/ideas/templates/meinberlin_ideas/idea_create_form.html
+++ b/meinberlin/apps/ideas/templates/meinberlin_ideas/idea_create_form.html
@@ -17,6 +17,11 @@
         </nav>
 
         <h1>{% trans 'Submit a new idea for this project'%}</h1>
+
+        {% if form.errors %}
+            <div class="errorlist">{% trans 'An error occurred while evaluating your data. Please check the data you entered again.' %}</div>
+        {% endif %}
+
         <div class="u-bottom-divider">
             <h3>{{ module.name }}</h3>
             {% if module.description %}
--- a/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_create_form.html
+++ b/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_create_form.html
@@ -17,6 +17,10 @@
 
         <h1>{% trans 'Submit a new proposal for this project'%}</h1>
 
+        {% if form.errors %}
+            <div class="errorlist">{% trans 'An error occurred while evaluating your data. Please check the data you entered again.' %}</div>
+        {% endif %}
+
         {% include "meinberlin_budgeting/includes/proposal_form.html" with proposal=proposal cancel=module.get_detail_url %}
     </div>
 </div>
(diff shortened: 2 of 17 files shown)


meinberlin/apps/ideas/templates/meinberlin_ideas/idea_create_form.html
```diff
@@ -17,6 +17,11 @@
         </nav>
 
         <h1>{% trans 'Submit a new idea for this project'%}</h1>
+
+        {% if form.errors %}
+            <div class="errorlist">{% trans 'An error occurred while evaluating your data. Please check the data you entered again.' %}</div>
+        {% endif %}
+
```
phillimorland, line 24: error does not trigger for not adding a title, is that expected behaviour? also for the edit forms

phillimorland:
@sabinammm the question of does the doubling up error messages make sense is either for PM and/or designer, you can ask them in the story with screen shots. 
As for the different implememnations of errors, maybe make screen shots of all the differnt examples and ask Marcelo how to proceed and make an issue.

meinberlin/apps/ideas/templates/meinberlin_ideas/idea_create_form.html
sabinammm: Yes, that's a good question, and what I meant by that there are different ways that error messages are displayed. In this case, I don't know where these error messages come from, and they're always in english:
image
I will create an issue for this.

phillimorland: @sabinammm they are error messages from the browser as that field is required in the template, the browser knows this and will display an error message when it is not filled, I think it only displays when there is no error handling in the code but maybe it does it even when there is? not sure.

fuzzylogic2000 (changes requested):
Nice! It works where it is implemented. But missing are the proposal (Bürgerhaushalt) and maptopic (Maptopicprio) forms. It would probably also be good to include it into the topic (Topicprio) form, in case we add something that will trigger error messages. I am not sure, I found all the idea derivate forms. Could you please test all the modules?

sabinammm:
> Nice! It works where it is implemented. But missing are the proposal (Bürgerhaushalt) and maptopic (Maptopicprio) forms. It would probably also be good to include it into the topic (Topicprio) form, in case we add something that will trigger error messages. I am not sure, I found all the idea derivate forms. Could you please test all the modules?
> 
> Edited: the offline events (Veranstaltungen) are also missing the alert. (only triggered by filling out everything, but not the description). - I think when you found all places to add it, a+ can also profit from that. :)

I checked all 13 modules, added those you mentioned as well as the Kiezkasse, and added the offline events in the project details as well. Thanks for clearing up all the issues regarding deviating error messages, as well:)!!

fuzzylogic2000 (approved):
Nice!
The only thing still missing I could find is in the dashboard project tabs for the moderators, when you add sth. that is not an email address, but as that is at the very top, I think that is fine.

PR #3986, opened 2021-11-25
apps/budgeting/proposal-item: using internationalized date

khamui:


khamui:
Instead of relying on another third party library i found this build-in way to make dates internationalized (similar to what we have in Django). So i wrapped it into a helper function.

In order to have this working, we need to pass `locale` information to the frontend. What do you guys think?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

PS:
I have to admit the template filters (`richtext`, `intcomma`, `date` etc.) is a cool concept in Django. Don't know why react does not offer these converters.

fuzzylogic2000 (approved):
Nice!
It will not have a huge impact in mB, where we force the locale to be German (https://github.com/liqd/admin/blob/a9c81b672cf1863c5f6e256402f96ea734cf8ec2/pillar/servers/berlin-adhocracy4-prod.sls#L119), but it might solve our problems elsewhere!

sabinammm (commented):
Looks really nice & working perfectly for me:)

PR #3994, opened 2021-11-29
Kt 2021 11 proposal filter bar

khamui:
- [x] adding components (filterbar, filterbardropdwon)
- [x] send requests with ordering and filters
- [x] adjust pagination to work with filters
- [x] adjust filter to work with final backend response
- [ ] adding icon to category dropdown (dep. on backend)
- [ ] adding tests for new components
- [x] adding listmap switch button
- [ ] setting default states for filters

### DIFF
--- a/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
+++ b/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
@@ -3,14 +3,16 @@ import django from 'django'
 import { BudgetingProposalListItem } from './BudgetingProposalListItem'
 import { Pagination } from './Pagination'
 import { CountDown } from '../../contrib/assets/CountDown'
+import { FilterBar } from './FilterBar'
 
 export const BudgetingProposalList = (props) => {
   const [data, setData] = useState([])
   const [meta, setMeta] = useState()
+  const [queryString, setQueryString] = useState('')
 
   const fetchProposals = (newIndex) => {
     const pageNumber = newIndex || 1
-    const url = `${props.proposals_api_url}?page=${pageNumber}`
+    const url = `${props.proposals_api_url}?page=${pageNumber}${queryString}`
     fetch(url)
       .then(resp => resp.json())
       .then(json => {
@@ -20,7 +22,8 @@ export const BudgetingProposalList = (props) => {
           page_count: json.page_count,
           is_paginated: json.page_count > 1,
           previous: json.previous,
-          next: json.next
+          next: json.next,
+          filters: json.filters
         })
       })
       .catch(error => console.log(error))
@@ -30,7 +33,11 @@ export const BudgetingProposalList = (props) => {
     fetchProposals(selectedPage)
   }
 
-  useEffect(fetchProposals, [])
+  const onChangeFilters = (filterString) => {
+    setQueryString(filterString)
+  }
+
+  useEffect(fetchProposals, [queryString])
 
   const renderList = (data) => {
     let list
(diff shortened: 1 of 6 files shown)


meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -2,14 +2,49 @@ import React, { useEffect, useState } from 'react'
 import django from 'django'
 import { BudgetingProposalListItem } from './BudgetingProposalListItem'
 import { Pagination } from './Pagination'
+import { FilterBar } from './FilterBar'
+
+// filter mock, naive assumption of object structure.
+// to be replaced and adjusted, depending on what comes
+// from backend at the end
+const FILTERS = [
```
khamui, line 10: @fuzzylogic2000 currently the frontend works with these object structure assumption. not sure if it makes sense, not sure if it is possible to pass the query name (here `type`) from backend?

fuzzylogic2000: I guess that is a question for @Rineee as she works on the category and ordering choices now. Should filter name and type also come from the backend? Or do we assume we know the backend and just add the choices?

khamui: it would be nice if we can get these infos (type, name) from backend. so we have a single source of truth.
i was wondering if the `All` state inside of choices should come from backend too? if so i can just fetch everything (except position) from backend and do not need to "serialize" in frontend again. but not sure if this is the responsibility of our backend, though. what do you think?

fuzzylogic2000: Well, I love when the backend does all the logic. :)

Rineee: Hahha, me too :) Will look into it now and see how I we can provide you all the info best!

fuzzylogic2000 (commented):
Before I start reviewing, a comment: the frontend tests are failing. :)
Maybe you can already have look while I read the code and try it out?

fuzzylogic2000 (approved):
I cannot quite wrap my head around the filter functions (especially the `encodeFilterString`), but it works like a charm!

meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -70,21 +77,31 @@ export const BudgetingProposalList = (props) => {
...
+                countText={getVoteCountText(voteCount)}
+                activeClass="btn btn--transparent btn--full u-spacer-bottom btn--huge u-primary"
+                inactiveClass="btn btn--full btn--light u-spacer-bottom btn--huge"
+                counter={voteCount}
+              />
+            </div>
+          </div>
+        </div>}
+      <FilterBar
+        filters={meta?.filters}
```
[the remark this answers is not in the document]
khamui: yes exactly, just to organize everything else in one state variable `meta`

meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -70,21 +77,31 @@ export const BudgetingProposalList = (props) => {
...
+        onChangeFilters={filterString => onChangeFilters(filterString)}
```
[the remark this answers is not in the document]
khamui: yes exactly. basically the `encodeFilterString` converts an object to a query- / filterstring.
`makeFilterObject()` makes that object to be consumed by `encodeFilterString`. For instance if you select different options from the ordering filter, these two methods make sure it is not doubled up, but substituted. Maybe it is the wording that makes it hard to understand....but i cannot find better words.

PR #3995, opened 2021-11-29
[#4987] apps/budgeting/__tests__: add snapshot for budgeting list without items

phillimorland:
This is just testing when there are no items

### DIFF
--- a/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
+++ b/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react'
+import django from 'django'
 import { BudgetingProposalListItem } from './BudgetingProposalListItem'
 import { Pagination } from './Pagination'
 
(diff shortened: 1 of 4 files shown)


meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -56,6 +53,9 @@ export const BudgetingProposalList = (props) => {
                     onPaginate={newUrl => onPaginate(newUrl)}
                   />}
               </>
+              )
+            : (
+              <div>{django.gettext('Nothing to show')}</div>
```
phillimorland, line 58: @khamui the snapshots always show this version, never the empty list version. I have tried a number of ways to change that but i'm still stumped, maybe we can have a look together?

meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -48,6 +50,21 @@ export const BudgetingProposalList = (props) => {
               pageCount={meta.page_count}
               onPaginate={newUrl => onPaginate(newUrl)}
             />}
+        </>
+      )
+    } else {
+      list = (
+        <span>{django.gettext('Nothing to show')}</span>
```
goapunk, line 57: Did you add this new empty state only for the test or was that something we wanted to have anyway?

phillimorland: this was there in the django template but was forgotton in the jsx, adding test meant that I found this out (even if test isn't quite working)

khamui:
cool

PR #3996, opened 2021-11-29
WIP: apps/contrib: use request instead of view for module dpendend filtering

### LINKED ISSUE #3976: FieldError: Cannot resolve keyword 'positive_rating_count' into field. Choices are: allow_contact, budget, ca...

fuzzylogic2000:
depending on https://github.com/liqd/adhocracy4/pull/906

fixes #3976 but only for the idea list (brainstorming and idea collection) so far

@Rineee @goapunk Do you think this is how we should do it? There would be a lot of ListViews still to fix and also the idea list test (tests/ideas/views/test_idea_list.py::test_list_view) is showing a problem in a4 
```
  File "/home/katharina/adhocracy4/adhocracy4/filters/filters.py", line 41, in __init__
    data = data.copy()
AttributeError: 'NoneType' object has no attribute 'copy'
```
which I also saw at some point while trying around with the filters.

### DIFF
--- a/meinberlin/apps/contrib/mixins.py
+++ b/meinberlin/apps/contrib/mixins.py
@@ -13,27 +13,45 @@ class DynamicChoicesMixin(object):
 
     Add callable functionality to filters that support the ``choices``
     argument. If the ``choices`` is callable, then it **must** accept the
-    ``view`` object as a single argument.
-    The ``view`` object may be None if the parent FilterSet is not class based.
+    ``request`` object as a single argument.
+
+    FIXME: check!
+    The ``request`` object may be None if the parent FilterSet is not class
+    based.
 
     This is useful for dymanic ``choices`` determined properties on the
-    ``view`` object.
+    ``request`` object.
+
+    Like
+    https://github.com/carltongibson/django-filter/blob/
+    ab7978323ff092c0ccd2b95e9b190ba8522bde11/django_filters/filters.py#L295
+
+    Was changed from request to view in
+    https://github.com/liqd/a4-meinberlin/commit/
+    ea74f6ad5b04922a7c10477a9074b3047189c8db
     """
 
     def __init__(self, *args, **kwargs):
         self.choices = kwargs.pop('choices')
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


goapunk:
I'm not very familiar with the code, so I leave this to @Rineee. In general looks like a good solution to me though.

Rineee:
I get an error when looking at a brainstorming

`<h1>AttributeError
       at /projekte/module/brainstorming-4/</h1>
  <pre class="exception_value">'NoneType' object has no attribute 'copy'</pre>
  

Request Method: | GET
-- | --
http://localhost:8003/projekte/module/brainstorming-4/
3.2.9
AttributeError
'NoneType' object has no attribute 'copy'
/home/rine/adhocracy4/adhocracy4/filters/filters.py, line 42, in __init__

`

fuzzylogic2000:
As annotating in the filter does not produce more SQL-queries (additionally to the supermany ones that are there...), I will close this and use the other approach.

PR #4006, opened 2021-11-30
meinberlin/apps/ideas and /contrib: added last edited text and date t…

sabinammm:
…o idea detail and list

**Notes and questions:**
1. The wording is based on the wording in the plan detail page, i.e. it always says "edited on [Date]" even if it has just been created. 

@fuzzylogic2000 Should this be different here maybe, i.e. switch to _created_ instead of _edited_ or is it more important that it's consistently the same as in plans? Maybe a question for Caro before it is implemented everywhere.

2. This change of the item-detail affects the map-idea detail, idea-detail, and maptopic detail pages. While the list item of the plain idea already has the change, the other list items do not.

### DIFF
--- a/meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
+++ b/meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
@@ -50,7 +50,7 @@
             <div class="item-detail__meta lr-bar">
                 <div class="lr-bar__left">
                     <strong class="item-detail__creator">{{ object.creator.username }}</strong>
-                    {% html_date object.created %}
+                    {% trans 'updated on ' %}{% if object.modified %}{{ object.modified|date }}{% else %}{{ object.created|date }}{% endif %}
                 </div>
                 <div class="lr-bar__right">
                     <strong>{% trans 'Reference No.' %}:</strong>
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (approved):
Yes, this is exactly how we said we would try it. We will ask @CarolingerSeilchenspringer and @mcastro-lqd the question you asked me. 

You also use the date filter instead of the html_date template tag. Could you please investigate, why we do use both in different places? (hint: if you found the file where it is used or added you can use the "blame" button to find out with which PR it was added and then you might find info in there.)

sabinammm:
> Yes, this is exactly how we said we would try it. We will ask @CarolingerSeilchenspringer and @mcastro-lqd the question you asked me.
> 
> You also use the date filter instead of the html_date template tag. Could you please investigate, why we do use both in different places? (hint: if you found the file where it is used or added you can use the "blame" button to find out with which PR it was added and then you might find info in there.)

I found where the date was added in plans and where the html_date was added everywhere else that the created date is shown. But I don't understand why this isn't also used in plans. It seems that the tag was created to make the date readable for screen readers. If this is no longer relevant because screen readers can now read the date without this extra parsing, then I would leave it off. Since the date only added over a year after the html_date tag, maybe this is why.

fuzzylogic2000:
@sabinammm I added an issue to find out, which one we should use: #4017 
I guess, we just forgot to use it when adding the plans. But maybe you are also right, that we do not need it anymore. 
Anyway, I hoped you would find out if we need it or don't (for example by opening an issue to discuss it or asking @phillimorland ) instead of guessing why we do use it only in some places. If there are inconsistencies in the code like that, it is seldomly on purpose.

PR #4007, opened 2021-11-30
apps/budgeting: add ordering filter to API

fuzzylogic2000:
to get the same results as in the original list, please add
ordering=-created or ordering=-comment_count or ordering=-positive_rating_count
as parameter to the API url.

You can already merge and use this. I will open a new PR for the other filter stuff, because I have to fix all the rest API filtering where we use `filter_fields`.


khamui:
nice! thanks

PR #4013, opened 2021-11-30
meinberlin/apps/contrib /ideas /mapideas: added updated and created t…

sabinammm:
…ext to ideas and proposals

To-do/check:

- [x] Budgeting proposal list item only in jsx ==> after merging new main, it's also in list view
- [x] published plans cannot show date created. They show modified as soon as they are published ==> date created also shown in draft for completeness

**1st commit:**
Ideas for the following modules now have date created & date modified:

- **Kiezkasse**
- **Brainstorming with map**
- **Brainstorming**
- **Idea collection**
- **Idea collection with map**
- **Bürger*innenhaushalt/Budget Proposals** (1/2/3 Phases)
- **Plans** in draft (created date if not modified), modified date in published

Needed for Finalization:

Requested feedback from design on whether the have the referenz number on the left side when it overlaps with the date when screen is small:
normal:
image

small:
image

### DIFF
--- a/meinberlin/apps/contrib/templates/meinberlin_contrib/includes/proposal_list_item.html
+++ b/meinberlin/apps/contrib/templates/meinberlin_contrib/includes/proposal_list_item.html
@@ -43,5 +43,5 @@
     <span class="list-item__author">
         {{ object.creator.username }}
     </span>
-    {% html_date object.created class='list-item__date' %}
+    {% if object.modified %}{% trans 'updated on ' %}{{ object.modified|date }}{% else %}{% trans 'created on ' %}{{ object.created|date }}{% endif %}
 </li>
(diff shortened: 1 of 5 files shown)


fuzzylogic2000:
> To-do/check:
> 
>     * [ ]  Budgeting proposal list item only in jsx

I readded the list in #4016, so after merge and rebase it can be done in the "old" list first.

>     * [ ]  published plans cannot show date created. They show modified as soon as they are published (???)

Yes, true! As the plan is updated when it is published, there shouldn't normally be the created, but always the modified date in the detail view. But for the preview it is still possible that there is only the created date and that is shown.

fuzzylogic2000 (approved):
So, the only thing still missing is the design change? But as that is not really broken (text is not going over each other, but reference number goes to next line), I would say we merge and see if we need to change the styling later.

PR #4014, opened 2021-12-01
Kl 2021 11 more propsal api filters

fuzzylogic2000:
This does the filtering. But we need to know which categories to display (and their icons). 
What is the best way to add that? I think previously we either added that to the template tag or to the serializer. But I guess the best option would be to add it to the response.data of the API, which should be possible? What do you think @Rineee ?

### DIFF
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -6,6 +6,7 @@ from rest_framework.pagination import PageNumberPagination
 
 from adhocracy4.api.mixins import ModuleMixin
 from adhocracy4.api.permissions import ViewSetRulesPermission
+from meinberlin.apps.contrib.filters import IdeaCategoryFilterBackend
 
 from .models import Proposal
 from .serializers import ProposalSerializer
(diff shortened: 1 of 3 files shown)


fuzzylogic2000:
> This does the filtering. But we need to know which categories to display (and their icons). What is the best way to add that? I think previously we either added that to the template tag or to the serializer. But I guess the best option would be to add it to the response.data of the API, which should be possible? What do you think @Rineee ?

And wherever we put the category choices, the moderator feedback choices should also go there. (Or of course stay where they are and we also put the category choices there.) https://github.com/liqd/a4-meinberlin/blob/fce4f4d47491cc78d0420c0684fa8c96e987a7b6/meinberlin/apps/budgeting/serializers.py#L73 What do we actually need them for, though?

khamui:
> > This does the filtering. But we need to know which categories to display (and their icons). What is the best way to add that? I think previously we either added that to the template tag or to the serializer. But I guess the best option would be to add it to the response.data of the API, which should be possible? What do you think @Rineee ?
> 
> And wherever we put the category choices, the moderator feedback choices should also go there. (Or of course stay where they are and we also put the category choices there.)
> 
> https://github.com/liqd/a4-meinberlin/blob/fce4f4d47491cc78d0420c0684fa8c96e987a7b6/meinberlin/apps/budgeting/serializers.py#L73
> 
> What do we actually need them for, though?

yes i was not sure where to add the choices best.
It is used here:
https://github.com/liqd/a4-meinberlin/blob/fce4f4d47491cc78d0420c0684fa8c96e987a7b6/meinberlin/apps/budgeting/assets/ListItemBadges.jsx#L7

khamui:
yeah nice! can be filtered on the api

Rineee:
@fuzzylogic2000 @khamui I will look into where/how to add the choices best!

PR #4015, opened 2021-12-01
apps/documents: loading document js only in template, not globally to…

phillimorland:
… reduce chunk size

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -17,12 +17,6 @@ module.exports = {
       './meinberlin/assets/scss/style.scss',
       './meinberlin/assets/js/app.js'
     ],
-    blueprint_picker: {
-      import: [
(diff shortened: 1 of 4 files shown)


webpack.common.js
```diff
@@ -43,19 +37,24 @@ module.exports = {
...
+    // error possibly due to needing to be loaded at specific time
+    blueprint_picker: {
+      import: './meinberlin/assets/js/blueprint-picker.js'
+    },
+    documents: {
+      import: './meinberlin/apps/documents/assets/react_documents_init.js'
+    },
+    unload_warning: {
+      import: './meinberlin/assets/js/unload_warning.js'
+    },
```
phillimorland, line 57: this gets ride of the annoying console errors finally, I think it's because these need to be initialised right away and do not have dependencies in adhocracy4.js but I don't know for sure

PR #4020, opened 2021-12-02
Ks 2021 11 add token form to proposal list view

Rineee:
This adds a form for the voting token to the proposal list. When the token is there, it simply displayed for now, we will then need a react component for showing the left votes (I will next add that to the api). 
Oh, and we also need a widget for the token, so that it can be entered with 3 x 4 characters.

### DIFF
--- a/meinberlin/apps/budgeting/templatetags/react_proposals.py
+++ b/meinberlin/apps/budgeting/templatetags/react_proposals.py
@@ -4,14 +4,21 @@ from django import template
 from django.urls import reverse
 from django.utils.html import format_html
 
+from adhocracy4.phases.predicates import has_feature_active
+from meinberlin.apps.budgeting.models import Proposal
+
 register = template.Library()
 
 
 @register.simple_tag(takes_context=True)
-def react_proposals(context, obj):
-    proposals_api_url = reverse('proposals-list', kwargs={'module_pk': obj.pk})
+def react_proposals(context, module):
+    proposals_api_url = reverse('proposals-list',
+                                kwargs={'module_pk': module.pk})
     attributes = {'proposals_api_url': proposals_api_url,
-                  'is_voting_phase': 'voting' in obj.active_phase.type}
+                  'is_voting_phase': has_feature_active(module,
+                                                        Proposal,
+                                                        'vote')
+                  }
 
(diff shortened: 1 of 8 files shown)


meinberlin/apps/budgeting/templatetags/react_proposals.py
```diff
@@ -4,14 +4,21 @@
...
 
 @register.simple_tag(takes_context=True)
-def react_proposals(context, obj):
-    proposals_api_url = reverse('proposals-list', kwargs={'module_pk': obj.pk})
+def react_proposals(context, module):
+    proposals_api_url = reverse('proposals-list',
+                                kwargs={'module_pk': module.pk})
     attributes = {'proposals_api_url': proposals_api_url,
-                  'is_voting_phase': 'voting' in obj.active_phase.type}
+                  'is_voting_phase': has_feature_active(module,
```
Rineee, line 18: @fuzzylogic2000 I fixed this for now, so we dont get the 500, but I think this should go to the api as has_voting_permission, so admins are always allowed to vote, right? What do we do with showing the ratings though? Should they be visible for admins all the time? So then also just show them with has_rating_permission?

khamui: nice

meinberlin/apps/votes/models.py
```diff
@@ -78,6 +80,11 @@ def is_valid(self, model):
     def is_valid_for_item(self, item):
         return item.module == self.module and self.is_valid(item.__class__)
 
+    def __str__(self):
+        return '{}-{}-{}'.format(self.token[0:4],
+                                 self.token[4:8],
+                                 self.token[8:13])
```
fuzzylogic2000, line 86: This should only go up to 12, `self.token[8:12]` or to the end `self.token[8:]`

fuzzylogic2000 (commented):
Very nice! Just a tinsi tiny thing. :)

meinberlin/apps/votes/models.py
Rineee: Haha, right, good catch!

fuzzylogic2000 (approved):

PR #4021, opened 2021-12-02
apps/maps: rm map files to a4 and update imports and deps

phillimorland:
to be tested with https://github.com/liqd/adhocracy4/pull/914/commits/2d41aa792366b117a22c205c4915dac3e05948de


khamui:
working fine (without showing actual maps on my local, but that is expected).

these are the things i checked. To be verified on dev once merged
- [ ] map in dashboard (where you select a Bezirk)
- [ ] map and pins in project overview (combined map)
- [ ] map and pins in list view (budgeting etc.)
- [ ] map and pins in proposal view

approved and can be merged when is best!

PR #4032, opened 2021-12-06
[#4987] apps/contrib: add dummy count component to be used in budgetList

phillimorland:


meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -60,9 +61,26 @@ export const BudgetingProposalList = (props) => {
     return list
   }
 
+  // this is just a placeholder until we have the votes
+  const voteCount = 1
+
+  const getVoteCountText = (votes) => {
+    const countText = django.ngettext('you have 1 vote left', 'you have %s votes left', votes)
+    return django.interpolate(countText, [votes])
```
khamui, line 69: nice!

khamui:
Nice! works fine.

Just one suggestion, what do you think about to omit the name `...Component`? `Count.jsx` is too general though?
Its just the word Component makes the name relatively long.

PR #4033, opened 2021-12-06
apps/votes: added widget to voting form

sabinammm:
This produces three instead of one input field for the voting token. Each field requires 4 characters.

Still to do

- [x] Form styling 
- [x] The form validation is in contradiction and requires changes, see https://github.com/liqd/a4-meinberlin/pull/4033#discussion_r763040554

@phillimorland would you check out the styling when you get the chance? 
@Rineee I added comments for the form validation that I can't get to work well, maybe you can check if you have time? I think the decompress and value_from_dict functions work.

### DIFF
--- a/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
+++ b/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
@@ -32,11 +32,11 @@
                 <form class="" action="{{ request.path }}" method="post">
                     <div class="form-group">
                         {% csrf_token %}
-                        <div class="widget widget--{{ form.token|widget_type }}">
+                        <div class="widget--{{ token_form.token|widget_type }}">
                             {{ token_form.token }}
+                            <button type="submit" class="btn">{% trans 'Enter code' %}</button>
                         </div>
                         {{ token_form.token.errors }}
-                        <button type="submit" class="btn btn--full u-spacer-bottom btn--huge">{% trans 'Enter code' %}</button>
                     </div>
                 </form>
             {% endif %}
--- a/meinberlin/apps/votes/forms.py
+++ b/meinberlin/apps/votes/forms.py
@@ -4,9 +4,31 @@ from django.utils.translation import gettext_lazy as _
 from meinberlin.apps.votes.models import VotingToken
 
 
-class TokenForm(forms.Form):
+class VotingTokenWidget(forms.MultiWidget):
 
-    token = forms.CharField(max_length=40)
+    def __init__(self, attrs=None):
+        widgets = (
+            forms.TextInput(attrs={'minlength': '4',
+                                   'maxlength': '4'}),
+            forms.TextInput(attrs={'minlength': '4',
+                                   'maxlength': '4'}),
+            forms.TextInput(attrs={'minlength': '4',
+                                   'maxlength': '4'}),
+        )
+        super(VotingTokenWidget, self).__init__(widgets, attrs)
+
+    def decompress(self, value):
+        if value:
+            return [value[i:i + 4] for i in range(0, len(value), 4)]
+        return [None, None, None]
+
+    def value_from_datadict(self, data, files, name):
+        values = super().value_from_datadict(data, files, name)
+        return "".join(values)
+
+
+class TokenForm(forms.Form):
+    token = forms.CharField(widget=VotingTokenWidget())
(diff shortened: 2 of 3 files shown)


meinberlin/apps/votes/forms.py
```diff
@@ -4,9 +4,29 @@
...
+        return [None, None, None]
 
-    token = forms.CharField(max_length=40)
+    def value_from_datadict(self, data, files, name):
+        values = super().value_from_datadict(data, files, name)
+        return "".join(values)
+
+
+class TokenForm(forms.Form):
+    token = forms.CharField(max_length=40, widget=VotingTokenWidget())
```
sabinammm, line 29: The max_length validation doesn't work well with the three fields if it is shortened to 4. When it is shortened, each field really only allows 4 characters to be entered, but since we have a total of 12, this is in contradtion.

meinberlin/apps/votes/forms.py
```diff
@@ -4,9 +4,29 @@
 from meinberlin.apps.votes.models import VotingToken
 
 
-class TokenForm(forms.Form):
+class VotingTokenWidget(forms.MultiWidget):
+
+    def __init__(self, attrs=None):
+        widgets = (
+            forms.TextInput(attrs={'size': '4', 'maxlength': '4'}),
```
sabinammm, line 11: While this limits the characters actually used by the field, the styling still looks aweful and this contradcits the max_length=40 defined in line 29

Rineee:
@sabinammm You were right that the maxlength from the widgets were ignored, because of the maxlength on the form field.. I found [this](https://stackoverflow.com/questions/49349907/django-doesnt-respect-the-maxlength-widget-attribute) issue, that wasnt answered, but pointed me in the right direction. I think it shouldnt be like that, but anyway it is and I now removed the maxlength from the form field to make it work. I also added minlength to the widgets, not sure if we then also need size or is that for styling, do you know?
I added the changes in another commit, so you could look at them, maybe you could squash before merging?

meinberlin/apps/votes/forms.py
```diff
@@ -4,9 +4,35 @@
...
+                                   'maxlength': '4'}),
+            forms.TextInput(attrs={'size': '4',
+                                   'minlength': '4',
+                                   'maxlength': '4'}),
+        )
+        super(VotingTokenWidget, self).__init__(widgets, attrs)
+
+    def decompress(self, value):
+        if value:
+            n = 4
```
Rineee, line 25: I think you could remove this and simply write 4 in the line below?

sabinammm: Yes, I thought about this, and then I wondered if we would ever change it and then it would be more explicit. But for it to be more explicit, it needs to be more of a telling variablename than simply 'n' I guess. I will change it! Thanks:)

sabinammm:
> @sabinammm You were right that the maxlength from the widgets were ignored, because of the maxlength on the form field.. I found [this](https://stackoverflow.com/questions/49349907/django-doesnt-respect-the-maxlength-widget-attribute) issue, that wasnt answered, but pointed me in the right direction. I think it shouldnt be like that, but anyway it is and I now removed the maxlength from the form field to make it work. I also added minlength to the widgets, not sure if we then also need the size attribute or is that for styling, do you know? I added the changes in another commit, so you could look at them, maybe you could squash before merging?

Yes, perfect!! Thanks!

phillimorland:
@sabinammm once you're happy with your code, let me know and i'll take over the styling :)

sabinammm:
> @sabinammm once you're happy with your code, let me know and i'll take over the styling :)

This is fine now and ready for styling. Thank you :)

meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
```diff
@@ -26,11 +26,11 @@
             <form class="" action="{{ request.path }}" method="post">
                 <div class="form-group">
                     {% csrf_token %}
-                    <div class="widget widget--{{ form.token|widget_type }}">
+                    <div class="widget widget--{{ form.token|widget_type }} widget--vote">
```
phillimorland, line 29: @sabinammm ok styling was more annoying then complicated but I had to add a class name here as there is nothing rendered from `widget_type` don't know why though?

phillimorland (approved):
very nice!

PR #4041, opened 2021-12-07
apps/projects plans /contrib /ideas /mapideas /cms: added missing html_tag for…

### LINKED ISSUE #4017: should we use html_date everywhere?
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** We do use our own template tag `html_date` in quite a few places, but seem to have forgotten in others (like here https://github.com/liqd/a4-meinberlin/blob/fb7aaf29e24bc904383e2521399548d94c4bac24/meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html#L92). I think we need to decide on one format and use that.
**important screensize:**
**device & browser:** 
**Comment/Question:** The main question to answer is: is the `html_date` better for a11y, because it wraps the dates in a html time element? @phillimorland do you know? https://github.com/liqd/a4-meinberlin/blob/23922a3797db0e06b6de6cd1aa87bc73224cbc17/meinberlin/apps/contrib/templatetags/contrib_tags.py#L44

Screenshot?

sabinammm:
… accessibility to dates

Fixes #4017

In order to find out which templates should use the **html_tag** for dates, the code was searched for the date filter. 
6 of the 7 templates changed had very similar lines that were changed. Only one was changed from a `blocktrans` to `trans` because  adding the **html_tag** to the the blocktrans would break it.

### DIFF
--- a/meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
+++ b/meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
@@ -50,7 +50,11 @@
             <div class="item-detail__meta lr-bar">
                 <div class="lr-bar__left">
                     <strong class="item-detail__creator">{{ object.creator.username }}</strong>
-                    {% if object.modified %}{% trans 'updated on ' %}{{ object.modified|date }}{% else %}{% trans 'created on ' %}{{ object.created|date }}{% endif %}
+                    {% if object.modified %}
+                        {% trans 'updated on ' %}{% html_date object.modified class='list-item__date' %}
+                    {% else %}
+                        {% trans 'created on ' %}{% html_date object.created class='list-item__date' %}
+                    {% endif %}
                 </div>
                 <div class="lr-bar__right">
                     <strong>{% trans 'Reference No.' %}:</strong>
--- a/meinberlin/apps/cms/templates/meinberlin_cms/includes/storefronts.html
+++ b/meinberlin/apps/cms/templates/meinberlin_cms/includes/storefronts.html
@@ -2,14 +2,30 @@
 {% load static contrib_tags %}
 
 {% if item.district %}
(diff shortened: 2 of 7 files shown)


meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
```diff
@@ -50,7 +50,7 @@ <h1 class="item-detail__title">{{ object.name }}</h1>
             <div class="item-detail__meta lr-bar">
                 <div class="lr-bar__left">
                     <strong class="item-detail__creator">{{ object.creator.username }}</strong>
-                    {% if object.modified %}{% trans 'updated on ' %}{{ object.modified|date }}{% else %}{% trans 'created on ' %}{{ object.created|date }}{% endif %}
+                    {% if object.modified %}{% trans 'updated on ' %}{% html_date object.modified class='list-item__date' %}{% else %}{% trans 'created on ' %}{% html_date object.created class='list-item__date' %}{% endif %}
```
khamui, line 53: maybe we could break this very long lines into shorter blocks? just for readability. what do you think?

khamui (commented):
very nice! working fine.

Note: we would need this for the new react list as well. Will make a note somewhere.

meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
sabinammm: Yes, that's a good idea. I will change it.

sabinammm:
> very nice! working fine. just a little comment :)
> 
> Note: we would need this for the new react list as well. Will make a note somewhere.

@khamui I was just writing an issue b/c I noticed the same for the Comments in a4.

sabinammm:
@khamui I shortened the changed code lines in commit 1. Commit 2 I shortened just the way too long lines in these files and changed some indents.

khamui:
yeah! i think it makes sense to improve readability whenever possible, so fine if you shortened lines which are not related to the task, but improves readability.

PR #4048, opened 2021-12-08
apps/budgeting/proposal-list-item: showing modified or created date

### LINKED ISSUE #4042: (react list) proposal-list-item should show created/modified date
**URL:** proposal-list
**user:** any
**expected behaviour:** list items (tiles) in react list should show `created on/modified on` in front of date (see django list)
**behaviour:** list items do not indicate created on or modified on, but shows only the date
**important screensize:** any
**device & browser:** any
**Comment/Question:** 

Screenshot?

Note: This will only affect where this particular react list is used (currently only used for budgeting/proposal-list)

khamui:
add modified to serializer
using new modified filed to distinguish in frontend
desctructured proposal to p

fixes #4042

### DIFF
--- a/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
+++ b/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
@@ -6,36 +6,36 @@ import { ListItemBadges } from './ListItemBadges'
 import { ListItemStats } from './ListItemStats'
 
 export const BudgetingProposalListItem = (props) => {
-  const { proposal: p, isVotingPhase } = props
-  const safeLocale = p.locale ? p.locale : undefined
-  const date = p.modified
-    ? `${django.gettext('modified on')} ${toLocaleDate(p.modified, safeLocale)}`
-    : `${django.gettext('created on')} ${toLocaleDate(p.created, safeLocale)}`
+  const { proposal, isVotingPhase } = props
+  const safeLocale = proposal.locale ? proposal.locale : undefined
+  const date = proposal.modified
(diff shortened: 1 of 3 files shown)


meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
```diff
@@ -6,42 +6,46 @@ import { ListItemBadges } from './ListItemBadges'
 import { ListItemStats } from './ListItemStats'
 
 export const BudgetingProposalListItem = (props) => {
-  const { proposal, isVotingPhase } = props
-  const safeLocale = proposal.locale ? proposal.locale : undefined
+  const { proposal: p, isVotingPhase } = props
+  const safeLocale = p.locale ? p.locale : undefined
```
phillimorland, line 10: I don't understand what you're doing here, why change it to p?

phillimorland (commented):
nice solution for date, not sure on refactor?!

meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
khamui: just to keep the following lines (12 and 13) in one line.
think with `p` it is still readable or even more readable, no?

khamui: i could also do it with, so we can get rid of the `p` entirely.

```
  const  {
    name,
    locale,
    ... } = proposal

phillimorland: Thanks for adding proposals back! I was mostly weirded out with the reassignment in const, I would come back in a month and wonder why it's there :p

PR #4051, opened 2021-12-09
apps/budgeting: add vote button component and template tag and load v…

phillimorland:
…ia app js

### DIFF
--- a/meinberlin/apps/budgeting/assets/VoteButton.jsx
+++ b/meinberlin/apps/budgeting/assets/VoteButton.jsx
@@ -1,0 +1,17 @@
+import React from 'react'
+import django from 'django'
+import CheckboxButton from '../../contrib/assets/CheckboxButton'
+
+export const VoteButton = (props) => {
+  return (
+    <>
+      <CheckboxButton
+        onText={django.gettext('Voted')}
+        offText={django.gettext('Give my vote')}
+        onClass="btn btn--full"
+        offClass="btn btn--full btn--light"
+        uniqueID={123}
+      />
+    </>
+  )
+}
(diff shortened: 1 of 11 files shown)


meinberlin/apps/budgeting/assets/react_vote_button_init.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { widget as ReactWidget } from 'adhocracy4'
+import { VoteButton } from './VoteButton.jsx'
+
+function init () {
+  ReactWidget.initialise('mb', 'vote_button',
+    function (el) {
+      const props = el.getAttribute('data-attributes')
+      ReactDOM.render(<VoteButton {...props} />, el)
```
khamui, line 10: somehow the passed props is a string, but should be an object.

meinberlin/apps/budgeting/assets/VoteButton.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import django from 'django'
+import CheckboxButton from '../../contrib/assets/CheckboxButton'
+
+export const VoteButton = (props) => {
```
khamui, line 5: I guess you have to wrap it to use as a widget?
Would it make sense to use the `VoteButton` for the list item as well?

khamui (commented):
cool! Button is there, just one thing below.

meinberlin/apps/budgeting/assets/VoteButton.jsx
phillimorland: yes it would

meinberlin/apps/budgeting/assets/react_vote_button_init.jsx
phillimorland: ok thats weird, will take a look

khamui: > somehow the passed props is a string, but should be an object.

turned out, this does not happen.

khamui:
cool! working fine

PR #4053, opened 2021-12-09
Kt 2021 12 js tests

khamui:


meinberlin/apps/budgeting/assets/__tests__/BudgetingProposalList.jest.jsx
```diff
@@ -1,42 +1,80 @@
...
-//     creator: 'creator part 2',
-//     created: '2021-10-29T09:33:36.280033+01:00'
-//   }
-// ]}
-
-// Mocks the fetch request
-global.fetch = jest.fn(() =>
-  Promise.resolve({
-    json: () => Promise.resolve({ data: {} })
+test('Budgeting Proposal List without list item (empty)', async () => {
```
khamui, line 5: @phillimorland finally i found a solution for the `...a test was not wrapped in act(...).` we had with the test of `BudgetingProposalList.jsx` . 

Source: https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning#the-dreaded-act-warning

Three cases are testing the list here:
- If no item was fetched
- If one (or more) items are fetched
- If a network error occured, while fetching

With these tests I think the snapshot of the BudgetingList is not needed anymore. So I deleted it. Hope that is fine.

phillimorland (commented):
Looking good, I think I get most of it but maybe @goapunk can take a look before merge?

meinberlin/apps/contrib/assets/__tests__/CountDown.jest.jsx
```diff
@@ -0,0 +1,29 @@
...
+      inactiveClass="inactive-class"
+      countText="votes"
+      counter={0}
+    />
+  )
+  const countComponent = screen.getByText('votes')
+  expect(countComponent.innerHTML).toBe('votes')
+})
+
+test('CountComponent is showing active', () => {
```
goapunk, line 18: these two tests seem extremely similar, would it maybe make sense to do a `getByText('0')` and `getByText('2')` also check that the count is correct?

goapunk: cool, thanks

meinberlin/apps/contrib/assets/__tests__/CountComponent.jest.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+import { CountComponent } from '../CountComponent'
+
+test('CountComponent is showing', () => {
+  render(
+    <CountComponent
+      activeClass="active-class"
+      inactiveClass="inactive-class"
+      countText="votes"
```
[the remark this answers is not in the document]
khamui: actually i thought that first too, but actually we are not testing the output/render of the class but the logic.
so whatever string is used for active and inactive should be applied, depending on the counters value (0 or > 0). Thats the only thing being tested here.

E.g. if we want to make sure this component looks exactly how it should look we would need to add a snapshot test. Meaning whenever we change the class name or adding new classes to it we would need to update the test.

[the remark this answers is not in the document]
phillimorland: coolio! I think you'll need to give me a rundown on these tests

phillimorland (approved):
very nice! :)

PR #4054, opened 2021-12-13
budgeting/templates/proposal-list: only show token form for 3 phase m…

### LINKED ISSUE #4044: frontend idea list in pb with 1+2 phases: Code field is shown
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-spandau/
**user:** any
**expected behaviour:** I cannot enter voting code in projects without voting
**behaviour:** in modules participatory budgeting with 1 and 2 phases the code field is shown
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2021-12-08 um 12 23 40

Rineee:
…odules

fixes #4044

### DIFF
--- a/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
+++ b/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
@@ -19,21 +19,27 @@
             {% trans 'Submit proposal' %}
         </a>
     {% endif %}
-    {% has_perm 'meinberlin_budgeting.vote_proposal' request.user module as vote_allowed %}
-    {% if vote_allowed %}
-        {% if request.session.voting_token %}
-            You are using token {{ request.session.voting_token }}
-        {% else %}
-            <form class="" action="{{ request.path }}" method="post">
-                <div class="form-group">
-                    {% csrf_token %}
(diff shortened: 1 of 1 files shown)


fuzzylogic2000 (commented):
Hmmm. But this only occured for admins anyway, right? Still confusing, though. But if we do it in this slightly weird way, maybe instead of counting the phases, we can use the type and check if we have voting in the phases? But that is still weird, so we can also merge like this. Whatever you like more.

Rineee:
> Hmmm. But this only occured for admins anyway, right? Still confusing, though. But if we do it in this slightly weird way, maybe instead of counting the phases, we can use the type and check if we have voting in the phases? But that is still weird, so we can also merge like this. Whatever you like more.

Well it only occurs for admins, because checking the permission always returns true for admins. For other users it checks the meinberlin_budgeting.vote_proposal, which is true if voting phase is active, so it doesnt show in 2 phase budgeting.

So I think we need to somehow know in the template if its a 2 or 3 phase? Or check if voting is in the phases, yes, thats also a possiblity, but I thought this was easier and kind of the same outcome? But if you prefer, I will also change it!

fuzzylogic2000 (approved):
yeah!

PR #4055, opened 2021-12-13
templates: blocktrans blocks refactored in all apps where these were not in 1 line

sabinammm:
Translations in po-files from blocktrans blocks contain linebreaks with \n in them, if they are not in 1 line in the templates.

Thus, in this PR, all blocktrans blocks in the project were checked to see whether they ran over more than one line. 

In the 1st commit, these were refactored to close on the same line. In addition, in order to _reinstate_ a necessary blocktranslate in the **cms//storefronts template**, the html_date tag was removed, as it is not possible to use this in a blocktrans block.

The 2nd and 3rd commits are the refactored as well as new translations.

### DIFF
--- a/locale/de_DE/LC_MESSAGES/django.po
+++ b/locale/de_DE/LC_MESSAGES/django.po
@@ -310,7 +310,7 @@ msgstr "hier können Sie alle Ideen exportieren"
 #: adhocracy4/exports/templates/a4exports/export_dashboard.html:31
 #: adhocracy4/exports/templates/a4exports/export_dashboard.html:31
 msgid "here you can export all subjects"
-msgstr ""
+msgstr "hier können Sie alle Themen exportieren"
 
 #: adhocracy4/exports/templates/a4exports/export_dashboard.html:43
 #: adhocracy4/exports/templates/a4exports/export_dashboard.html:43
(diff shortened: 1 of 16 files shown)


locale/de_DE/LC_MESSAGES/django.po
```diff
@@ -2114,30 +2113,29 @@ msgstr[1] "%(counter)s Projekte anzeigen"
 msgid "City wide"
 msgstr "Gesamtstädtisch"
 
-#: meinberlin/apps/cms/templates/meinberlin_cms/includes/storefronts.html:37
-#: meinberlin/apps/projects/templates/meinberlin_projects/includes/project_tile.html:65
+#: meinberlin/apps/cms/templates/meinberlin_cms/includes/storefronts.html:55
 #, python-format
-msgid " %(time_left)s remaining "
-msgstr " noch %(time_left)s"
+msgid "%(time_left)s remaining"
```
fuzzylogic2000, line 2118: Here it's easy to see why we need the blocktrans to include the variable. :)

fuzzylogic2000 (approved):
Very nice!
I will merge now, not to loose your translations. Could you please still check afterwards:
1. if using the variable and with that the date outside of blocktrans works and thus we can use the date with the blocktrans
2. if we need the newline in the participant invite form (and because moderator and participant invite are normally nearly the same, what we do in the moderator invite).
(I do not know if it works or if we need the newline, just try it out!) Thank you!

PR #4062, opened 2021-12-14
apps/budgeting: add filter info and choices to proposal API

fuzzylogic2000:
Alright, this is the filter info. 
@khamui Can you do with that list or do you need another format? The icons for example might be better in a dict? Or does that not matter for React anyway?
And @Rineee I hope, you don't hate it too much...


khamui:
> 

nice! that would work fine 👍🏽

About the icon, would it be possible to get the location/path to the icon file?
Currently i am getting a list `["4", "halbkreis"]` or is there a robust way to generate that href in frontend?

I am merging this now so I can work on the filters @fuzzylogic2000

PR #4063, opened 2021-12-14
apps/budgeting/api: more info in API and serializer

fuzzylogic2000:
@khamui I think that is all the info we need in the API now! :) And quite a lot of refactoring for the frontend, then...

### DIFF
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -7,6 +7,8 @@ from rest_framework.pagination import PageNumberPagination
 
 from adhocracy4.api.mixins import ModuleMixin
 from adhocracy4.api.permissions import ViewSetRulesPermission
+from adhocracy4.categories import get_category_icon_url
+from adhocracy4.categories import has_icons
 from adhocracy4.categories.models import Category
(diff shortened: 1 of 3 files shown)


meinberlin/apps/budgeting/api.py
```diff
@@ -28,6 +29,13 @@ def get_paginated_response(self, data):
         return response
 
 
+class LocaleInfoMixin:
```
khamui, line 32: nice

meinberlin/apps/budgeting/serializers.py
```diff
@@ -64,12 +63,7 @@ def get_url(self, proposal):
...
         if hasattr(proposal, 'moderator_feedback'):
-            return proposal.moderator_feedback
-        else:
-            return None
-
-    def get_moderator_feedback_choices(self, proposal):
-        if hasattr(proposal, 'moderator_feedback_choices'):
-            return proposal.moderator_feedback_choices
+            return (proposal.moderator_feedback,
+                    proposal.get_moderator_feedback_display())
```
khamui, line 67: better now :)

khamui:
nice!

PR #4067, opened 2021-12-16
Ks 2021 12 voting api

Rineee:
The permissions are not yet there, so this is currently only really working for logged in users with a valid token (for not logged in users, the deletion is not allowed).
 Also the count of votes left is not there yet as I wanted to wait if updating of the proposal list will work/how it will work and then I would it add it to response.data of proposals serializer. Do you think that is a good solution @fuzzylogic2000 ?

@kha to test, you need a valid token. You can go to django admin and create one for your module. The annoying thing is that in the list of modules you can choose from  only the internal name of the modules are displayed. To find the right one, you can check under modules what the name is.
As you know, the re-rendering doesnt work yet and also I am using this helper fuction from livequestions, that we could maybe use, but then move to contrib or write our own?

My todo is:

- [ ] permission to allow only five votes
- [ ] delete permission should allow not logged in users
- [x] add number of votes in appropriate place
- [ ] return proper response if api is called without token

I dont know if we want to merge this already and go on from there or keep on working in branches? Up to you :)


fuzzylogic2000 (approved):
Supercool!

PR #4068, opened 2021-12-16
[#50499] restrict moderation to not change projects

sabinammm:
This PR is in regard to [US 50499](https://taiga.liqd.net/project/liqd-liquid-software/us/5049?milestone=290).

The meinBerlin-specific rules for changing projects were removed, leaving the original a4 permissions that do not specify moderators.
The tests were changed according to this new permission setup.


fuzzylogic2000 (approved):
Nice!

PR #4072, opened 2021-12-16
apps/budgeting/proposal-list/filters: adjusting filters to new backend info

khamui:


meinberlin/apps/budgeting/assets/FilterBar.jsx
```diff
@@ -36,6 +36,29 @@ export const FilterBar = props => {
...
+    filter.current = selectedFilter(filterType) || defaultFilter
+    return filter
+  }
+
+  const selectedFilter = filterType =>
+    filterObject && filterObject[filterType] && filterObject[filterType][1]
+
+  const getDefaultFilter = filter => {
+    // Return Early, if no default, take "All"-choice
+    if (!filter.default) return filter.choices[0][1]
```
fuzzylogic2000, line 51: Then we also have to make sure that the "all" choice is always the first choice, right? That is not the case for the is_archived filter... But as the default is something else there and should be first in the filter dropdown, I think that is alright.

fuzzylogic2000: And instead of thinking that the all-choice is always first, it does make sense to always put the default choice first (for example also for the ordering filter), so that shouldn't go too wrong. :)

fuzzylogic2000 (approved):
Works great and looks good!
There is only one thing: the icons are only displayed for categories that exist in the list. 
Screenshot from 2021-12-17 09-28-48

But I will add another task and merge. :)

meinberlin/apps/budgeting/assets/FilterBar.jsx
khamui: oh yes, that was an assumption. thanks for pointing out.

PR #4074, opened 2021-12-16
apps/budgeting/voting: voting button updating after request and helpers function shifted to contrib

khamui:
@Rineee regarding the `left votes count` i think we can make it in the same way. meaning directly using the props instead of having it wrapped in `this.state`. But happy to pair program on this as well :)


Rineee (approved):
Super cool, thank you!!

PR #4075, opened 2021-12-20
apps/budgeting/proposal-list-item/voting: fixing glitch with voting (…

khamui:
…POST/DELETE)

It seems that this fixes the glitch with the button not updating instantly.


Rineee (approved):
Nice!

PR #4088, opened 2021-12-21
apps/budgeting/js-tests: FilterBar.jsx and FilterBarDropdown.jsx

khamui:


fuzzylogic2000 (approved):
Looks good to me, tests the filters and ups the coverage! :tada:

khamui:
Note: just to mention that the coverage of the `FilterBar` itself is not very high (something around 75%). Problem is that we cannot/should not test internal functions. I will have a research how to handle this better. But definitely not a prio.

PR #4092, opened 2021-12-21
apps/budgeting: switch category and archived filters

### LINKED ISSUE #4094: testing 5024: order of filter
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-spandau/?mode=list
**user:** any
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** the order of the filter was 1) category 2) archive. It should stay this way but as we will touch the filter again later we can also do it then but need to remember. So if easy maybe change now?

Screenshot?
Bildschirmfoto 2021-12-21 um 16 25 15

fuzzylogic2000:
To make the filters displayed in the same order as in the django template list

fixes #4094

### DIFF
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -56,16 +56,6 @@ class ProposalFilterInfoMixin(ModuleMixin):
             'default': '-created',
         }
 
-        filters['is_archived'] = {
-            'label': _('Archived'),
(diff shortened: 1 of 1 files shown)


khamui:
i thought i screwed in in frontend with the latest PR, than i saw this! 🙏

PR #4106, opened 2021-12-21
Ks 2021 12 token voting permissions

Rineee:
I did some more research and decided to not use rules for the token voting api permissions as rules are dependent on users and arent really useful here. 
I instead introduced a new Permission class, called TokenVotePermission that checks manually the relevant requirements on the token for create and destroy actions.
Also, I added a check in TokenVoteMixin to make sure that the token is valid before anything else is checked.

### DIFF
--- a/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
+++ b/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
@@ -41,7 +41,7 @@ export const BudgetingProposalListItem = (props) => {
           <span className="list-item__author">{proposal.creator}</span>
           {date}
         </div>
-        {isVotingPhase && (
+        {proposal.vote_allowed && (
           <VoteButton
             objectID={proposal.pk}
             tokenvoteApiUrl={tokenvoteApiUrl}
--- a/meinberlin/apps/budgeting/predicates.py
+++ b/meinberlin/apps/budgeting/predicates.py
@@ -1,0 +1,8 @@
+import rules
+
+from adhocracy4.phases.predicates import phase_allows_vote
+
+
+@rules.predicate
+def is_allowed_vote_proposal(user, proposal):
+    return (not proposal.is_archived) and phase_allows_vote(user, proposal)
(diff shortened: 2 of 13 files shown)


Rineee:
@fuzzylogic2000 Now I remember why I also wanted to add the phase to the serializer, we also use the isVotingPhase for not showing the rating here: https://github.com/liqd/a4-meinberlin/blob/d69a4dd624705c7f28a053e123e0dcdf847fbe3d/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx#L23 
Not sure how to deal with it if not with serializer info, do you have an idea?

fuzzylogic2000:
> @fuzzylogic2000 Now I remember why I also wanted to add the phase to the serializer, we also use the isVotingPhase for not showing the ratings here:
> 
> https://github.com/liqd/a4-meinberlin/blob/d69a4dd624705c7f28a053e123e0dcdf847fbe3d/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx#L23
> 
> Not sure how to deal with it if not with serializer info, do you have an idea?

So, the decision was to just keep it in the template tag for now, right?

Rineee:
> > @fuzzylogic2000 Now I remember why I also wanted to add the phase to the serializer, we also use the isVotingPhase for not showing the ratings here:
> > https://github.com/liqd/a4-meinberlin/blob/d69a4dd624705c7f28a053e123e0dcdf847fbe3d/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx#L23
> > 
> > Not sure how to deal with it if not with serializer info, do you have an idea?
> 
> So, the decision was to just keep it in the template tag for now, right?

Yes! Should I write an issue to think about moving it to the api?

phillimorland (commented):
install and passing tests but vote button functioonality no longer working, is this expected?

Rineee:
> install and passing tests but vote button functioonality no longer working, is this expected?

@phillimorland No, that should work still! What does it say in the network tab?

phillimorland:
@Rineee ignore me, I forgot I had to re-enter my token :p however I will add that we must disable hover styling on the disabled vote button

PR #4111, opened 2021-12-22
[#5049] Kl 2021 12 restrict moderators

fuzzylogic2000:
This is for the story restricting the moderators from changing projects.
I am not supersure if we use and need the last commit (noone noticed that group members were not allowed the CK editor upload so far. But the tests look like this is right now.


phillimorland (approved):
nice!

Issue #4107, opened 2021-12-21
budgeting: use api instead of props through template tag in react_proposals_vote


Rineee:
**URL:** 
**user:** 
**expected behaviour:** consistent passing of information
**behaviour:** For now, we are using the proposal api to get all info on the proposal list, but on detail view put all information in the template tag* 
**important screensize:**
**device & browser:** 
**Comment/Question:** With the proposal pk, we could also just pass the the proposal detail api url in the [template tag](https://github.com/liqd/a4-meinberlin/blob/d69a4dd624705c7f28a053e123e0dcdf847fbe3d/meinberlin/apps/budgeting/templatetags/react_proposals_vote.py) and then fetch the proposal information from there. @khamui @phillimorland I dont know how much refactoring this would mean? We would still need to pass token info though or write another mixin that adds that info to proposal detail. 

 

Screenshot?

khamui:
yes, it gets messy quickly if we have to maintain two "bridges".
I will think about how we can unify that better in frontend, so we only go one way and keep the template tag attribute as small as possible.

btw: @Rineee you mentioned the other Kha again, sorry Kha...

khamui:
just assigning, so i do not forget about it

phillimorland:
@khamui maybe doing the refactor you can also add disabled behaviour for when user has used their 5 votes: https://taiga.liqd.net/project/liqd-liquid-software/task/5130

phillimorland:
NOTE: hover styling still present even when disabled, maybe due to it being a styled checkbox, maybe need to consider additional styl rules?

phillimorland:
@fuzzylogic2000 I just re-read the issue and i'm a little confused, do we still want to consider doing this?

fuzzylogic2000:
@phillimorland Yes, we have 
this for the list: https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/apps/budgeting/templatetags/react_proposals.py
and this for the detail view: https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/apps/budgeting/templatetags/react_proposals_vote.py
That is quite confusing. 

On the other hand, it's probaly not that important right now as we hope to get the detail view into React as well at some point and would need to redo the template tag for that anyway. Sooo, let's close this issue and do it when we work on it anyway?! Reopen if you diagree. @Rineee or @phillimorland

PR #4114, opened 2021-12-23
[DISCUSSION] centralizing api calls

khamui:
@phillimorland @fuzzylogic2000 @Rineee @sabinammm 
today i felt like testing around to unify how we make api calls from the frontend.
In a4 we have something like that, but i think using the context API is more elegant.

it can -maybe- solve the issue described here: #4107

What you guys think?

Note: it is just a draft, not all api request methods are implemented yet.

### DIFF
--- a/meinberlin/apps/budgeting/assets/VoteButton.jsx
+++ b/meinberlin/apps/budgeting/assets/VoteButton.jsx
@@ -1,64 +1,64 @@
 import React from 'react'
 import django from 'django'
-import { updateItem } from '../../contrib/assets/helpers.js'
+import { useApi } from '../../contrib/assets/ApiProvider'
 
-export default class VoteButton extends React.Component {
-  async addVote () {
+export const VoteButton = props => {
+  const api = useApi()
+
+  const addVote = async () => {
     const data = {
-      object_id: this.props.objectID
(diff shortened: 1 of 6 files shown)


meinberlin/apps/budgeting/assets/VoteButton.jsx
```diff
@@ -1,64 +1,64 @@
 import React from 'react'
 import django from 'django'
-import { updateItem } from '../../contrib/assets/helpers.js'
+import { useApi } from '../../contrib/assets/ApiProvider'
 
-export default class VoteButton extends React.Component {
-  async addVote () {
+export const VoteButton = props => {
```
khamui, line 5: it looks like a lot of changes, actually just converting from class notation to function notation.

phillimorland:
@khamui would you be ok to go through the pros and cons of this over what we currently do after the next stand up where we are all there? just so we're all on the same page, and we can then discuss

khamui:
> @khamui would you be ok to go through the pros and cons of this over what we currently do after the next stand up where we are all there? just so we're all on the same page, and we can then discuss

absolutely! lets do that next time we are all there.

phillimorland:
closing for now but have written issue to record conclusion

PR #4122, opened 2021-12-29
budgeting/assets/VoteButton: add tabindex so vote is tabable

phillimorland:


meinberlin/apps/budgeting/assets/VoteButton.jsx
```diff
@@ -47,6 +47,7 @@ export default class VoteButton extends React.Component {
         <label
           htmlFor={this.props.objectID}
           className={this.props.isChecked ? checkedClass : uncheckedClass}
+          tabIndex="0" // eslint-disable-line jsx-a11y/no-noninteractive-tabindex
```
phillimorland, line 50: @fuzzylogic2000 i'm a bit torn here, the label is actually interactive as we are using it implicitely here so the setting doesn't make sense but it's still maybe not best practice? although I think the issue is to do with element being within a list item as could not get it into tabbing order without this

fuzzylogic2000: Because we need to be able to tab to the label which we use to set the checkbox? I think that is fine!

fuzzylogic2000 (approved):

PR #4123, opened 2021-12-29
Ks 2021 12 votes tests

Rineee:
The tests for proposal api/serializers are still missing, I started with it, but not finished yet.. 

Sorry, I got a bit crazy with the tests for the voting api, but decided to rather write a new test for every user role as it was too annoying with the token saved in the session when doing the login/logout procedure..

### DIFF
--- a/meinberlin/apps/budgeting/assets/VoteButton.jsx
+++ b/meinberlin/apps/budgeting/assets/VoteButton.jsx
@@ -41,12 +41,13 @@ export default class VoteButton extends React.Component {
     const uncheckedText = django.gettext('Give my vote')
     const checkedClass = 'btn btn--full'
     const uncheckedClass = 'btn btn--full btn--light'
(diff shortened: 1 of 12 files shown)


fuzzylogic2000 (approved):
Nice! I don't think, you can go too crazy with tests. :)

fuzzylogic2000:
@sabinammm I guess you also found this:

```
E           AssertionError: assert ['Dieser Abst...ist ungültig'] == ['This token is not valid']
E             At index 0 diff: 'Dieser Abstimmungsschlüssel ist ungültig' != 'This token is not valid'
E             Use -v to get the full diff

tests/budgeting/test_views_integration.py:71: AssertionError
```

I will look into that!

fuzzylogic2000:
@sabinammm And the last commit should fix that! I will merge once the build has run.

PR #4126, opened 2022-01-03
[#5042] release translations

sabinammm:
Translation task for US #5042 prepare release.
This also fixes #4096

PR #4136, opened 2022-01-05
[#4126 (comment)] apps/budgeting: consolidate wording proposal list

sabinammm:
This is in reference to https://github.com/liqd/a4-meinberlin/pull/4126#discussion_r777473157
This change consolidates the wording use for the new react proposal list to use the same wording as is used in this context everywhere else (commit 1), and adds the corresponding translation (commits 2 & 3).


fuzzylogic2000 (approved):
Nice!

Issue #4150, opened 2022-01-11
missing translation project overview


CarolingerSeilchenspringer:
**URL:** https://mein.berlin.de/projekte/
**user:** any
**expected behaviour:** german
**behaviour:** english
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
![image002 Kopie](https://user-images.githubusercontent.com/35491681/148986610-14ed30cb-38d0-4668-9c23-4fcf18218e67.png)

Issue #4149, opened 2022-01-11
missing translations interactive event


CarolingerSeilchenspringer:
**URL:** https://mein.berlin.de/projekte/module/interaktive-veranstaltung-23/
**user:** any
**expected behaviour:** german
**behaviour:** english
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
<img width="336" alt="Bildschirmfoto 2022-01-11 um 16 47 07" src="https://user-images.githubusercontent.com/35491681/148975033-0e1fdb96-3549-48a4-9cf5-e7b127e61dfd.png">

<img width="619" alt="Bildschirmfoto 2022-01-11 um 16 48 44" src="https://user-images.githubusercontent.com/35491681/148975254-4d1ba06f-6318-4b0b-b4cf-f31e6009d557.png">

PR #4151, opened 2022-01-12
apps/livequestions apps/plans: fixed missing translations by moving translation strings into variables at top of files in all affected jsx files

sabinammm:
Fixes Issues #4150  and #4149 as caused by PR #4126

Leaving this in WIP until [question](https://github.com/liqd/a4-meinberlin/pull/4126/files#r783143986) is resolved, i.e. if there is another PR besides #4126 that deleted translations that are actually still needed.

### DIFF
--- a/meinberlin/apps/plans/assets/PlansList.jsx
+++ b/meinberlin/apps/plans/assets/PlansList.jsx
@@ -2,6 +2,14 @@
 const React = require('react')
 const Moment = require('moment')
 
+const remainingStr = django.gettext('remaining')
+const moreThanStr = django.gettext('More than 1 year remaining')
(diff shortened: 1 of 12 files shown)


meinberlin/apps/plans/assets/PlansList.jsx
```diff
@@ -2,6 +2,14 @@
 const React = require('react')
 const Moment = require('moment')
 
+const remainingStr = django.gettext('remaining')
```
fuzzylogic2000, line 5: nice, that you also moved these to the front!

fuzzylogic2000 (approved):
Nice!

PR #4168, opened 2022-01-25
django-admin: fix custom CK editor fields

sabinammm:
Similar as fix for [#1656 in a+](https://github.com/liqd/adhocracy-plus/issues/1656), done [here](https://github.com/liqd/adhocracy-plus/pull/1664/commits/6c9be35bb186c8cd504f492255f866dd224ff95c). In meinBerlin, this fix was required for the django-admin forms for

- Projects
- Activities
- Offline Events
- Plans.


fuzzylogic2000 (approved):
Nice!

PR #4172, opened 2022-01-27
Move translations in JSX-Files to variables

sabinammm:


locale/de_DE/LC_MESSAGES/djangojs.po
```diff
@@ -1027,42 +1019,42 @@ msgstr "Beantwortete Fragen"
...
 msgstr "Projekte anzeigen"
 
-#: meinberlin/apps/plans/assets/FilterNav.jsx:330
+#: meinberlin/apps/plans/assets/FilterNav.jsx:14
 msgid "more filters"
 msgstr "weitere Filter"
 
+#: meinberlin/apps/plans/assets/FilterNav.jsx:16
+msgid " in the area of "
+msgstr "und dem Bereich"
```
fuzzylogic2000, line 1056: Why do we need the spaces in the original string, but not in the translation? That's something that could be fixed with this PR! (If we don't need the spaces, also remove them in the original string.)

fuzzylogic2000 (approved):
Looks good, just one very small comment for a drive-by fix.

fuzzylogic2000 (approved):

PR #4184, opened 2022-02-17
apps/plans/forms//modal//plan_form: add form field to modal, adapt in…

phillimorland:
… form and add to template -no migrations


meinberlin/apps/plans/forms.py
```diff
@@ -49,6 +50,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields['district'].empty_label = _('City wide')
         self.fields['contact_address_text'].widget.attrs['rows'] = 6
+        self.fields['participation_explanation'].widget.attrs['rows'] = 1
```
Rineee, line 53: as I understood, it should look like address field, so also 6 rows here?

phillimorland: address field reference was just to ensure linebreaks so it should actually go in the user side story, in designs it is just one line but can be expanded

fuzzylogic2000 (approved):
Nice! 
I will fix the tests!

phillimorland:
@fuzzylogic2000 ahh right tests, thanky you!

PR #4188, opened 2022-02-18
[#5404] apps/organisations: make organisation slugs editable in django-form

sabinammm:
[](https://taiga.liqd.net/)#5404
[mB] make Organisation and project URLs changeable via django


meinberlin/apps/organisations/forms.py
```diff
@@ -24,6 +24,7 @@ class Meta:
         }
 
     def clean(self):
+        self._validate_unique = True
```
fuzzylogic2000, line 27: @sabinammm That was hard to find. Here in mB we use this form in the django-admin to validate the groups. In the clean-method of django, the cleaned_data is returned as well, but first, it's validated that the fields are unique. https://github.com/django/django/blob/7c318a8bdd66f8c5241864c9970dddb525d0ca4c/django/forms/models.py#L432 I added that to our method overwriting the clean-method as well.

meinberlin/apps/organisations/models.py
```diff
@@ -9,7 +9,14 @@
 
 
 class Organisation(models.Model):
-    slug = AutoSlugField(populate_from='name', unique=True)
+    slug = AutoSlugField(
+        populate_from='name',
+        unique=True,
+        editable=True,
+        help_text=_(
```
fuzzylogic2000, line 16: So, now you can remove the help text again and redo the migration. :)

fuzzylogic2000 (commented):
Could you remove the second commit, so that we can do that with the update of a4 in a separate PR?

meinberlin/apps/organisations/models.py
sabinammm: Awesome:) Thank you! I will do this and remove the second commt and put it into a different PR

meinberlin/apps/organisations/forms.py
sabinammm: :pray:

fuzzylogic2000 (approved):
Nice!

PR #4189, opened 2022-02-18
Ks 2022 02 participation filter

Rineee:


meinberlin/apps/projects/serializers.py
```diff
@@ -153,9 +152,6 @@ def get_status(self, instance):
     def get_participation(self, instance):
         return Plan.PARTICIPATION_CONSULTATION
 
-    def get_participation_display(self, instance):
-        return _('Yes')
-
```
Rineee, line 157: @fuzzylogic2000 I could not find that we use this somewhere, so I removed it?

sabinammm (approved):
Looks lovely and works great!

fuzzylogic2000 (approved):
NIce!

PR #4191, opened 2022-02-22
documentation: adding howto testing emails in mB

khamui:


fuzzylogic2000 (approved):
Yeah! I think this is it!

PR #4192, opened 2022-02-22
[#5404] apps/projects: show editable project slug field in django-form

sabinammm:
Depends on a4 PR[ #992](https://github.com/liqd/adhocracy4/pull/992)

### DIFF
--- a/meinberlin/apps/projects/admin.py
+++ b/meinberlin/apps/projects/admin.py
@@ -38,7 +38,7 @@ class ProjectAdminForm(forms.ModelForm):
 class ProjectAdmin(admin.ModelAdmin):
     form = ProjectAdminForm
     list_display = (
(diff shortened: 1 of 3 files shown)


fuzzylogic2000 (approved):
Nice!

sabinammm:
> Nice!



> Nice!

OH! Now it is :laughing: . Thanks for adding the hash here as well - now I know to update a4-dependent project-PRs like this in the same PR.

fuzzylogic2000:
> > Nice!
> 
> > Nice!
> 
> OH! Now it is laughing . Thanks for adding the hash here as well - now I know to update a4-dependent project-PRs like this in the same PR.

Yeah, normally we do that in an extra commit, but as we had so few changes here... :)

PR #4193, opened 2022-02-23
ci: upgrade container to ubuntu 20.04

goapunk:
* use ubuntu 20.04
* use python 3.8
* use postgresql 3.8


fuzzylogic2000 (approved):
:upside_down_face:

PR #4195, opened 2022-02-23
apps/plans: add participation field info to plan detail

fuzzylogic2000:
still to do:

- [x] add visually-hidden spans to explain more clearly ("toggle read more"/"toggle read less")
- [x] remove one of the borders between description and text

### DIFF
--- a/meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
+++ b/meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
@@ -83,9 +83,9 @@
                             <dt>{% trans 'Participation' %}</dt>
                             <dd>
                                 {{ object.get_participation_display }}
-                                {% if object.participation == object.PARTICIPATION_YES and object.published_projects.count %},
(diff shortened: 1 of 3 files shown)


fuzzylogic2000:
@phillimorland I didn't really know how to add an extra sr-only span and have it still make sense.

Also, I didn't know how to remove the double border best. There is a border on all sides on the new `item-detail-2__info-box` and a border-bottom on the `item-detail-2__content`, but we use that also elsewhere. Removing the upper border from the box seemed weird to me, because we would make that unusable elsewhere. But also adding another new class to have the content again without the border and only the margin seems weird. What would you do?

phillimorland:
@fuzzylogic2000 so about the border, the design removes the border around the content so I would just do that? as for the sr span I think I would do something like 
```
<span class="link--more">
<span class="sr-only">{% trans 'toggle to' %}</span>
{% trans 'Read more' %}
<span class="sr-only">{% trans ' about level of participation' %}</span>
</span>
```

phillimorland (approved):
coolio

Issue #4135, opened 2022-01-05
Initiators that are only group members and not intiators in other organisations, also have initiator rights for the other organisation's projects


fuzzylogic2000:
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** Initiators in an organisation can also create and change projects in another organisation, if they are group members of this organisation without explicitly being named as initiators in this organisation
**important screensize:**
**device & browser:** 
**Comment/Question:** https://github.com/liqd/a4-meinberlin/blob/3058cdf7bdfea3ebb785f197b30e09a66b1be897/meinberlin/apps/users/models.py#L80

Screenshot?

sabinammm:
No: initators who are group members in other orgs do not loose access to the dashboard.
BUT, according to manual testing and test written:
-A group member who is an initiator in group2 of orga2, but only member of group1 of orga1
-Not only has access to the dashboards of both organisations, but also can act as an initiator in both
Is that how it should be? Why would we have initiators if group members automatically have the rights of an initator?

@CarolingerSeilchenspringer this is a question you asked before: Is there a difference b/w Initiator and Organisation Member?
And according to this test, there is not.

fuzzylogic2000:
@sabinammm What exactly do you mean by act as initiator? The group member should be allowed to add projects. But shouldn't be allowed to change projects that do not belong to the group.

fuzzylogic2000:
@sabinammm @CarolingerSeilchenspringer I try to make this more clear also concerning the question about the difference.
So, initiators are allowed to add projects and they are allowed to change all projects from the organisation they are initiator in.
Members in organisation groups are also allowed to add projects and change these projects. They are also allowed to change projects belonging to the same group. But they shouldn't be allowed to change projects, that do not belong to the group.
So, the only difference should be that iniators are allowed to change all projects while group members are only allowed to change projects of their group.

@sabinammm So, the issue is that a group member, that is also initiator in another orga can see all projects of the orga she is group member in and also change them, even though these projects do not belong to the group she is in? Could you
1. change this issue (headline and description) accordingly?
2. try to reproduce that in a test?

CarolingerSeilchenspringer:
[...]

PR #4196, opened 2022-02-24
tests/users: add testing of initator rights in other organisations

sabinammm:
This test was updated to test initiator righs as discussed in #4135


fuzzylogic2000 (approved):
Nice!

PR #4201, opened 2022-02-24
Pm 2022 02 filter a11y tweaks

phillimorland:


phillimorland:
i'm not sure about all the naming here but my brain is not great today so will leave it and can be tested and I will maybe update tomorrow, but if you have suggestions please let me know :p

meinberlin/apps/contrib/assets/CheckboxBtn.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 
-export default class SwitchButton extends React.Component {
+export default class CheckboxBtn extends React.Component {
```
fuzzylogic2000, line 3: I cannot find the CheckboxBtn used anywhere, but also not the SwitchButton and am slightly confused

fuzzylogic2000 (approved):
Nice refactor and works like a charm!

PR #4202, opened 2022-02-24
Ks 2022 02 csrf failure template

### LINKED ISSUE #4170: 403 csrf error page should have same styling as other error pages
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** 
**important screensize:**
**device & browser:** 
**Comment/Question:** https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-CSRF_FAILURE_VIEW

https://docs.djangoproject.com/en/3.2/ref/csrf/

Screenshot?
Screenshot from 2022-01-25 16-36-16

Rineee:
Not super sure about this approach.. the request is aborted, so not there if csrf fails, thats why I had to add that check in userindicator. Also, because request is not there, it then shows the 'anmelden/register' button, even though the user is still logged in.. Not sure if we need (or even can) change that behaviour? Should I investigate more, what do you think @fuzzylogic2000 ?

fixes #4170

### DIFF
--- a/meinberlin/templates/403_csrf.html
+++ b/meinberlin/templates/403_csrf.html
@@ -1,0 +1,6 @@
+{% extends "403.html" %}
+{% load i18n %}
+
+{% block error_description %}
+    <p>{% trans 'The verification of your input has failed. Please reload the page and make sure that your browser settings allow setting session cookies.' %}</p>
+{% endblock %}


meinberlin/templates/403_csrf.html
```diff
@@ -0,0 +1,6 @@
+{% extends "403.html" %}
+{% load i18n %}
+
+{% block error_description %}
+    <p>{% trans 'CSRF verification failed. Request aborted.' %}</p>
```
fuzzylogic2000, line 5: Maybe we also say. "Please reload and try again."?

Actually that is a question for @CarolingerSeilchenspringer ! What should the text be when the csrf fails? For example if someone cannot log in or an idea form is sent, but cannot be saved because the csrf cookie is not there or not valid.

meinberlin/apps/users/templatetags/userindicator.py
```diff
@@ -26,5 +26,6 @@ def get_next_url(request):
 
 @register.inclusion_tag('meinberlin_users/indicator.html', takes_context=True)
 def userindicator(context):
-    context['redirect_field_value'] = get_next_url(context['request'])
+    if hasattr(context, 'request'):
+        context['redirect_field_value'] = get_next_url(context['request'])
```
fuzzylogic2000, line 30: Hmm, I don't get that. For me, the error page is already shown without that and also the Anmelden/Registrieren button is shown with or without that.

(I needed so long to get the error pages run locally, we should document that somewhere...
- set DEBUG=False and add ALLOWED_HOSTS = ['localhost'] in/to settings/dev.py
- start server with: python manage.py runserver --insecure
)

fuzzylogic2000 (commented):
I don't get it, sorry! Let's talk about it.

meinberlin/templates/403_csrf.html
CarolingerSeilchenspringer: @fuzzylogic2000 would this be correct: Die Authentifizierung Ihrer Eingabe ist fehlgeschlagen. Bitte laden Sie die Seite neu und stellen Sie sicher, dass Ihre Systemeinstellungen das Setzen eines Authentifizierungs-Cookies (Token) erlaubt. Bei Fragen können Sie sich an support@mein.berlin.de wenden.

Rineee:
> I don't get it, sorry! Let's talk about it.

Sure!

meinberlin/templates/403_csrf.html
fuzzylogic2000: Die Verifizierung Ihrer Eingabe ist fehlgeschlagen. Bitte laden Sie die Seite neu und stellen Sie sicher, dass Ihre Systemeinstellungen das Setzen von Session-Cookies erlaubt. Bei Fragen können Sie sich an [support@mein.berlin.de](mailto:support@mein.berlin.de) wenden.
@CarolingerSeilchenspringer That would be ok, as the CSRD verification can also fail when submitting an idea or any other form. Are you also fine with that?

fuzzylogic2000: @CarolingerSeilchenspringer Sorry, I thought we already use that email somewhere in a template, but is currently only in the settings and then used in the emails as sender. So, we would need to add that constant from the settings to a view or write a template tag just to include the email. So, we just skip that last sentence.

fuzzylogic2000 (approved):
I will just merge this first commit with that text now. :) 
The second commit to merge after getting a 500 is here: https://github.com/liqd/a4-meinberlin/pull/4204

PR #4203, opened 2022-02-25
tests/projects/rules: added test to show initiator vs group member rights

sabinammm:
Regarding Issue #4135
Added test to show that initiators in one organisation only have initator rights in another organisation, if they are group members of the specific project.

### DIFF
--- a/tests/projects/rules/test_rules_add_change_delete.py
+++ b/tests/projects/rules/test_rules_add_change_delete.py
@@ -100,3 +100,56 @@ def test_delete_project(user_factory, group_factory, organisation,
                               project)
     assert rules.has_perm(perm_name_delete, initiator, project)
     assert rules.has_perm(perm_name_delete, admin, project)
+
+
+@pytest.mark.django_db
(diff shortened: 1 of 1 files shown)


fuzzylogic2000 (commented):
I am a bit confused about what's what here. But I think, it proves that the issue actually isn't one? No, it doesn't, because group1 needs to belong to orga1 to prove that? I will rewrite this a bit.

sabinammm:
> I am a bit confused about what's what here. But I think, it proves that the issue actually isn't one? No, it doesn't, because group1 needs to belong to orga1 to prove that? I will rewrite this a bit.

My mistake: the initiator has to be taken out of group1 though. And then group1 is project1! Sorry, this is quite confusing...

fuzzylogic2000:
@sabinammm As you said, comments help! I hope... Could you have a look and then we have another talk? Also to talk about the issue, so you can show me, what's wrong on the page, now that we couldn't reproduce that in this test.

PR #4204, opened 2022-02-25
Ks 2022 02 csrf failure template 2

fuzzylogic2000:


Rineee:
Alright, csrf failure throws 500, so will merge this!

PR #4228, opened 2022-03-02
apps/bplan/tasks: fix UnicodeDecodeError to return None to be handled…

fuzzylogic2000:
… correctly

fixes https://sentry.liqd.net/organizations/liqd/issues/1948/

### DIFF
--- a/meinberlin/apps/bplan/tasks.py
+++ b/meinberlin/apps/bplan/tasks.py
@@ -36,7 +36,7 @@ def get_bplan_point_and_district_pk(bplan_identifier):
 
     except UnicodeEncodeError:
         # catches bplan-identifiers with problematic chars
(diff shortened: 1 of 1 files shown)


goapunk:
Looks good. Just wondering, do we catch those plans somewhere else so we can tell people to fix the name? or do we just silently drop it?

fuzzylogic2000:
> Looks good. Just wondering, what happens if we return (None, None)? will the bplan be just dropped?

No, it just does not add the point or the district for that bpland and makes sentry send the email that the identifier is wrong. Buuuut, there could and should be more error handling or the conditions should be a bit different and the whole setting and saving should go into the if. Will change that!

goapunk (approved):
cool :)

PR #4233, opened 2022-03-03
apps/plans & apps/plans/templates: default text for participation explanation

### LINKED ISSUE #4222: remove default text from plan - participation explanation?
**URL:** 
**user:** 
**expected behaviour:** I get an empty field for participation explanation when I add a plan.
**behaviour:** I have to remove the default text when I add a new plan. This invites to leaving the default text, which will then not be shown. Something we also did not want to encourage.
**important screensize:**
**device & browser:** 
**Comment/Question:** We can fix that by just removing the default from the field again. Then it will behave like the other required fields: initially empty and complaining if nothing is filled in. This is for @CarolingerSeilchenspringer to decide!

Screenshot?

### LINKED ISSUE #4225: plan detail: when participation explanation has default text, the whole box shouldn't be shown
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** Only the explanation is not shwon when it has the default text
**important screensize:**
**device & browser:** 
**Comment/Question:** @CarolingerSeilchenspringer said yesterday, that the whole box should be gone when the explanation has the default.

Screenshot?

sabinammm:
Commit 1:
apps/plans: remove default text from plan-participation-explanation
Fixes #4222 

Commit 2:
apps/plans/templates: plan detail shows no participation explanation box when explanation is default text in old plans
Fixes #4225


meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -145,8 +145,8 @@ <h3 class="item-detail-2__info-box-headline">{% trans 'Level of participation' %
                         {% else %}
                             {{ object.participation_explanation }}
                         {% endif %}
```
fuzzylogic2000, line 147: This whole block needs to be indented one indent less. I.e. has to be shifted one indent left.

fuzzylogic2000 (changes requested):
Nice! Just one change to keep the template code as readable as possible.

fuzzylogic2000 (approved):
Nice!

PR #4249, opened 2022-03-09
Pm 2022 03 milestone issues

phillimorland:


meinberlin/assets/scss/components/_item_detail_2.scss
```diff
@@ -33,7 +33,7 @@
     border-bottom: 1px solid $border-color;
 }
 
-.item-detail-2__content-borderless {
+.item-detail-2__content--borderless {
```
fuzzylogic2000, line 36: I did not use the variant notation, because I thought for a variant, the main class without the `--variant` also needs to be there.
But I am happy with using it, because it does make more sense here.

fuzzylogic2000 (approved):
NIce! And all done! :upside_down_face:

PR #4264, opened 2022-03-14
[#5629] meinberlin/templates: updated header to latest berlin.de version

sabinammm:
Regarding US 5629 [mB] change header wording


fuzzylogic2000 (approved):
Yeah!

Issue #4251, opened 2022-03-10
testing 5388: design of box, distance box and text


CarolingerSeilchenspringer:
**URL:** https://meinberlin-dev.liqd.net/vorhaben/2021-00082/
**user:** any
**expected behaviour:** 
**behaviour:** design of box is not as in zeplin; distance between text inside box and frame of box is larger than should be. The distance to the contact below the box is smaller than supposed to be.
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
on dev:
<img width="507" alt="Bildschirmfoto 2022-03-10 um 09 21 05" src="https://user-images.githubusercontent.com/35491681/157618620-27965375-2b03-44e4-86fe-21bf267d4da4.png">

on zeplin:
<img width="372" alt="Bildschirmfoto 2022-03-10 um 09 21 54" src="https://user-images.githubusercontent.com/35491681/157618747-47f2ff64-553a-44f3-9ebb-ab4c73e44229.png">

fuzzylogic2000:
Last issue! I will assign you both @khamui and @phillimorland . So, who gets here first. :)

PR #4265, opened 2022-03-14
modul/vorhaben/level-of-participation: styles fix infobox on vorhaben…

khamui:
… detailpage

closing #4251

### DIFF
--- a/meinberlin/assets/scss/components/_item_detail_2.scss
+++ b/meinberlin/assets/scss/components/_item_detail_2.scss
@@ -1,7 +1,7 @@
 .item-detail-2__smalltitle {
     font-weight: bold;
     font-size: inherit;
-    margin-top: 0;
+    margin: 1em 0 0;
 }
 
 .item-detail-2__factsheet {
@@ -52,7 +52,11 @@
 .item-detail-2__info-box {
     border: 1px solid $border-color;
     padding: 1em;
-    margin: 2 * $spacer 0 3.3 * $spacer;
+    margin: 2 * $spacer 0;
+
+    .item-detail-2__smalltitle {
+        margin-top: 0;
+    }
 
     button {
         color: inherit;
--- a/meinberlin/assets/scss/components/_item_detail_2.scss
+++ b/meinberlin/assets/scss/components/_item_detail_2.scss
@@ -1,7 +1,7 @@
 .item-detail-2__smalltitle {
     font-weight: bold;
     font-size: inherit;
-    margin: 1em 0 0;
+    margin-top: 0;
 }
 
 .item-detail-2__factsheet {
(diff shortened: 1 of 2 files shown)


meinberlin/assets/scss/components/_item_detail_2.scss
```diff
@@ -1,7 +1,7 @@
 .item-detail-2__smalltitle {
     font-weight: bold;
     font-size: inherit;
-    margin: 1em 0 0;
+    margin-top: 0;
```
phillimorland, line 4: @khamui is this class used elsewhere? if so have you checked that changing it doesn't break styling?

meinberlin/assets/scss/components/_item_detail_2.scss
```diff
@@ -52,4 +52,10 @@
 .item-detail-2__info-box {
     border: 1px solid $border-color;
     padding: 1em;
+    margin: 2 * $spacer 0 3.3 * $spacer;
```
phillimorland, line 55: why use 3.3 here? it's good to keep our spacing as uniform as possible as makes everything look neater

phillimorland (changes requested):
just couple things

meinberlin/assets/scss/components/_item_detail_2.scss
khamui: yes it is used, but those are fixed (and improved) by this line

https://github.com/liqd/a4-meinberlin/blob/21bfccd629aca0e42209de00535fef11f2742b25/meinberlin/assets/scss/components/_item_detail_2.scss#L55

meinberlin/assets/scss/components/_item_detail_2.scss
khamui: actually from the design it would be 3.375rem. shall i go with 3 or 3.5 then?

phillimorland: 3.5 is good :)

meinberlin/assets/scss/components/_item_detail_2.scss
phillimorland: what about when there isn't the info-box?

khamui: hmm..good point! yes thats not working for this case.

meinberlin/assets/scss/components/_item_detail_2.scss
khamui: actually I changed to 2 because the smalltitle has its own margin-top now (restored from how it was before).

meinberlin/assets/scss/components/_item_detail_2.scss
khamui: I restored the margin-top for smalltitles which are not inside of the box. also good for mobile, if they get wrapped the still have spacing inbetween

phillimorland (approved):
cool!

PR #4297, opened 2022-03-25
config: remove captcha_url

goapunk:


goapunk:
we first need to apply the salt states, otherwise the captcha will break

fuzzylogic2000 (approved):
Cool! You were already on it. :)

PR #4304, opened 2022-03-31
tests: add missing CAPTCHA_URL

goapunk:


fuzzylogic2000 (approved):
Ah! Because it was in the base.py before. :)

PR #4321, opened 2022-04-22
tests/newsletter: check that not choosing project does give an error …

fuzzylogic2000:
…instead of 500

I wanted to properly check that https://github.com/liqd/a4-meinberlin/blob/2ef541440cde3561e77ae64e29fcb11e40f878a4/meinberlin/apps/newsletters/forms.py#L48 gives a FieldError when there is no project added. But when posting, the error is just added to the error list in the content. So, this checks that we stay on the form page instead of getting a redirect like when successfull form submit. Should still be fine to test that it does not yet work in a+ https://github.com/liqd/adhocracy-plus/issues/1737


goapunk (approved):
super cool!

PR #4322, opened 2022-04-25
apps/newsletters: make clean more robust

fuzzylogic2000:
like third commit https://github.com/liqd/adhocracy-plus/pull/1800/commits


goapunk (approved):
cool

PR #4363, opened 2022-06-21
[5929] apps/plans: add explanation to export

fuzzylogic2000:
I used mB with Node 16 and the legacy-install.


phillimorland:
@fuzzylogic2000 to get it running I tried:
`nvm install v16`
`npm install --legacy-peer-deps`
`make watch`
was that it or did I miss a step?

fuzzylogic2000:
> @fuzzylogic2000 to get it running I tried: `nvm install v16` `npm install --legacy-peer-deps` `make watch` was that it or did I miss a step?

@phillimorland I deleted everything before doing that. So,
```
npm cache clean --force
make clean
nvm use v16
npm install --legacy-peer-deps
make install
make watch
```
I hope, that works!?

phillimorland:
@fuzzylogic2000 thanks for pointer, still won't run `make install` for me :(

fuzzylogic2000:
Gah, well, maybe we can just merge, so it can be tested on dev? @phillimorland

PR #4388, opened 2022-09-01
update most dependencies to react 18

goapunk:


phillimorland (approved):
main is installing again, will merge and fix the react rendering

fuzzylogic2000 (commented):
Yeah, it does install! But all react components do not work.

fuzzylogic2000:
> main is installing again, will merge and fix the react rendering

Ah, cool!

fuzzylogic2000 (approved):
I also added shpjs now, but with that it works! Installing AND all the components. :)

PR #4413, opened 2022-09-02
makefile: make relase and install with npm legacy-peer-deps

fuzzylogic2000:
main is not building without that. So, maybe we can add it and then remove once we don't need it anymore?


fuzzylogic2000:
well, I will just merge, because I would like to go on with the updates...

PR #4425, opened 2022-09-02
deps/python: remove bleach - as is only used in a4

fuzzylogic2000:
Installing locally works without bleach. We only use it in a4: https://github.com/liqd/adhocracy4/blob/main/adhocracy4/transforms.py But then also in a+ for the ck editor, I guess.

But why do we only have the css-bleach in a4? Don't we need the other bleach for the tags and attributes? https://github.com/liqd/adhocracy4/blob/043494bbd9e66e3da0159fa643315d89112db140/tests/project/settings.py#L177

Hmm, unsure what to do here.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -15,7 +15,6 @@ whitenoise==6.2.0
 zeep==4.1.0
 
 # Inherited a4-core requirements
-bleach==4.1.0
 Django==3.2.15
 django-allauth==0.51.0
 git+https://github.com/fuzzylogic2000/django-autoslug.git@master#egg=django-autoslug


fuzzylogic2000:
Ok, yeah, sure, to make it work with the new bleach with updating a4, we need bleach[css]. Will add it, then. But why only that and not the other?

goapunk:
@fuzzylogic2000 what do you mean by "the other" ? :sweat_smile:

fuzzylogic2000:
> @fuzzylogic2000 what do you mean by "the other" ? sweat_smile

:) I meant "bleach" without "[css]". The tags and attributes are not css, and we want to make sure, they are also cleaned, right?

goapunk:
ah, got it. Yeah, they moved the css cleaning into an optional [css] module, with bleach[css] we intall bleach + the optional css module

fuzzylogic2000:
@goapunk That is confusing naming on their side. Thanks for the explanation!

PR #4448, opened 2022-09-06

phillimorland:
client cpu load. Skip rendering until the tile becomes visible. - currently does not decrease load time much due to back-end serving whole list

Need: 
- [x] refactoring
- [ ] only fetch 'laufend' list on first page load?
- [ ] to work with filters on both map and list view

### DIFF
--- a/meinberlin/apps/plans/assets/PlansList.jsx
+++ b/meinberlin/apps/plans/assets/PlansList.jsx
@@ -10,6 +10,9 @@ const endedParticipationStr = django.gettext('Participation ended. Read result.'
 const statusStr = django.gettext('Status: ')
 const nothingStr = django.gettext('Nothing to show')
 
+const autoScrollThreshold = 500
+const itemListInterval = 10
+
 class LazyBackground extends React.Component {
   constructor (props) {
     super(props)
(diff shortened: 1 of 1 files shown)


phillimorland:
retagging in new PR as old one closed due to revert causing rebase to close it
@fuzzylogic2000 @Rineee @goapunk been thinking about this and the actual most important load is the initial one where we have a default filter for 'laufend', however we fetch the whole list can anyone remember why we died that instead of requesting only the laufend list from backend? I feel there probably was a good reason we do the filtering in the front but I don't know it?!

goapunk:
I guess ideally we'd have some kind of pagination ? e.g. fetch first 10 from every type (or the max we can display without scrolling) because if we just leave out the past ones we don't really gain much:

* active: 422ms
* future: 112ms
* past: 2061ms
* plans: 2560ms
* extprojects: 3802ms
* containers: 420ms
* privateprojects: 438ms

They are all fetched in parallel (I think), so we always have to wait for the slowest - unless we improve that one leaving out others won't help us.

phillimorland:
closing as this never worked, conversation continuing https://github.com/liqd/a4-meinberlin/issues/4832

PR #4449, opened 2022-09-06
Ks 2022 09 add module types

Rineee:
Also see https://github.com/liqd/adhocracy4/pull/1207, which should be merged first. This PR is not really dependent on it, but without the A4 changes, new external projects/project containers/bplans wont get a type when saving.

### DIFF
--- a/meinberlin/config/settings/base.py
+++ b/meinberlin/config/settings/base.py
@@ -536,6 +536,25 @@ A4_PROJECT_TOPICS = (
     ('TRA', _('Traffic'))
 )
 
+A4_BLUEPRINT_TYPES = [
+    ('BS', 'brainstorming'),
+    ('MBS', 'map brainstorming'),
+    ('IC', 'idea collection'),
+    ('MIC', 'map idea collection'),
+    ('PB', 'participatory budgeting (1 phase)'),
+    ('PB2', 'participatory budgeting (2 phase)'),
+    ('PB3', 'participatory budgeting (3 phase)'),
+    ('KK', 'kiezkasse'),
+    ('TP', 'topic prioritization'),
+    ('MTP', 'map topic prioritization'),
+    ('TR', 'text review'),
+    ('PO', 'poll'),
+    ('IE', 'interactive event'),
+    ('EP', 'external project'),
+    ('BP', 'bebauungsplan'),
+    ('PC', 'project container'),
+]
+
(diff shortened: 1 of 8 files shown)


meinberlin/config/settings/base.py
```diff
@@ -536,6 +536,25 @@
...
+    ('PB3', 'participatory budgeting (3 phase)'),
+    ('KK', 'kiezkasse'),
+    ('TP', 'topic prioritization'),
+    ('MTP', 'map topic prioritization'),
+    ('TR', 'text review'),
+    ('PO', 'poll'),
+    ('IE', 'interactive event'),
+    ('EP', 'external project'),
+    ('BP', 'bebauungsplan'),
+    ('PC', 'project container'),
```
Rineee, line 555: Wasnt sure if we want these as blueprint types, but as they appear in the list of modules and otherwise dont have a type, I added them

fuzzylogic2000: Yes, I think, we also want these. Even though, we can also distinguish these projects from the a4 projects via the project type.
About adding the types to the settings: I like that all the types are here in one place and also have a human-readable name here. (Maybe we can make use of the names in the django-admin?) I don't like that they are in the code twice then and that the types from the settings are only used in a4, not here.

fuzzylogic2000: Right! But we also don't use the types from the settings in a4, but only check if they are there. So, can we instead add a setting `USE_A4_BLUEPRINT_TYPES = True`?

fuzzylogic2000 (commented):
The migration works perfectly! And for the settings, maybe we can change them a bit?

meinberlin/config/settings/base.py
Rineee: Ah, I am not really following :see_no_evil: 
Do you mean use the names of the blueprints instead of adding names here, is that what you meant? 
And then remove setting A4_BLUEPRINT_TYPES and add another setting USE_A4_BLUEPRINT_TYPES and use that in A4?

Rineee: Alright, the A4_BLUEPRINT_TYPES are needed in A4 to initialize the choices for field module.blueprint_type, see here https://github.com/liqd/adhocracy4/pull/1136/ . So we decided to leave it like this for now..

fuzzylogic2000 (approved):
Nice!

PR #4451, opened 2022-09-07
[#6422] templates/pagination.html: update pagination template

phillimorland:


fuzzylogic2000:
There were two commit redicing the page size to 4 and then 2. I out that into one commit and will make it merge as soon as the CI is run and fine.

phillimorland:
@fuzzylogic2000 thanks, sorry missed that!

PR #4452, opened 2022-09-07
WIP: Kl 2022 09 token generation bulk create

fuzzylogic2000:
Two commits trying different ways of token creation. The bulk-version is much faster, but we need to think about uniqueness and about disabling the button and/or rules (should all initiators be allowed to do it???).


fuzzylogic2000:
I will make a new branch with the changes we talked about.

PR #4455, opened 2022-09-08
assets/*: additional design fixes due to moving to bs

phillimorland:


meinberlin/assets/scss/components/_button.scss
```diff
@@ -21,10 +21,6 @@
 
     &.is-disabled,
     &:disabled {
-        // $is-light: luma($color) > luma($color-contrast);
-        // $weight-bg: if($is-light, 95%, 50%);
-        // $weight-fg: if($is-light, 60%, 10%);
-
         // border-color: mix($color, $color-contrast, $weight-bg);
         // background-color: mix($color, $color-contrast, $weight-bg);
         // color: mix($color, $color-contrast, $weight-fg);
```
goapunk, line 26: is this still relevant or can it also be removed?

phillimorland: i'm still debating that, I wanted to leave it in till it's been tested a bit in case Caro notices anything amiss and I can update the styles using non planifolia styles

meinberlin/assets/scss/_base.scss
```diff
@@ -127,8 +127,9 @@ dt {
     margin-top: $em-spacer;
 }
 
-button,
-[type="submit"] {
+// FIXME what was this for? breaks button styling?!
+// [type="submit"]
+button {
```
goapunk, line 132: this will now apply to all buttons (if there are others?) instead of just the ones which do a submit, do we want that?

fuzzylogic2000 (approved):
Nice!

meinberlin/assets/scss/components/_poll.scss
```diff
@@ -187,8 +187,10 @@ $checkbox-size: 20px;
 }
 
 .editpoll__question-container {
-    display: block;
+    // display: block;
```
goapunk, line 190: no dead code ? :)

fuzzylogic2000:
Sorry, that I already merged!

PR #4458, opened 2022-09-08
replace node-sass with sass

goapunk:


goapunk:
should also help with ci speed I think

phillimorland (approved):
nice!

PR #4464, opened 2022-09-09
Ks 2022 09 module type fixes

Rineee:
I could not find anything else to fix, also looked into templates.. we will probably find more later!

### DIFF
--- a/meinberlin/apps/budgeting/dashboard.py
+++ b/meinberlin/apps/budgeting/dashboard.py
@@ -14,8 +14,7 @@ class ExportBudgetingComponent(DashboardComponent):
     label = _('Export Excel')
 
     def is_effective(self, module):
-        module_app = module.phases[0].content().app
-        return (module_app == 'meinberlin_budgeting' and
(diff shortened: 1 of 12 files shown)


Rineee:
Oh no, will fix test on Monday.. although it does not fail locally, which is weird

fuzzylogic2000 (approved):
Looks great!
Should we merge now or after adding more tests? :)

Rineee:
> Looks great! Should we merge now or after adding more tests? :)

Hm, I did not really add anything, why did the coverage go down? Its mostly the get_progress in the dashboard components, that is now not covered anymore, can investigate why that is and if I can add tests..

fuzzylogic2000:
> > Looks great! Should we merge now or after adding more tests? :)
> 
> Hm, I did not really add anything, why did the coverage go down? Its mostly the get_progress in the dashboard components, that is now not covered anymore, can investigate why that is and if I can add tests..

Let's just go craaazy and merge without the tests! :rofl:

PR #4468, opened 2022-09-12
[5044] apps/votes: add token generation component to dashboard, use async tasks

fuzzylogic2000:
This is the generation in an extra dashboard component. It puts the bulk_create in tasks split up in chunks of 100,000. So, to test remember to start the background tasks.

### DIFF
--- a/meinberlin/apps/votes/templates/meinberlin_votes/voting_code_dashboard.html
+++ b/meinberlin/apps/votes/templates/meinberlin_votes/voting_code_dashboard.html
@@ -1,0 +1,40 @@
+{% extends "a4dashboard/base_dashboard_project.html" %}
+{% load i18n %}
+
+{% block dashboard_project_content %}
+<h1 class="u-first-heading mt-0">{% translate 'Generate voting codes' %}</h1>
+
+<p class="form-hint">{% blocktranslate %}For voting in the third phase, participants need a voting code. You can generate and download the codes here. {% endblocktranslate %}</p>
+<p>{% blocktranslate count counter=number_of_module_tokens %}
+One code has been generated for this module so far.
+{% plural %}
+{{ counter }} codes have been generated for this module so far.
+{% endblocktranslate %}</p>
+<table class="export-table">
+    <thead>
+    <tr>
+        <th></th>
+        <th></th>
+    </tr>
+    </thead>
+    <tbody>
+        <tr>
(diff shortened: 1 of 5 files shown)


meinberlin/apps/votes/templates/meinberlin_votes/voting_code_dashboard.html
```diff
@@ -0,0 +1,40 @@
+{% extends "a4dashboard/base_dashboard_project.html" %}
+{% load i18n %}
+
+{% block dashboard_project_content %}
+<h1 class="u-first-heading mt-0">{% translate 'Generate voting codes' %}</h1>
+
+<p class="form-hint">{% blocktranslate %}For voting in the third phase, participants need a voting code. You can generate and download the codes here. {% endblocktranslate %}</p>
+<p>{% blocktranslate count counter=number_of_module_tokens %}
+One code has been generated for this module so far.
```
khamui, line 9: out of curiosity. with adding the voting module it will generate one code per se?

khamui (approved):
Pretty decent! working fine.
Ofc leaving for @Rineee to check and merge.

meinberlin/apps/votes/templates/meinberlin_votes/voting_code_dashboard.html
fuzzylogic2000: No, you are right, we should use the number here as well as in other languages the 0 might have an extra form. In English and German the plural is used, so it should be fine. I will check to be sure, though!

fuzzylogic2000: Yes, it does work with the 0 in English and it will also be fine in German with the translation

Rineee (commented):
Yeah cool, it is sooooo much faster now!

I checked what happens when the unique constraint is violated, then the whole background task fails and no code is generated... but should we merge and handle that later then @fuzzylogic2000 ?

fuzzylogic2000:
> Yeah cool, it is sooooo much faster now!
> 
> I checked what happens when the unique constraint is violated, then the whole background task fails and no code is generated... but should we merge and handle that later then @fuzzylogic2000 ?

@Rineee Yes, let's merge and I will add a task to handle that. Would be good to do with tests anyway. :)

PR #4469, opened 2022-09-12
[6421] apps/notifications: send email on moderator feedback to proposal cont…

fuzzylogic2000:
…act if not same as creator email

- [x] Tests are still missing....


Rineee (approved):
Nice!

PR #4471, opened 2022-09-13
tests: test dashboard components

Rineee:
Could not find a more elegant way to test this :woman_shrugging:


fuzzylogic2000 (approved):
Nice! I like that it tests that the components are shown when we intend them to be shown. And also that you added the progress and the other bits!

PR #4473, opened 2022-09-14
Ks 2022 09 proposal form display

Rineee:


khamui (approved):
Looking good to me. Not sure about the translation process (weblate?), so asking @fuzzylogic2000 to doublecheck.

Rineee:
> Looking good to me. Not sure about the translation process (weblate?), so asking @fuzzylogic2000 to doublecheck.

Yes, probably we use weblate? At least I found meinBerlin on it there.. so remove the last commit @fuzzylogic2000 ?

fuzzylogic2000:
> > Looking good to me. Not sure about the translation process (weblate?), so asking @fuzzylogic2000 to doublecheck.
> 
> Yes, probably we use weblate? At least I found meinBerlin on it there.. so remove the last commit @fuzzylogic2000 ?

I also don't know exactly how we will use it. But we can just handle that later and use these translations. And then we can set up the proper workflow next week. (The strings that are already there can just be pushed to weblate, so that should be fine!)

fuzzylogic2000 (approved):
Perfect!

PR #4475, opened 2022-09-14
Ks 2022 09 proposals random sorting

Rineee:
The random stuff is copied from [here](https://github.com/liqd/adhocracy4/blob/21dd3f00af6ce178688b120ad7872a669fc254c8/adhocracy4/filters/filters.py#L113). Im a bit unhappy with doubling everything, but did not know how else to do it in a good way? 

ToDo:

- [x] add and fix tests


meinberlin/apps/contrib/filters.py
```diff
@@ -11,3 +17,26 @@ def filter_queryset(self, request, queryset, view):
...
+
+class OrderingFilterWithDailyRandom(OrderingFilter):
+
+    def filter_queryset(self, request, queryset, view):
+        ordering = self.get_ordering(request, queryset, view)
+
+        if ordering:
+            if ordering == ['-daily_random']:
+                pks = list(queryset.values_list('pk', flat=True))
+                random.seed(str(date.today()))
```
khamui, line 30: nice!

khamui (approved):
working fine!

the `preserved` variable is hard to read for me, but its my inexperience :) Leaving for @fuzzylogic2000 to doublecheck!

fuzzylogic2000 (approved):
Nice! Also, more testing. :)

fuzzylogic2000 (commented):
It looks good, but it seems like the newest filter breaks with that? (Fine on dev, but different to the second list locally.) Will investigate more later!

meinberlin/apps/budgeting/api.py
```diff
@@ -85,12 +85,13 @@ def list(self, request, *args, **kwargs):
         ordering_choices = [('-created', _('Most recent')), ]
         if self.module.has_feature('rate', Proposal):
             ordering_choices += ('-positive_rating_count', _('Most popular')),
-        ordering_choices += ('-comment_count', _('Most commented')),
+        ordering_choices += ('-comment_count', _('Most commented')), \
+                            ('-daily_random', _('Random')),
```
fuzzylogic2000, line 89: Why the -? It probably doesn't matter if we use the random ascending or descending.

fuzzylogic2000 (approved):
It is all fine looking at it again. And also it doesn't make sense that the code shouldn't work like before.

PR #4480, opened 2022-09-15
WIP budgeting-list/filterbar/search: adding freetext search components

khamui:


khamui:
- [ ] check a11y
- [ ] styling, polishing (mobile)
- [ ] make work with api

meinberlin/assets/scss/components/_filter_bar.scss
```diff
@@ -235,11 +235,13 @@ $divider-margin: 1.25rem;
 
 .filter-bar__search {
     display: inline;
-    width: calc(100% - 63px);
+    width: calc(100% - 37.1875px);
```
khamui, line 238: the `magnifyier icon` seems to changed in width (for the projects search it looks better now as well)

meinberlin/assets/scss/components/_filter_bar.scss
```diff
@@ -235,11 +235,13 @@ $divider-margin: 1.25rem;
...
     margin-bottom: 2 * $spacer;
+    border-top-left-radius: 0.3em !important;
```
khamui, line 240: will look into this one again, to avoid `!important`. This gets overwritten by another class, will check if can be removed there.

meinberlin/assets/scss/components/_control-bar.scss
```diff
@@ -22,15 +22,9 @@
             // @include grid-same-width(2);
         }
 
-        > :last-child {
```
khamui, line 25: this only had effect on the react filterbar. so removing is fine i think.

khamui:
Note: the last commit touches unrelated files, but were mentioned in the lighthouse scan (in case how i solved this is not ideal, i can remove the commit from this PR).

meinberlin/assets/scss/components/_filter_bar.scss
khamui: okay, doublechecked it and saw that another class is overwriting `border-radius: 0`. This class is widely used so leaving the `!important` for now, if that is fine @phillimorland ?

phillimorland: I'm thinking maybe we should re-write our search inputs to use bootstrap mark up 
```
<div class="input-group mb-3">
  <input type="text" class="form-control" placeholder="search" aria-label="search" aria-describedby="id-search">
  <div class="input-group-append">
    <button class="btn btn-outline-secondary" type="button" id="id-search">Button</button>
  </div>
</div>
```
what do you think?

meinberlin/apps/votes/forms.py
```diff
@@ -9,11 +9,14 @@ class VotingTokenWidget(forms.MultiWidget):
     def __init__(self, attrs=None):
         widgets = (
             forms.TextInput(attrs={'minlength': '4',
-                                   'maxlength': '4'}),
+                                   'maxlength': '4',
```
phillimorland, line 12: this can be removed as it is being refactors as I looked into it and having multiple inputs is bad a11y so we will make a single input

phillimorland (commented):
sorry changed name to include the story number, would be good if we can all start doing that more! 
also my local environment is broken right now so couldn't actually install and look at it but added notes about the code, let me know if you want to talk through the possibility of refactoring

meinberlin/assets/scss/components/_filter_bar.scss
khamui: yes that makes sense, i will have a look and give it a try @phillimorland

meinberlin/apps/budgeting/assets/FilterBarListMapSwitch.jsx
```diff
@@ -6,13 +6,13 @@ export const FilterBarListMapSwitch = ({ query }) => {
   return (
     <div className="btn-group__container">
       <div className="btn-group" role="group">
-        <div
+        <button
```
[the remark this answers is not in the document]
khamui: good point, didn't know about `aria-pressed` tbh. Actually I realized this component is a little bit of a fake, because this button never changes its state. E.g. the buttons `View as List` and `View as Maps` will remain always the same in this component.

Why is that? Because we use the Map in a Django template, if we switch to toggle the `Maps` button, then it will show this button as a django template.

khamui:
to mention: the first commit is all commits before @phillimorland 's review, squashed together.
So now we are using Bootstrap5 classes for the search-filter inputs (most of them) and to make them responsive and layouted correctly i tried to utilize classes that we already have in use, instead of adding more. But in the end i had to add some overwrites to make it look how it is in the design.

meinberlin/apps/budgeting/assets/FilterBar.jsx
```diff
@@ -71,24 +71,22 @@ export const FilterBar = props => {
         <FilterBarListMapSwitch query={filterString} />
       </div>
       <div className="offset-lg-2 col-lg-8">
-        <div className="filter-bar__section">
-          <div className="filter-bar__section-item">
+        <div className="control-bar control-bar-search">
+          <div className="searchinput">
```
khamui, line 75: these are equivalent `sub-wrappers` to `dropdown`, for the search input and the filters toggle button.

meinberlin/assets/scss/components/_control-bar.scss
```diff
@@ -1,8 +1,18 @@
 .control-bar {
     @include clearfix;
 
+    &-search {
```
khamui, line 4: i could not find an easy way to make the search input grow without using flexbox :(

`[searchbox ----->  grow                 ] [dropdown] [togglebutton]`

meinberlin/apps/budgeting/assets/FilterBar.jsx
phillimorland: > these are equivalent `sub-wrappers` to `dropdown`, for the search input and the filters toggle button. So nested in `control-bar`, we want to have either `dropdown`, `searchinput` or `togglebutton`.
> 
> maybe instead of adding two new specific, it is better to rename `dropdown` to something more general? `control-bar-item`?

not sure I follow, you want to rename `dropdown` class? or just the wrapper classes? I would be in favour of having a general wrapper class but we should do it everywhere so any instance of a control bar should be updated so we are only maintaining 1 style and delete the old code.

phillimorland (changes requested):
few comments about BEM also we discussed ensuring filters are the same everywhere and simplifying the overwrites, however that isn't apparent here, did you decide against it? i'm not sure if this may just increase the mess that's already there.

meinberlin/apps/budgeting/assets/FilterBar.jsx
```diff
@@ -71,24 +71,22 @@ export const FilterBar = props => {
...
-            </div>
+            <FilterBarDropdown
+              key="ordering_dropdown"
+              filter={props.filters.ordering}
+              filterId="id_ordering"
+              onSelectFilter={choice => applyFilter('ordering', choice)}
+            />
           )}
-          <div className="filter-bar__section-item">
+          <div className="togglebutton">
```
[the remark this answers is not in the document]
khamui: sorry @phillimorland, but not sure if i get it all right.

yes i wasn't sure about these class names, i thought i use something that fits to the existing one `dropdown`.
I know `dropdown` is kind of a keyword in BS5 and is used as a `modifier`. So bs5 is not BEM right?

the mess (at least in my head) comes with mixing bootstrap 5 and BEM. So bs5 uses `dropdown` to identify functionality/modifier. But here it gets overwritten and styled (margin-bottom).

Additionally we seem to not following BEM regarding spacing. Because buttons, dropdowns and general elements have spacing applied. As i understand BEM, spacing should be the responsibility of the container?

>Also to match out existing convention we should either use filter-bar as out block element and extend to control bar in the component style file or change the name of the component.

Not sure if I follow this correctly. So i am creating a new class `filter-bar` and use `control-bar` as base and would have something like `filter-bar__search` and `filter-bar__btn-toggle`?

khamui:
again i squashed all previous commits together into the first commit!

phillimorland (approved):
super cool! nice and clean! :) @khamui do you want to squash all commits? then it's good to merge

phillimorland (approved):

PR #4482, opened 2022-09-19
[6421] apps/budgeting: only sent moderated email to contact saved in form, n…

fuzzylogic2000:
…ot to creator

@CarolingerSeilchenspringer and me decided to always sent the notification to the contact from the proposal, because we don't need to worry that the creator has disabled notifications then.


goapunk (approved):
looks good to me!

PR #4483, opened 2022-09-19
[#6425 A11Y] lighthouse fixes

phillimorland:


meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
```diff
@@ -27,6 +27,7 @@
                     <div class="form-group">
                         {% csrf_token %}
                         <div class="widget--{{ token_form.token|widget_type }}">
+                            <label class="visually-hidden-focusable">{% trans 'Voting token form fields' %}</label>
```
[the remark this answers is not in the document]
phillimorland: after some research I think we might need to refactor this and do it like in the design with 1 input but number sets separated by a space, i will look into this now

meinberlin/apps/votes/forms.py
```diff
@@ -1,34 +1,62 @@
...
+            raise forms.ValidationError(self.error_messages['invalid'])
+
+        return value
+
+    def token_from_number(self, num):
+        # find this token, based on the token number
+        for token in self.token:
+            for pattern in token['patterns']:
+                if (str(pattern) == str(num)[:len(str(pattern))]):
+                    return token
```
phillimorland, line 54: @fuzzylogic2000 here it is if you want to take a look, I will keep poking till lunch then maybe go on to react update test branch

fuzzylogic2000: Right, I can also fix the (backend) test! Maybe when we are done with everything here?

meinberlin/apps/votes/assets/token_input.js
```diff
@@ -0,0 +1,109 @@
+function init () {
```
phillimorland, line 1: @goapunk @khamui so I got this from a stackoverflow suggestion on how to autoformat a credit card entry, I have kind of cleaned it but not enough but before i spend more time I wanted to sanity check with you guys?

PR #4484, opened 2022-09-19
Kl 2022 09 dashboard token perms

fuzzylogic2000:
I am unhappy with how I did it and with the naming, but I couldn't think of a better (and not also overly complex) way. It should work better with the permissions from the view somehow, but as we don't have the user in the dashboard components, and I didn't want to add that complexity, I did it like that. :woman_shrugging: And I thought, if we ever need more and other perms in the dashboard, we can still improve from here.

Depending on https://github.com/liqd/adhocracy4/pull/1216

Right, and the two first commits are from https://github.com/liqd/a4-meinberlin/pull/4477

So, these both need to be merged first and this cleaned up then.

### DIFF
--- a/meinberlin/apps/documents/dashboard.py
+++ b/meinberlin/apps/documents/dashboard.py
@@ -58,7 +58,7 @@ class ExportDocumentComponent(DashboardComponent):
     def get_urls(self):
         return [
             (r'^modules/(?P<module_slug>[-\w_]+)/export/document/$',
-             views.DocumentDashboardExportView.as_view(),
+             views.DocumentDashboardExportView.as_view(component=self),
              'document-export-module'),
             (r'^modules/(?P<module_slug>[-\w_]+)/export/document/comments/$',
              exports.DocumentExportView.as_view(),
--- a/meinberlin/apps/votes/dashboard.py
+++ b/meinberlin/apps/votes/dashboard.py
@@ -26,7 +26,7 @@ class VotesComponent(DashboardComponent):
     def get_urls(self):
         return [
             (r'^modules/(?P<module_slug>[-\w_]+)/voting/$',
-             views.VotingDashboardView.as_view(),
+             views.VotingDashboardView.as_view(component=self),
              'voting-tokens'),
             (r'^modules/(?P<module_slug>[-\w_]+)/voting/export-token/$',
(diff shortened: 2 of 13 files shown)


meinberlin/apps/votes/dashboard.py
```diff
@@ -38,6 +38,7 @@ class GenerateVotesComponent(DashboardComponent):
     identifier = 'voting_token_generation'
     weight = 49
     label = _('Generate voting codes')
+    for_superuser = True
```
khamui, line 41: think the wording is fine and easy to understand, imho it would be a tiny little bit more accurate:
`for_superuser_only`, because `for_superuser = False` sounds like all other users are can, except of superusers.

khamui (commented):
Cool!

meinberlin/templates/a4dashboard/includes/nav_modules_item.html
```diff
@@ -23,6 +23,7 @@
         <div class="dashboard-nav__menu-content">
             <ul>
             {% for item in module_menu.menu %}
+                {% if not item.for_superuser or request.user.is_superuser %}
```
fuzzylogic2000, line 26: @Rineee Concerning https://github.com/liqd/adhocracy4/pull/1216#pullrequestreview-1113438639
I check in the template. :grimacing:

Rineee: Haha, yes, I just found this PR, after my review there :)

meinberlin/apps/documents/dashboard.py
```diff
@@ -58,7 +58,7 @@ def get_base_url(self, module):
     def get_urls(self):
         return [
             (r'^modules/(?P<module_slug>[-\w_]+)/export/document/$',
-             views.DocumentDashboardExportView.as_view(),
+             views.DocumentDashboardExportView.as_view(component=self),
```
Rineee, line 61: Good catch! I think also for the budgeting export it is also still missing..

Rineee (approved):
Really cool, I think this is a good solution! 
So fine for me to merge after naming update and adding component to budgeting export view!

Rineee (approved):
Yeah!

PR #4489, opened 2022-09-20
[#6460]apps/*/assets/*: update render function syntax| deps: update a4 hash t…

phillimorland:
…o working PR
Dependent on https://github.com/liqd/adhocracy4/pull/1217

Tested:

- [x] Poll dashboard
       ! button styling broken
       ! popper menu styling broken and lib unsupported now
       ! moved questions don't save (known issue)
       NTH: re-structure file to separate dashboard and user components
       NTH: re-implement the loading animation for resultes
- [x] Poll User
- [x] Comments (not with functional components)
       ! styling of copy modal wrong
       ! buttons wrong styling
       ! cannot save answer
       ! linking doesn't work-no scrolling or styling
       ! constant loading icon
- [x] Rating
- [x] PlansMap
- [x] PlansMap Filter 
       ! filters are not selected immediately every time, behaviour is erratic 
- [x] Documents dashboard
- [x] Documents user
- [x] Interactive Event dashboard
- [x] Interactive Event user
      ! button styling wrong
      ! intro page when event not running seems to be skipped but that isn't react?
      NTH: implement loading animation if we do it in polls
- [x] Following
- [x] Alert
- [x] Budgeting List 
- [x] Vote (Budgeting)
      ! button styling broken

### DIFF
--- a/meinberlin/apps/plans/assets/PlansMap.jsx
+++ b/meinberlin/apps/plans/assets/PlansMap.jsx
@@ -85,6 +85,12 @@ class PlansMap extends Component {
     this.markers = this.addMarkers(this.cluster)
   }
 
+  componentWillUnmount () {
+    if (this.map !== undefined) {
+      this.map.remove()
+    }
+  }
+
   bindMap (element) {
     this.mapElement = element
   }
--- a/meinberlin/apps/plans/assets/react_plans_map.jsx
+++ b/meinberlin/apps/plans/assets/react_plans_map.jsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { createRoot } from 'react-dom/client'
 import { CookiesProvider } from 'react-cookie'
-import ListMapBox from './ListMapBox'
+import PlansListMapBox from './PlansListMapBox'
 
 function init () {
   const plansMapBox = document.querySelectorAll('[data-map="plans"]')
@@ -27,7 +27,7 @@ function init () {
     const root = createRoot(el)
     root.render(
       <CookiesProvider>
-        <ListMapBox
+        <PlansListMapBox
           selectedDistrict={selectedDistrict}
           selectedTopic={selectedTopic}
           projectApiUrl={projectApiUrl}
--- a/meinberlin/apps/plans/assets/react_plans_map.jsx
+++ b/meinberlin/apps/plans/assets/react_plans_map.jsx
@@ -26,28 +26,30 @@ function init () {
     const participationChoices = JSON.parse(el.getAttribute('data-participation-choices'))
     const root = createRoot(el)
     root.render(
-      <CookiesProvider>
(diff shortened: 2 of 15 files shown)


phillimorland:
I have added the list to the story so we can work on it from there

meinberlin/apps/plans/assets/react_plans_map.jsx
```diff
@@ -0,0 +1,55 @@
...
+    const districts = JSON.parse(el.getAttribute('data-districts'))
+    const organisations = JSON.parse(el.getAttribute('data-organisations'))
+    const districtnames = JSON.parse(el.getAttribute('data-district-names'))
+    const topicChoices = JSON.parse(el.getAttribute('data-topic-choices'))
+    const mapboxToken = el.getAttribute('data-mapbox-token')
+    const omtToken = el.getAttribute('data-omt-token')
+    const useVectorMap = el.getAttribute('data-use_vector_map')
+    const participationChoices = JSON.parse(el.getAttribute('data-participation-choices'))
+    const root = createRoot(el)
+    root.render(
```
goapunk, line 28: You are wrapping some of the returns in `PlansListMapBox` in strict mode instead of just wrapping the whole component in strict mode here, is that because not everything works with strict mode here?

phillimorland: the map cannot be wrapped, I mentioned it in the commit message but didn't add the error, basically complains that the map is already initialised

goapunk: I start to suspect that this is our first bug found/triggered by strict mode, maybe we can discuss it next week

phillimorland: I also suspected that although i did find similar issues to do with how leaflet initializes the map but yes we can talk about it.

meinberlin/apps/plans/assets/PlansMap.jsx
```diff
@@ -68,7 +68,9 @@ class PlansMap extends Component {
   }
 
   componentDidMount () {
-    this.map = this.setupMap()
+    if (this.map === undefined) {
```
goapunk, line 71: works for me now! One thing I'm unsure about is the remaining code in `componentDidMount`, should it also be in the if statement? or instead of skipping `setupMap`, should we rather do `if(this.map !== undefined) {this.map.remove()}` in `componentWillUnmount`? I don't know what's best here

phillimorland: I think having it in the unmount is cleaner and i think it has sorted out the filter behaviour but I will test and record behaviour for some filter selections on dev now and check them again after merge

goapunk (approved):
cool, looks good to me!

PR #4490, opened 2022-09-20
apps/votes: add more info to task

fuzzylogic2000:
We did add the management command to notify us of failed tasks. https://github.com/liqd/a4-meinberlin/pull/1408 And we did add the MAX_ATTEMPTS in the server settings. I think we need to keep the MAX_ATTEMPTS at 1 because we don't want to have emails fail multiple times.
Unluckily there is no way of having different MAX_ATTEMPTS for different tasks (https://github.com/arteria/django-background-tasks/issues/256) and the package seems unmaintained by now (https://github.com/arteria/django-background-tasks/issues/267). So, I would say we leave it like that for now.

### DIFF
--- a/tests/votes/dashboard_components/test_token_generation.py
+++ b/tests/votes/dashboard_components/test_token_generation.py
@@ -3,12 +3,16 @@ from unittest.mock import patch
 import pytest
 from background_task.models import Task
 from django.contrib.messages import get_messages
+from django.db import transaction
+from django.db.utils import IntegrityError
 from django.utils import translation
 
 from adhocracy4.dashboard import components
 from adhocracy4.test.helpers import redirect_target
 from adhocracy4.test.helpers import setup_phase
 from meinberlin.apps.budgeting.phases import VotingPhase
+from meinberlin.apps.votes.models import VotingToken
+from meinberlin.apps.votes.tasks import generate_voting_tokens_batch
 
 component = components.modules.get('voting_token_generation')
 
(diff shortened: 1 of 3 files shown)


tests/votes/dashboard_components/test_token_generation.py
```diff
@@ -78,3 +80,14 @@ def test_token_generate_view_max_validation(
         'Only 5 tokens are allowed per module. You are allowed to '
         'generate 3 more.')
     assert Task.objects.all().count() == 0
+
+
+@patch('meinberlin.apps.votes.models.get_token',
```
fuzzylogic2000, line 85: @Rineee @goapunk I wanted to get `django.db.utils.IntegrityError: UNIQUE constraint failed` by pathcing the default callable from the token field. Buuuut that just doesn't work and I cannot get it working and I will probably try to think of another solution (not including tests, but handling the error) tomorrow.
But if you have an idea or want to try or get inspired, it would be great if you would have a go at it!

goapunk: @Rineee @fuzzylogic2000 It took me really long to have the idea to mock the secrets module instead of get_token but with this it works

fuzzylogic2000: @goapunk Supercool! Thank you!

Rineee (approved):
Yes, cool, I think this is the best we can do for now!

PR #4493, opened 2022-09-21
[#6486] budgeting-list/tiles: adding reference number to tile

khamui:


khamui:
just saw gh actions failing, need to fix tests

meinberlin/apps/budgeting/serializers.py
```diff
@@ -122,3 +123,6 @@ def get_vote_allowed(self, proposal):
             return has_voting_permission and is_three_phase_budgeting
 
         return False
+
+    def reference_number(self):
+        return '{:d}-{:05d}'.format(self.created.year, self.pk)
```
fuzzylogic2000, line 128: Cool! But now that we annotate it in the api (@Rineee s PR), I think you can just use that. 
Hmm, but maybe this is more robust as we don't need to add fallback when it's not there like for the comment_count and so on.

khamui: i will leave it for now, that might be not correct for future non-zero leading numbers probably?
Just fixed the tests for now

goapunk: I'm generally fine with both, just one thing to keep in mind: having two different sources for the reference number might be confusing? so using the annotation might avoid some confusion later...and it means just changing the function to
```
def reference_number(self):                                                 
          if hasattr(proposal, 'ref_number':                                      
              return proposal.ref_number                                          
          else:                                                                   
              return ''
```

goapunk (approved):
LGTM, just a comment regarding the annotation

meinberlin/apps/budgeting/serializers.py
fuzzylogic2000: yes, I was with "not that important now", but @goapunk is right, one change now, but will be less confusion later and we will notice when it's not annotated. :)

fuzzylogic2000 (approved):
Nice!

PR #4495, opened 2022-09-21
add isort config

goapunk:


goapunk:
the isort part shouldn't be in the editorconfig as it's not an editor setting but an isort setting. It makes my editor very sad

fuzzylogic2000 (approved):
LGTM

phillimorland (approved):
we don't want a sad editor :p LGTM

PR #4496, opened 2022-09-21
Ks 2022 09 budgeting search filter

Rineee:
I am not very happy with all of the cases in the annotation, but I think there is no way to use a python function in an annotation, so this is the solution I came up with.. happy for any improvements!

@khamui I did not (yet) add the search filter to the other filters in the api as there is no real information, right? 

The search works with url/?search=my+search+term, so spaces work with '+'. You can just look at the filtered list in the browser, eg. http://localhost:8003/api/modules/$moduleId/proposals/?search=something+here should work.


fuzzylogic2000 (approved):
Nice! Even though the annotation is quite complicated, I like it much more this way than overwriting the filter! Cool!

PR #4505, opened 2022-09-22
budgeting/test_propsal_api: test search filter

Rineee:
Test is being skipped at the moment, so dont know if it makes sense to merge it now or later when pagesize has been reset?


fuzzylogic2000 (approved):
Cool! 
And I am unsure about merging. There shouldn't be any conflicts when we undo the page size commits, but it is nicer to merge tests that are actually running. Let's wait for Monday?!

fuzzylogic2000 (approved):
Cool!

Issue #4510, opened 2022-09-25
testing 6422: pagionation in two rows


CarolingerSeilchenspringer:
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-3-phasen-testprojekt/?mode=list
**user:** all
**expected behaviour:** pagination is always in one row
**behaviour:** from page 4 on pagination gets wider and is then in two rows
**important screensize:** smaller than 380 px
**device & browser:** 
**Comment/Question:** does the Berlin style guide have a different pagination for smaller screens?

Screenshot?
![Bildschirmfoto am 2022-09-21 um 17 31 43 (1)](https://user-images.githubusercontent.com/35491681/192148302-82059c9d-3e50-4d05-8bf4-0cc2758c5fcc.png)

PR #4513, opened 2022-09-26
[#6422] pagination issues fixes

khamui:


meinberlin/apps/budgeting/assets/PaginationButton.jsx
```diff
@@ -16,6 +16,8 @@ export const PaginationButton = (props) => {
       return <i className="fa fa-chevron-left" aria-hidden="true" />
     } else if (props.type === 'next') {
       return <i className="fa fa-chevron-right" aria-hidden="true" />
+    } else if (props.type === 'mobile') {
+      return `${props.pageIndex} / ${props.pageCount}`
```
goapunk, line 20: the text color is grey for this and not very readable, I assume we should also use the default black here?

goapunk (changes requested):
nice fixes! just one small thing

meinberlin/apps/budgeting/assets/Pagination.jsx
```diff
@@ -31,17 +38,26 @@ export const Pagination = (props) => {
...
+            />
+          ))
+          : (
+            <PaginationButton
+              type="mobile"
+              key={getKeyStr(0, 0)}
+              pageIndex={currentPage}
+              pageCount={pageCount}
+              onClick={onPaginate}
+              isDisabled
```
goapunk, line 59: we should pass `false` here I think, otherwise the text is grey and hard to read

khamui:
just realize #4510 is just fixed for the react pagination, still the django pagination needs to be done, but not sure if this is quite a bit of rework (at least for me).

would it be fine to do that in another PR so this can be merged already?

goapunk (approved):
nice, looks good to me!

phillimorland (changes requested):
few questions?

meinberlin/assets/scss/_variables.scss
```diff
@@ -55,6 +55,7 @@ $font-size-sm: 0.9rem !default;
 $font-size-xs: 0.8rem !default;
 
 // use for min-width media-queries
+$breakpoint-xxs: 23.75rem !default;
```
phillimorland, line 58: why do you want to add an additional breakpoint? also this is added as a min width value but you use it in a max width query, we don't really want to add additional break points as it add complexity, applying the same style up to a width of 512px is pretty standard, bootstraps first breakpoint is 576 so we aren't far off the norm.

phillimorland (changes requested):
cool, just a few comments and I just noticed when checking it this time that we don't use BEM, maybe we can change pagination-item to pagination__item?

meinberlin/assets/scss/components/_pagination.scss
```diff
@@ -45,10 +48,29 @@
...
+            font-size: $font-size-xs;
+
+            &.disabled {
+                cursor: not-allowed;
+                pointer-events: none;
+                color: $border-color;
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
phillimorland: maybe 0.7 or 0.75 just from having a quick play in the browser but your call

fuzzylogic2000:
What is the status here? @khamui It is finished and can be reviewed again? Will you do that @phillimorland ? Or should I? (I am not sure I will notice everything you talked about.)

phillimorland:
@fuzzylogic2000 sorry didn't see the re-tag, will take a look now

phillimorland (approved):
perfect! :)

PR #4514, opened 2022-09-26
Jd 2022 09 add typescript

goapunk:
* add required typescript packages and configs
* rewrite a very simple component in typescript


meinberlin/apps/documents/assets/__tests__/ChapterForm.jest.jsx
```diff
@@ -7,7 +7,13 @@ import ChapterForm from '../ChapterForm'
 // FIXME: Because of mocking ParagraphForm, some things cannot be tested
 // E.g. not able to reach 100% coverage for this component at the moment.
 // Note: () => () => {} to unwrap it properly.
-jest.mock('../ParagraphForm', () => () => <div>MockedParagraphForm</div>)
+function mockParagraphForm () {
```
goapunk, line 10: fixes a linting error

meinberlin/apps/contrib/assets/CountDown.tsx
```diff
@@ -1,6 +1,13 @@
...
 
-export const CountDown = (props) => {
+interface CountDownProps {
+  activeClass: string
+  inactiveClass: string
+  countText: string
+  counter: number
+}
+
+export const CountDown: React.FC<CountDownProps> = (props) => {
```
khamui, line 10: for me this is already not straightforward to understand tbh. After staring at it a while i get the idea of this typing, but not sure if this will be a pain in the bigger scale?

I would expect something like this:

```
export const CountDown: React.FC = (props: CountDownProps) => {
    ...
}
```

even better if we could omit the component type to have less code pollution, imho it doesn't give
much information to the dev:

```
export const CountDown = (props: CountDownProps) => {
    ...
}
```
Can it be inferred like this?

khamui (commented):
Cool Typescript!!

But already something to discuss

meinberlin/apps/contrib/assets/CountDown.tsx
khamui: what do you think @goapunk and what do the others think @phillimorland @fuzzylogic2000 @Rineee ?

.eslintrc
```diff
@@ -1,42 +1,46 @@
 {
-  "parser": "@babel/eslint-parser",
```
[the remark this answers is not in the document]
khamui: i guess it is fine, if it does not lead to linting issues? I guess we need this part of eslint-config for jsx/js files?
And the part below is the config for tsx/ts files?

meinberlin/apps/contrib/assets/CountDown.tsx
goapunk: @khamui I agree that the return type is annoying, I initially had the second option you posted but the linter was complaining about the missing return type. From what I understand typescript will infer `JSX.Element` instead of `React.FC`, so I guess the question is whether we want the extra type checking we get by adding `React.FC` or we prefer simpler syntax. Option 1 doesn't work for me.

I'm torn between the easier syntax and the proper type checking, but currently tend to say we should probably go for the strict-but-ugly way

.eslintrc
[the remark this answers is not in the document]
goapunk: yeah, using eslint on a mixed code-base is a bit hacky. I wasn't sure if setting the ecma version to 13 is a problem but since we use babel to transpile the code during build anyway it shouldn't matter

meinberlin/apps/contrib/assets/CountDown.tsx
phillimorland: I do have a preference for easier syntax although looking through some typescript examples, if we decide to go with it, I feel it will not be particularly readable.

khamui: @phillimorland i agree that we have to get use to some of typescript words and "syntax", but i am really torn about additions that does not provide any further information. In this particular case `React.FC` does not help us. We know by the function component notation that it is a `FC`. In those cases i would prefer a config, which allows us to omit this unnecessary additions. I am afraid that it makes our code harder to read, to understand and that it slows us down in the end.

@goapunk regarding the React.FC i found this from Facebook/React:
https://github.com/facebook/create-react-app/pull/8177

i find the following better, would that work?
`const Component1 = (props: CountDownProps): JSX.Element => { /*...*/ }`

similar from here https://stackoverflow.com/a/61547010

goapunk: The biggest point people critize about `React.FC` seems to be as in your link: `Provides an implicit definition of children` which is no longer true for react 18. I do agree that we can go without it however, question then is if we want to add the JSX.Element or infer it

meinberlin/apps/actions/assets/timestamps.ts
```diff
@@ -0,0 +1,16 @@
+import { format } from 'timeago.js'
```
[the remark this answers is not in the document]
khamui: somehow the type annotation there is weird to me.
I tried something and ended up with a fully typed version of this function. Maybe to much?

```
const formatTime: () => void = (): void => {
  const times: NodeListOf<HTMLTimeElement> =
    document.querySelectorAll('time.relative')

  times.forEach((element: HTMLTimeElement) => {
    const sevenDays: number = 60 * 60 * 24 * 7 * 1000
    const datetime: Date = new Date(element.dateTime)

    if (new Date().valueOf() - datetime.valueOf() < sevenDays) {
      element.textContent = format(datetime, 'de')
    }
  })
}
```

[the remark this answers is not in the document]
khamui: i know, it is a bit contradictive to what I said about the typing of functions, but in these cases it is helpful.
I kind of like the typing of every single variable, for instance the `NodeListOf<>` and then the single element `HTMLTimeElement` is helpful imho..

what do you think?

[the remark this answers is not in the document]
goapunk: same, It's very helpful because it will warn you if an attribute doesn't exist for example. It just took me a bit to figure out what type I need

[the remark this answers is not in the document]
goapunk: @khamui 
`const formatTime: () => void = (): void => {`
I find this hard to read and I don't think declaring the function type here adds anything helpful. Adding the return type makes sense, because you need to know what the function has to return, knowing the function type on the other side at best tells us it's not a react component, but we already know that because it's returning void.

`const times: NodeListOf<HTMLTimeElement> = document.querySelectorAll('time.relative')`
Works for me as well, I do prefer my shorter version though.

`times.forEach((element: HTMLTimeElement) => {`
I prefer to leave out the type here as it is safely inferred and you can get the information by looking at the type of `times` easily.  But yeah, just personal preference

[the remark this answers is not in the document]
goapunk: haha yeah, I guess we will all have different preferences. I kind of agree with your `NodeListOf` argument, will change that!

meinberlin/apps/contrib/assets/Alert.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const closeStr: string = django.gettext('Close')
+
+export const Alert = ({
+  type,
+  message,
+  onClick
+}: AlertProps): JSX.Element | null => {
+  if (type !== undefined && type !== '') {
```
khamui, line 17: ts forced me to explicitly check both, which is a good thing.

goapunk:
closing as we won't use typescript for now.

PR #4526, opened 2022-09-27
[#6425] lighthouse fix for token input

phillimorland:


fuzzylogic2000 (approved):
It looks good and works! 
I will fix the tests. :)

PR #4532, opened 2022-09-29
[updates] js syntax updating 

phillimorland:
while looking for the js bug found some odd things?! so fixed them as didn't find much with the bug (wanted to feel less useless :p)


goapunk (approved):
I amended  a fix for the failing tests, looks good to me

PR #4540, opened 2022-10-04
jstests: test conditional rendering of filters after click

khamui:


meinberlin/apps/budgeting/assets/ControlBar.jsx
```diff
@@ -94,6 +94,7 @@ export const ControlBar = props => {
                   : 'btn btn--light'
               }
               aria-label={translated.toggleFilters}
+              data-testid="id_toggle-filters"
```
khamui, line 97: i don't like to add this explicitly, but not sure how to do it better?

goapunk: We should be able to use `findByLabelText` to find the element by aria-label if we change the mock of `gettext` a little. I will push a PR for that and then it should work if you use it here

goapunk: pushed https://github.com/liqd/a4-meinberlin/pull/4541

goapunk (changes requested):
cool, just the note about using `getLabelText` instead of the test ids

meinberlin/apps/budgeting/assets/__tests__/ControlBar.jest.jsx
```diff
@@ -30,3 +30,18 @@ test('ControlBar shows filters', () => {
...
+  render(
+    <ControlBar
+      filters={filters}
+      onChangeFilters={() => onChangeFiltersFn()}
+      numOfResults={2}
+    />
+  )
+  const filterButton = screen.getByLabelText('Toggle filters')
+  fireEvent.click(filterButton)
+  const expandedFilter = screen.getByText(/category1/)
```
goapunk, line 45: sorry, didn't think about that earlier but should we maybe check that it's not expanded before the click ? e.g. `expected(expandedFilter).toBeFalsy` ?

goapunk (approved):
cool :)

PR #4542, opened 2022-10-04
[#6519] label on tiles

khamui:


khamui:
oh, wip again, fixing tests

meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
```diff
@@ -21,6 +22,40 @@ export const BudgetingProposalListItem = (props) => {
         safeLocale
       )}`
 
+  // Before passing this data to 'ListItemBadges' component
+  // it needs to be normalized as the data types are different
+  // category: null || object
+  // labels: [] || array
+  // budget: 0 || number
+  // pointLabel: '' || string
+  // modFeedback: [null, null] || array
```
fuzzylogic2000, line 31: Uh. This makes sense, but shouldn't we change the serializer to give better values? We will have to change this for TS anyway. And in a lot of places. Maybe here is a good place to start and find out how we would like to get data from the API that can be there or not? We have all the examples you can think of in one place. :p

meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
```diff
@@ -34,10 +69,9 @@ export const BudgetingProposalListItem = (props) => {
         <a href={proposal.url}>{proposal.name}</a>
       </h2>
       <ListItemBadges
-        category={proposal.category}
-        budget={proposal.budget}
-        pointLabel={proposal.point_label}
-        moderatorFeedback={proposal.moderator_feedback}
+        badges={badges.slice(0, BADGES_LIMIT)}
```
fuzzylogic2000, line 72: good idea to handle the whole number of badges thing in a variable!

fuzzylogic2000 (commented):
Yes, looks good and works!
But I am unsure about the isTruthy and if we shouldn't find another solution.
And I think, in the story, we were meant to cut very long categories and labels after a certain amount of chars and append ...

fuzzylogic2000:
Oh, and I thought of another thing: is that the right order of the labels? Shouldn't the labels (Merkmale) be last as only these can have multiple labels and all the moderator's status thing should probably always be shown?

meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
khamui: @fuzzylogic2000 that is true. I will look into how to send them similar from the serializer

khamui:
> Oh, and I thought of another thing: is that the right order of the labels? Shouldn't the labels (Merkmale) be last as only these can have multiple labels and all the moderator's status thing should probably always be shown?

Good question, idk. Bringing @mcastro-lqd into the loop..what do you think?

PR #4544, opened 2022-10-05
CategorySelect: manually trigger select2()

goapunk:
fix the broken select in the live event


goapunk:
@khamui @phillimorland @fuzzylogic2000 @Rineee I'm a bit unsure about this, we do want to get rid of jquery (and therefore select2) so we should probably replace this altogether and use something else. The problem is there is no vanilla js library afaik, so we'd either have to:
1. build one ourselves
2. use react-select for react components and something else for non-react selects (annoying because we need two libraries or write everything in react)
3. use bootstrap with the browser inbuilt select (can't really customize this one, so not a real option)
4. ??

phillimorland:
@goapunk I might have a css only solution for this but will test and let you know, our current solution is not a11y compliant anyway so we do have to change it

phillimorland (approved):
tested the react-select but didn't get it working properly and it SR was reading it out oddley so have just added a fixme for now, maybe will be more inspired tomorrow

PR #4546, opened 2022-10-05
fix header image size in plans

goapunk:


goapunk:
fixes the blurry upscaled images on the vorhaben header image (https://meinberlin-dev.liqd.net/vorhaben/2022-00087/)

Rineee:
Looks good to me! Should we wait for @phillimorland to verify?

Rineee (approved):

PR #4553, opened 2022-10-07
WIP budgeting/react_proposals: pass initial query params in templatetag

Rineee:
I wanted to fix the following problem: when coming from the map view, the query params are in the url and thus react does not know about them, so all filters get lost. I am now passing them in the templatetag (both as querystring and dictionary, maybe we dont need both), but then could not make it work that the filters in ControlBar actually take those values. I was always running into troubles with state updates and useEffect ending up in a circle :woman_shrugging: 
I thought maybe @khamui you could have a look as you know best how the filters work?

### DIFF
--- a/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
+++ b/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
@@ -10,7 +10,7 @@ const nothingStr = django.gettext('Nothing to show')
 export const BudgetingProposalList = (props) => {
   const [data, setData] = useState([])
   const [meta, setMeta] = useState()
-  const [queryString, setQueryString] = useState(props.initialQueryString)
+  const [queryString, setQueryString] = useState('')
 
   const fetchProposals = (newIndex) => {
     const pageNumber = newIndex || 1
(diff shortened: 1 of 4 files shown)


khamui:
@Rineee I added a commit to make the switch from Map, back to List using the query params as initial filter setting.
I am not happy with the solution. It solves it just partly (does not work for the free text filter) and i think we are running after more and more issues.

One issue that comes with this solution is that we have to modify the url (not the fetch url, but the actual in the browser). For example if you have searched for a term before `&search=something`, then we could make it work that it searches for this word. But if we want to search for another word, we have to modify the url, which would lead to a page reload.

I guess to solve this we would need to use something like `react-router`. And therefore we would need to have a component structure like: `BudgetingViewContainer` which has `BudgetingListView` and `BudgetingMapView` as children. The switching between those components would be managed by `react-router`.

fuzzylogic2000:
@Rineee @khamui What was the status? We can close this PR, right?

Rineee:
> @Rineee @khamui What was the status? We can close this PR, right?

Yes, as we merged the router solution!

PR #4562, opened 2022-10-13
assets/dropdown: update focus colour and update aria label, rm pointl…

phillimorland:
…ess class

### DIFF
--- a/meinberlin/apps/budgeting/assets/ControlBar.jsx
+++ b/meinberlin/apps/budgeting/assets/ControlBar.jsx
@@ -7,7 +7,8 @@ import { SpacedSpan } from './SpacedSpan'
 import django from 'django'
 
 const translated = {
-  toggleFilters: django.gettext('Toggle filters'),
+  showFilters: django.gettext('Show filters'),
+  hideFilters: django.gettext('Hide filters'),
   filters: django.gettext('Filters'),
   results: django.gettext('results found.')
 }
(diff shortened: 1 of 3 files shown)


meinberlin/apps/budgeting/assets/ControlBar.jsx
```diff
@@ -88,12 +89,12 @@ export const ControlBar = props => {
           )}
           <div className="control-bar__item control-bar__right">
             <button
-              className={
+              className="btn btn--light"
+              aria-label={
                 expandFilters
-                  ? 'btn btn--light active'
```
phillimorland, line 93: there's no styling for active class on btns

fuzzylogic2000 (approved):
Cool!

meinberlin/apps/budgeting/assets/ControlBar.jsx
khamui: @phillimorland hm, i was sure that worked? maybe because of bootstrap. 

In these cases i am not sure what to do though, the design asks for a visual active state (grey background). Now there is no background color anymore right?

phillimorland: @khamui btn had active style after click until there was another action on the page (like filter click) then it returns to a non active state and style which was the same behaviour as when the class was added

PR #4573, opened 2022-10-17
filters via router

### LINKED ISSUE #4522: filter budgeting list: pluralisation of wirding for results
**URL:** 
**user:** 
**expected behaviour:** 
**behaviour:** For the freetext filter, when there was only one proposal found, it says "1 results found"
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

### LINKED ISSUE #4550: proposal list: result count for filtered list depends on pagination
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-3-phasen/?mode=list
**user:** 
**expected behaviour:** All proposals within the filtered list are counted
**behaviour:** If there are more proposals in the filtered list than there are on the first page, below the filters it says "15 results found", even though that is only the first page of the results.
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

### LINKED ISSUE #4568: budgeting freetext filter: if 0 results number of results is displayed twice
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-3-phasen-testprojekt/?mode=list
**user:** any
**expected behaviour:** 
**behaviour:** when there are no results it says "0 results found" above and "keine Ergebnisse" below search tag.
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2022-10-14 um 11 10 44

### LINKED ISSUE #4570: Switching from Map to List (react) does not apply filters (from before)
**URL:** 
**user:** 
**expected behaviour:** Filters that has been set in the list view should be applied again, when switching from Map to List
**behaviour:** Nothing happens
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

khamui:
@Rineee i implemented the entire filter logic by using `react-router` and i think it solves all our problems. Currently in WIP because pagination is broken.

But all other concerns regarding the filters are much easier now with the router hooks `useSearchParams` to extract query words from the url and to fetch with `useLocation`, so url and fetch-url are the same!

It might make our other PR obsolete #4553 .

issues:
fixes #4568 
fixes #4570 
fixes #4550
fixes #4522

### DIFF
--- a/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
+++ b/meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
@@ -4,23 +4,22 @@ import { BudgetingProposalListItem } from './BudgetingProposalListItem'
 import { Pagination } from './Pagination'
 import { CountDown } from '../../contrib/assets/CountDown'
 import { ControlBar } from './ControlBar'
-
-const nothingStr = django.gettext('Nothing to show')
+import { useLocation, useSearchParams } from 'react-router-dom'
 
 export const BudgetingProposalList = (props) => {
   const [data, setData] = useState([])
   const [meta, setMeta] = useState()
-  const [queryString, setQueryString] = useState('')
+  const location = useLocation()
+  const [queryParams] = useSearchParams()
 
-  const fetchProposals = (newIndex) => {
-    const pageNumber = newIndex || 1
-    const url = `${props.proposals_api_url}?page=${pageNumber}${queryString}`
+  const fetchProposals = () => {
+    const url = props.proposals_api_url + location.search
     fetch(url)
       .then(resp => resp.json())
       .then(json => {
         setData(json.results)
         setMeta({
-          current_page: pageNumber,
+          current_page: queryParams.get('page') || 1,
           page_count: json.page_count,
           is_paginated: json.page_count > 1,
           previous: json.previous,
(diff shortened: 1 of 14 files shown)


meinberlin/apps/budgeting/assets/Pagination.jsx
```diff
@@ -1,31 +1,37 @@
...
     elidedRange,
     nextPage,
-    onPaginate,
     prevPage
   } = props
+  const [queryParams, setQueryParams] = useSearchParams()
+  const currPage = parseInt(queryParams.get('page')) || 1
 
   const getKeyStr = (num, idx) => {
     return `page-${typeof num !== 'number' ? `ellip-${idx}` : num}`
```
fuzzylogic2000, line 20: I know this is not new, but still: whaat?

fuzzylogic2000 (approved):
Looks mostly good to me (some questions) and works! But I think @phillimorland or @goapunk should also have a look!

meinberlin/apps/budgeting/assets/ControlBar.jsx
```diff
@@ -96,11 +96,12 @@ export const ControlBar = props => {
...
         </div>}
-      <div className="offset-lg-2 col-lg-8">
-        <div className="control-bar">
-          {`${props.numOfResults} ${translated.results}`}
-        </div>
-      </div>
+      {props.numOfResults && (
+        <div className="offset-lg-2 col-lg-8">
+          <div className="control-bar">
+            {`${props.numOfResults} ${translated.results}`}
```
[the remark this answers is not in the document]
fuzzylogic2000: Yes, you did that in the next commit. :)

fuzzylogic2000:
Right, and also the frontend tests are failing.

meinberlin/apps/budgeting/assets/ControlBar.jsx
```diff
@@ -25,6 +25,12 @@ export const ControlBar = props => {
     } else {
       queryParams.delete(filterType)
     }
+
+    // to avoid empty pagination page for given
+    // filter settings, always show first page,
+    // except of sorting which returns same pagination range
+    filterType !== 'ordering' && queryParams.delete('page')
```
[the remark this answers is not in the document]
khamui: yes i wasn't totally sure about this. while testing i found it more confusing to be back in the first page though, thats why i added this exception for 'ordering'.

meinberlin/apps/budgeting/assets/Pagination.jsx
khamui: now that you pointed it out: yes too hard to read. will try to make it more readable.

meinberlin/apps/budgeting/assets/ControlBarDropdown.jsx
```diff
@@ -1,31 +1,33 @@
...
+// case 2: if there is a current, get the name of the current
+// case 3: if there is no default, get the name of the first choice
+const getDefaultName = (f, curr) => {
+  let filterChoice
+  if (curr) {
+    filterChoice = f.choices.filter(c => c[0] === curr)[0]
+  } else if (f.default) {
+    filterChoice = f.choices.filter(c => c[0] === f.default)[0]
+  } else {
+    filterChoice = f.choices[0]
```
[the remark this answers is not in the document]
khamui: not sure if i follow, this is to set the wording of the filters. Not the actual filter itself.
If there is no default or current set it will use the first of the choices name. So you suggest to return always a translated 'all' ?

meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -4,23 +4,23 @@ import { BudgetingProposalListItem } from './BudgetingProposalListItem'
...
   const [meta, setMeta] = useState()
-  const [queryString, setQueryString] = useState('')
+  const location = useLocation()
+  const [queryParams] = useSearchParams()
 
-  const fetchProposals = (newIndex) => {
-    const pageNumber = newIndex || 1
-    const url = `${props.proposals_api_url}?page=${pageNumber}${queryString}`
+  const fetchProposals = () => {
+    const url = `${props.proposals_api_url}${location.search}`
```
goapunk, line 16: personal pref: as there is no need for the template here just use `props.proposals_api_url + location.search` ?

meinberlin/apps/budgeting/assets/ControlBarDropdown.jsx
```diff
@@ -1,31 +1,33 @@
-import React, { useEffect, useState } from 'react'
+import React, { useState } from 'react'
+
+// FIXME: maybe more elegant way?
+// this is only relevant for the first page load,
+// where the url does not contain all default query params.
+// case 1: if there is a default, get its verbose name
+// case 2: if there is a current, get the name of the current
```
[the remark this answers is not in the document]
khamui: actually no, it can have a default, but also a current. In this case current to be taken over default.

goapunk (approved):
looks very good to me! only a mock for `ngettext` missing so the tests will pass

meinberlin/apps/budgeting/assets/ControlBar.jsx
```diff
@@ -1,88 +1,66 @@
...
   hideFilters: django.gettext('Hide filters'),
-  filters: django.gettext('Filters'),
-  results: django.gettext('results found.')
+  filters: django.gettext('Filters')
+}
+
+const getResultCountText = (count) => {
+  const foundProposalsText =
+    django.ngettext('1 proposal found.', '%s proposals found.', count)
+  return django.interpolate(foundProposalsText, [count])
```
goapunk, line 19: Not really sure what interpolate does here, it reads like it's mostly if there is more than one number?

khamui: tbh i just copied from `BudgetProposalList`, tried without but doesn't seem to work without `interpolate()`.

goapunk: ah, it's different in js than in python! cool, works nicely

fuzzylogic2000 (approved):
Nice!

PR #4576, opened 2022-10-18
budgeting/map/switchbutton: fix #4523

### LINKED ISSUE #4523: Buttons "Map" and "List" don't work properly on mobile version
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-3-phasen-testprojekt/
**user:** participant
**expected behaviour:** Changing the overview of the projects from map to list should work by clicking on the according button.
**behaviour:** On the mobile version there is a problem regarding the "Map" and "List" button. In most of the projects it is not possible to switch between the views, but it works on the desktop version.
**important screensize:** Samsung Galaxy S9+, Iphone 13
**device & browser:** Mobile phone Samsung Galaxy S9+, Google Chrome; Iphone 13, Safari
**Comment/Question:** Both buttons work also on the mobile version, however, on the project https://meinberlin-dev.liqd.net/projekte/carolin-findet-ihrberlin-2/ . For this project there are no filter options neither for the map nor for the list (except for the list sorting function). Maybe it has something to do with that.

khamui:


phillimorland (approved):
cool

PR #4577, opened 2022-10-18
assets/poll: remove popper styling and package and update poll styling

phillimorland:


meinberlin/assets/scss/components/_poll.scss
```diff
@@ -187,14 +192,13 @@ $checkbox-size: 20px;
 }
 
 .editpoll__question-container {
-    // display: block;
+    display: flex;
```
khamui, line 195: from screensize 992px to 991px it looks off

khamui (commented):
one comment and frontend tests are failing (bc. of ErrorList location)

meinberlin/assets/scss/components/_poll.scss
phillimorland: @khamui having trouble replicating also that's 1px difference, is that the size you mean?

jest.config.js
```diff
@@ -1,7 +1,11 @@
 const config = {
   testEnvironment: 'jsdom',
   modulePaths: [
-    '<rootDir>'
+    '<rootDir>',
+    '!venv'
```
goapunk, line 5: why do we need this? we have the `modulePathIgnorePatterns` for it?

jest.config.js
```diff
@@ -1,7 +1,11 @@
...
+  ],
+  modulePathIgnorePatterns: [
+    '<rootDir>/venv/*'
```
goapunk, line 8: the asterisk is redundant here

meinberlin/assets/scss/components/_poll.scss
khamui: @phillimorland  i doublechecked and this occurs for me (in chrome), but not sure if that change was intended?

Bildschirmfoto 2022-10-24 um 13 37 04

before the questions were spanning all the way to the right edge?

Bildschirmfoto 2022-10-24 um 13 39 07

phillimorland: cheers for screen shots, change wasn't intended, @goapunk is doing some jest config updates right now, will look at this once that's pushed

phillimorland: @khamui ready for another look, can't tag you again

khamui (approved):
very nice! working great. Is it still depending on a4?
Leaving merge to @goapunk 

Question: do we need to update eslint and jest config in other places where we use them as well?

goapunk:
> very nice! working great. Is it still depending on a4? Leaving merge to @goapunk
> 
> Question: do we need to update eslint and jest config in other places where we use them as well?

This already comes from a+ so at least there it's not necessary

goapunk:
I bumped the a4 hash again so we don't have to do it again soon and set it to merge once CI passed

Issue #4574, opened 2022-10-18
Budgeting: Votes are not shown on Map Pins


khamui:
**URL:** 
**user:** 
**expected behaviour:** Clicking on pins on map should show vote counts
**behaviour:** Switching from list to map does not show the vote counts when clicked on the pins on the map
**important screensize:**
**device & browser:** 
**Comment/Question:** 

**if `Ordering` is set to `-daily_random` it does not appear**

<img width="1548" alt="Bildschirmfoto 2022-10-18 um 09 49 42" src="https://user-images.githubusercontent.com/52801593/196383549-08f741f2-b7f2-4cee-a97d-36dae0939512.png">

**if `Ordering` is set to `-created` it appears**

<img width="1542" alt="Bildschirmfoto 2022-10-18 um 09 49 57" src="https://user-images.githubusercontent.com/52801593/196383651-0082470f-2f3d-47e2-b547-2b387f18f0af.png">

PR #4582, opened 2022-10-19
budgeting//contrib/filters: add random ordering to ProposalFilterSet …

Rineee:
…and rename filter in api to be able to pass it

depends on https://github.com/liqd/adhocracy4/pull/1246

fixes https://github.com/liqd/a4-meinberlin/issues/4574

I had to add the daily random sorting to the list view as otherwise the annotations dont work and thus dont show up in the map pop ups. 

The daily random sorting is different for the django list and react list, that is because in the react list we are only ordering the proposals of a certain module whereas in the django list, the ordering is determined on all proposal in the db and then later the proposals are filtered for the module. I did not dare to touch that as that is deep in a4 and django. But since we will always have either or, it should be fine?


meinberlin/apps/budgeting/api.py
```diff
@@ -112,12 +112,12 @@ def list(self, request, *args, **kwargs):
         if has_feature_active(self.module, Proposal, 'support'):
             ordering_choices += ('-positive_rating_count', _('Most support')),
         ordering_choices += ('-comment_count', _('Most commented')), \
-                            ('-daily_random', _('Random')),
+                            ('dailyrandom', _('Random')),
```
fuzzylogic2000, line 115: :+1: It doesn't matter if it's ascending or descending random. :rofl:

fuzzylogic2000 (approved):
Nice! 
And an interesting find, that we order the Django a4-one before filtering for the module. But I also think you are right, it doesn't matter as we always use either or and don't need the same random ordering in both cases.
Happy that you merge when you updated a4!

PR #4583, opened 2022-10-19
Pm 2022 10 wagtail update 2

phillimorland:


goapunk:
@phillimorland I just checked out your PR and make install worked without error. Running `makemigrations` says there are no new migrations and tests all pass. So all good here

phillimorland:
@goapunk I didn't add the breaking commit

meinberlin/apps/cms/migrations/0039_alter_homepage_body.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('meinberlin_cms', '0038_streamfield_add_use_json_field'),
+    ]
+
+    operations = [
+        migrations.RunPython(empty_to_valid_json),
```
goapunk, line 25: The migration for the `use_json_field` fails if the `body` of the homepage is blank, which it seems to be initially when starting with a fresh db. Once you edit the homepage you have to set all values. This migration sets the body to a valid json string, this will make the migration work.

meinberlin/apps/cms/migrations/0039_alter_homepage_body.py
```diff
@@ -0,0 +1,31 @@
...
+import wagtail.blocks
+import wagtail.fields
+import wagtail.images.blocks
+
+
+def empty_to_valid_json(apps, schema_editor):
+    HomePage = apps.get_model('meinberlin_cms', 'HomePage')
+
+    for homepage in HomePage.objects.all():
+        if len(homepage.body) == 0:
```
goapunk, line 13: I'm not sure if this is the best way to check but it seems to work

phillimorland: I would say either @fuzzylogic2000 or @Rineee are better people to ask about that?

meinberlin/config/settings/base.py
```diff
@@ -261,7 +261,7 @@
 
 # Base URL to use when referring to full URLs within the Wagtail admin backend -
 # e.g. in notification emails. Don't include '/admin' or a trailing slash
-BASE_URL = 'http://localhost:8000'
+WAGTAILADMIN_BASE_URL = 'http://localhost:8000'
```
fuzzylogic2000, line 264: I think this should be `localhost:8003`, but that can be for another day.

meinberlin/apps/cms/migrations/0039_alter_homepage_body.py
fuzzylogic2000: I also don't know if it's the best way, but if it works it's fine! And it makes sense for what it's meant to check. :)

phillimorland: @fuzzylogic2000 @goapunk i' happy to merge if you guys are?

phillimorland: @goapunk we have a backup db incase it breaks dev right? :p

PR #4585, opened 2022-10-19
apps/item detail templates: use label/badges from model property and …

fuzzylogic2000:
…move badge properties to mixin to use in topics

I used the badge properties also in the detail views. But we need to check what to do with the budget badges on the tile. 

So, questions for you @CarolingerSeilchenspringer :
We did use different solutions for all the different badges and all different idea derivates. With redoing the badges on the list, I put them into the same order and added the same behaviour everywhere:
- order (detail view and list): moderator feedback, budget, point_label, category, labels
- behaviour on the list: after three badges, the others are counted and put behind the badges as link "... more"
That is fine, right?

But for the detail view, we only added the "budget not specified" to budgeting proposals, not kiezkasse. I also changed that and kiezkasse proposals are the same now. Also fine (just improving), right?

But finally the question we need a decision on: so far the budget is not displayed in the list when it is 0. https://meinberlin-dev.liqd.net/projekte/module/kiezkasse-4/?mode=list It would be easier if we could change that and also display the 0-label on the list: 
Screenshot from 2022-10-19 17-05-36
But we can also remove it from the list. And keep it like it is now. Please tell us @CarolingerSeilchenspringer

### DIFF
--- a/meinberlin/apps/budgeting/assets/__tests__/ListItemBadges.jest.jsx
+++ b/meinberlin/apps/budgeting/assets/__tests__/ListItemBadges.jest.jsx
@@ -3,59 +3,29 @@ import { render, screen } from '@testing-library/react'
 import { ListItemBadges } from '../ListItemBadges'
 
 // testing data:
-const categoryBadge = [{
-  type: 'category',
-  value: { id: 0, name: 'category1' }
-}]
-
-const labelsBadges = [
-  {
-    type: 'label',
-    value: { id: 0, name: 'label1' }
-  },
-  {
-    type: 'label',
-    value: { id: 1, name: 'label2' }
-  },
-  {
-    type: 'label',
-    value: { id: 2, name: 'label3' }
-  }
+const someBadges = [
+  ['category', 'category1'],
+  ['label', 'label1'],
+  ['label', 'label2']
 ]
 
-const pointLabelBadge = [{
-  type: 'pointLabel',
-  value: 'labelwithicon'
-}]
+const pointLabelBadge = [
+  ['point_label', 'labelwithicon']
+]
 
-const budgetBadge = [{
-  type: 'budget',
-  value: '20000'
-}]
-
-const modFeedbackBadge = [{
-  type: 'modFeedback',
-  value: ['CONSIDERATION', 'Under consideration']
-}]
-
-test('displaying category badge', () => {
-  render(
-    <ListItemBadges
-      badges={categoryBadge}
-    />
-  )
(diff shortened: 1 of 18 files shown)


fuzzylogic2000:
@CarolingerSeilchenspringer That was a long text and a lot of questions yesterday, but the only question actually is: Can we keep the "Kosten: keine Angabe" when the budget is 0 on the lists for Bürgerhaushalt and Kiezkasse or shouldn't there be any badge in that case?

CarolingerSeilchenspringer:
@fuzzylogic2000 checked and find all three solutions you mention good. Have it the same everywhere sounds the best option.

package.json
```diff
@@ -20,6 +20,7 @@
     "autoprefixer": "10.4.12",
     "babel-loader": "8.2.5",
     "bootstrap": "5.2.2",
+    "cookie-js": "^0.0.1",
```
Rineee, line 23: @fuzzylogic2000 Was this intended?

fuzzylogic2000: No, will fix that, thanks!

fuzzylogic2000:
@Rineee Added some more things. :)

meinberlin/apps/budgeting/serializers.py
```diff
@@ -48,20 +20,18 @@ class ProposalSerializer(serializers.ModelSerializer):
     class Meta:
         model = Proposal
         fields = (
-            'budget', 'category', 'comment_count', 'created',
-            'creator', 'is_archived', 'labels', 'moderator_feedback',
+            'additional_item_badges_for_list_count', 'comment_count',
```
Rineee, line 23: Cool, I never realized that you can serialize properties just like that!

meinberlin/apps/budgeting/assets/__tests__/ListItemBadges.jest.jsx
```diff
@@ -3,59 +3,29 @@ import { render, screen } from '@testing-library/react'
...
+]
 
-test('displaying category badge', () => {
-  render(
-    <ListItemBadges
-      badges={categoryBadge}
-    />
-  )
-  expect(screen.getByText('category1')).toBeTruthy()
-})
```
fuzzylogic2000, line 48: I was a bit sad to delete tests, but as the component got a lot dumber with the serializer having all the info, they didn't make sense anymore. When I got it right, it only tests if the badges are there, right? We could still test if the moderator_statement is yellow and if the point_label has the icon or do the tests magically do that? @khamui Maybe you could have a look? (I will merge if the tests are actually all fine this time, though. :sweat_smile: )

PR #4602, opened 2022-10-21
[6560] apps/moderator feedback: add new choice and filter to budgeting react list

fuzzylogic2000:
I am still writing tests, but this should be fine to be merged and testes on dev


khamui:
Nice, working great!

PR #4605, opened 2022-10-24
btn-light/styles: remove grey background on focus

### LINKED ISSUE #4547: budgeting filter buttons don't lose focus/hover after click
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-3-phasen-testprojekt/?mode=list
**user:** any
**expected behaviour:** button state turns to not hover/not focus, background turns white
**behaviour:** button state stays focused/hovering, background is grey
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

khamui:
fixes #4547

### DIFF
--- a/meinberlin/assets/scss/components/_button.scss
+++ b/meinberlin/assets/scss/components/_button.scss
@@ -21,7 +21,6 @@
     font-weight: normal;
     border-color: $border-color;
 
-    &:focus,
     &:hover,
     &:active {
         background-color: $bg-secondary;


fuzzylogic2000 (commented):
I find the focus while tabbing a bit hard to see without that (at least in firefox). :( But we also don't want the button to stay grey, so don't know.

khamui:
Would it help to change the border color of the focus instead of using the background color @fuzzylogic2000 ? I think the search input uses the primary color for the focus ring.

fuzzylogic2000:
I don't know. I guess that's a question for @phillimorland ? Might also be fine as it is here?

phillimorland:
@khamui @fuzzylogic2000 we definitely want to keep a focused state style I would say, i'm also not really sure what the issue is to be honest, that the button remains grey? that's because it's focussed when focus moves the style is removed which is correct behaviour

fuzzylogic2000:
@phillimorland The grey focus-ring is still there in this PR, but I find it harder to see than when the whoole button is grey. I think the problem in the issue is, that the ring is gone, but the (filter expand) button is still grey after it was used.

phillimorland:
> @phillimorland The grey focus-ring is still there in this PR, but I find it harder to see than when the whoole button is grey. I think the problem in the issue is, that the ring is gone, but the (filter expand) button is still grey after it was used.

I think this might be an issue of expectation, it's still grey because after being clicked the focus is still on the button, if you tab through and press enter to open the filters, you must tab again to move the focus to the next item so the behaviour is correct for keyboard users but I guess not great for mouse users? not really sure what to do here?!

fuzzylogic2000:
I think, we should take the discussion to the issue, then. I already took the last comment there.

PR #4606, opened 2022-10-24
budgeting: add rule to view support also between phases and use rule …

Rineee:
…for map pop up in map view

~depends on https://github.com/liqd/adhocracy4/pull/1251~

Did not yet apply the rule in list and detail views, but added separate tasks in story.

### DIFF
--- a/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
+++ b/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
@@ -55,7 +55,12 @@
                     </div>
                 </div>
             </div>
-            {% map_display_points object_list module.settings_instance.polygon %}
+            {% if module.blueprint_type == 'PB3' %}
+                {% has_or_would_have_perm 'meinberlin_budgeting.view_support' request.user module as may_view_support %}
+                {% map_display_points object_list module.settings_instance.polygon 'true' may_view_support|yesno:"false,true" %}
+            {% else %}
+                {% map_display_points object_list module.settings_instance.polygon %}
+            {% endif %}
         </div>
     {% else %}
         {% react_proposals view.module %}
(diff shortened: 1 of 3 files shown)


meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html
```diff
@@ -55,7 +55,12 @@
                     </div>
                 </div>
             </div>
-            {% map_display_points object_list module.settings_instance.polygon %}
+            {% if module.blueprint_type == 'PB3' %}
+                {% has_or_would_have_perm 'meinberlin_budgeting.view_support' request.user module as may_view_support %}
+                {% map_display_points object_list module.settings_instance.polygon 'true' may_view_support|yesno:"false,true" %}
```
Rineee, line 60: This is a bit hacky with the [yesno template tag](https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#yesno), but did not really find a better solution for negating a bool in template :woman_shrugging:

Rineee:
@fuzzylogic2000 I decided to do the detail and list view permissions in another PR, so this one could already be reviewed, it only handles the map pop up. I added a task for the missing test..

package.json
```diff
@@ -16,7 +16,7 @@
     "@fortawesome/fontawesome-free": "6.2.0",
     "@maplibre/maplibre-gl-leaflet": "0.0.17",
     "acorn": "8.8.1",
-    "adhocracy4": "liqd/adhocracy4#23e09419fe20c5a4f86ba0f9fa6ef33d268f5e92",
+    "adhocracy4": "liqd/adhocracy4#ec13ccd26008108a45379db95829655de241c5e5",
```
goapunk, line 19: mb is already on https://github.com/liqd/adhocracy4/commit/23e09419fe20c5a4f86ba0f9fa6ef33d268f5e92 so we probably don't need to update again ? :)

Rineee: oh yes, thanks, will remove the commit!

fuzzylogic2000 (approved):
Cool!

PR #4609, opened 2022-10-24
Pm 2022 10 styling issues

phillimorland:


khamui (approved):
Really cool fixes!

Just one question about the button text color, it is now secondary, in the designs it is black(ish). Is that intended?

fuzzylogic2000 (approved):
Looks good! What's the status? Can we rebase and merge?

phillimorland:
@fuzzylogic2000 if you wouldn't mind that would be fab! :)

PR #4615, opened 2022-10-25
base.txt: update wagtail, cms/settings: update BaseSetting to BaseSit…

phillimorland:
…eSetting


goapunk (approved):
looks good to me. There's a console error on the image tab in wagtail but we have the same on liqd site, so probably ok to ignore

PR #4618, opened 2022-10-25
renovate: wagtail < 4.2

goapunk:
stay on LTS

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -32,6 +32,10 @@
       {
         "matchPackageNames": ["Django"],
         "allowedVersions": "=3.2"
+      },
+      {
+        "matchPackageNames": ["wagtail"],
+        "allowedVersions": "=4.1"
       }
     ],
 


renovate.json
```diff
@@ -32,6 +32,10 @@
       {
         "matchPackageNames": ["Django"],
         "allowedVersions": "=3.2"
+      },
+      {
+        "matchPackageNames": ["wagtail"],
+        "allowedVersions": "< 4.2"
```
fuzzylogic2000, line 38: Yeah! But can we go for one version? So, either here "=4.1" or above also "<3.3"

goapunk: I initially went for `=4.1` but then was unsure since we are still on `4.0` and wouldn't get updates for that.

fuzzylogic2000 (approved):
Cool!

renovate.json
fuzzylogic2000: Oh sorry, only saw that now. Yes, that makes sense. Let's change both!

PR #4622, opened 2022-10-25
fixes #4608

### LINKED ISSUE #4608: #6460 problem with selecting filters in project overview
**URL:** https://meinberlin-dev.liqd.net/projekte/
**user:** any
**expected behaviour:** To select a “kind of participation” or “project status” filter, I can click anywhere on the respective button.
**behaviour:** Filter only gets selected, when the text inside the button is clicked.
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2022-10-24 um 15 23 19 (2)

khamui:


fuzzylogic2000 (approved):
Nice! It works for the project overview filter, but I am not sure if there are sideffects in places where we don't want to have them? Do you know @phillimorland ?

goapunk:
One minor issue with this quick fix is that there is still a padding inside the button, clicking on that still doesn't do anything. I guess we need to decide whether we want to properly fix it (takes more time) or accept this for now

khamui:
@fuzzylogic2000 i will have another look and check if it has unwanted side effects.

@goapunk yes that was my idea, as this issue was/is not prioritized i decided to improve, but not able to fix it accordingly.

phillimorland:
@fuzzylogic2000 @khamui the whole element is still not clickable like @goapunk said, also the hover colour is broken but I think that came with the removal of the old library. They also stay active after initial click which as they are styled checkboxes not buttons shouldn't happen (I don't think?) so the colour after clicking is too dark and there is a border radius, not sure from when that is from? I would say they need a full refactor. If we don't have time for that then the whole element should be clickable

khamui:
@phillimorland I didn't manage to expand the clickable area to the entire width/height. Also the hover styling feels a bit hacky. Feel free to scrap this, if you have something better. Sorry for leaving this to you!

phillimorland:
closed in favor of https://github.com/liqd/a4-meinberlin/pull/4673

PR #4623, opened 2022-10-25
[6560] [6536] add tests

fuzzylogic2000:


Rineee (approved):
Cool! Good to split the api filter tests as they got more and more..

Issue #4596, opened 2022-10-20
Screenreader doesnt read error messages


SandisStudi:
**URL:** e.g. https://meinberlin-dev.liqd.net/dashboard/modules/burgerinnenhaushalt-3-phasen-10/generate-codes/

**user:** admin
**expected behaviour:** If I already reached the code generating limit of 5 Mio codes and try to generate more, so that there is a visual error message, that this is not possible, the most used screenreaders (JAWS and NVDA) should also tell me this
**behaviour:** NVDA doesnt read the visually shown error message, that the code generation limit is reached and it is not possible to generate more codes 
**important screensize:** -----
**device & browser:** Mac/Windows with Chrome, Firefox, Edge
**Comment/Question:**

CarolingerSeilchenspringer:
as discussed with @fuzzylogic2000 and @Rineee we need to check if this is true for all error messages, then globally change

phillimorland:
NOTE: error messages for jsx should now be read, I so know that I looked into adding `role=alert` to the django errors but failed, @fuzzylogic2000 @Rineee do you know how to add attributes to default form errors? the info alerts should have a role polite so we need to be able to alter them

fuzzylogic2000:
Uh, yes, I'll have a look! Could you point me at the changes for the jsx, so that I can add the same roles for errors, warnings and info alerts?

phillimorland:
@fuzzylogic2000 so I actually messed up with the jsx and just added alert but i'll change that today but as a temporary solution I did just add it to the surround div in a few places, might be easier? https://github.com/liqd/a4-meinberlin/blob/b75ddedf17e0c937ec657f9871140e5954b703d4/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html#L34
also just searched for `role="alert"` and we might want to have a little overhaul there, there quite a few that are not great

fuzzylogic2000:
@phillimorland Yes, adding it to the surrounding div instead of overwriting whatever Django does might be easier. But to fix it once and for all coming alerts, we should probably add it to the alert code. So, it's good enough for now, but we need to better fix it later?

phillimorland:
@fuzzylogic2000 I mean we assign styling depending on the error so we do kind of know that? so maybe there is a way to change the role too?! are you working on this right now?

PR #4633, opened 2022-10-26
Pm 2022 10 a11y errors

phillimorland:
fixes https://github.com/liqd/a4-meinberlin/issues/4596

dependent on https://github.com/liqd/adhocracy4/pull/1255

### DIFF
--- a/meinberlin/apps/budgeting/templates/meinberlin_budgeting/includes/proposal_form.html
+++ b/meinberlin/apps/budgeting/templates/meinberlin_budgeting/includes/proposal_form.html
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load i18n widget_tweaks %}
 <form enctype="multipart/form-data" action="{{ request.path }}" method="post">
     {% csrf_token %}
     {{ form.media }}
@@ -35,13 +35,17 @@
 
     <div class="form-check">
         <label class="form-check__label">
-            {{ form.contact_storage_consent }}
+            {{ form.contact_storage_consent|add_error_attr:"aria-invalid:true" }}
             {% blocktrans with privacy_policy="/datenschutz" %}I hereby expressly consent to the storage of the contact information I have entered. The contact information may be used by the responsible administration exclusively for contacting me in the present project. I can revoke this consent at any time. I have read and accept the <a href="{{privacy_policy}}" target="_blank">privacy policy</a>.{% endblocktrans %}
         </label>
-        {{ form.contact_storage_consent.errors }}
+        {% if form.contact_storage_consent.errors %}
+            <ul class="errorlist" aria-live="assertive" aria-atomic="true">
+            {% for error in form.contact_storage_consent.errors %}
+                <li>{{ error|escape }}</li>
+            {% endfor %}
+            </ul>
+        {% endif %}
     </div>
-
-
 
     <div class="u-spacer-bottom">
         <input type="submit" class="btn btn--primary" value="{% trans 'Save'%}" />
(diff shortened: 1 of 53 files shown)


phillimorland:
@fuzzylogic2000 tests are broken on a4 but will take a look tomorrow but this is it for now :)

package.json
```diff
@@ -29,6 +29,7 @@
     "file-saver": "2.0.5",
     "immutability-helper": "3.1.1",
     "jquery": "3.6.1",
+    "js-cookie": "^3.0.1",
```
fuzzylogic2000, line 32: No caret please. Oh right, and this got here by accident anyway, right?

fuzzylogic2000 (approved):
NIce! Just one question. And I will look into the other Django errors now.

package.json
phillimorland: yes it's an accident, for some reason my a4linking doesn't work without it any more, need to troubleshoot that but for now have just been installing it while linked, sorry forgot to remove

phillimorland:
@fuzzylogic2000 sorry didn't realise you had done hash, have re-added

meinberlin/apps/budgeting/templates/meinberlin_budgeting/includes/proposal_form.html
```diff
@@ -35,14 +35,20 @@ <h2>{% blocktrans %}Contact Information{% endblocktrans %}</h2>
 
     <div class="form-check">
         <label class="form-check__label">
-            {{ form.contact_storage_consent }}
+            {{ form.contact_storage_consent|add_error_attr:"aria-invalid:true" }}
```
fuzzylogic2000, line 38: I needed so long to find this, but the django_widget_tweaks can do that! :partying_face: 
https://pypi.org/project/django-widget-tweaks/

If it's fine like this @phillimorland , I will also add the individual errors to the non_field_errors and then do the same in a4.

meinberlin/apps/budgeting/templates/meinberlin_budgeting/includes/proposal_form.html
```diff
@@ -35,14 +35,20 @@ <h2>{% blocktrans %}Contact Information{% endblocktrans %}</h2>
...
             {% blocktrans with privacy_policy="/datenschutz" %}I hereby expressly consent to the storage of the contact information I have entered. The contact information may be used by the responsible administration exclusively for contacting me in the present project. I can revoke this consent at any time. I have read and accept the <a href="{{privacy_policy}}" target="_blank">privacy policy</a>.{% endblocktrans %}
         </label>
-        {{ form.contact_storage_consent.errors }}
+        {% if form.contact_storage_consent.errors %}
+            <ul class="errorlist">
+            {% for error in form.contact_storage_consent.errors %}
+                <li aria-live="assertive" aria-atomic="true" id="error-{{ form.contact_storage_consent.id_for_label }}-{{ forloop.counter }}">
```
fuzzylogic2000, line 44: @phillimorland I have added it to the list elements, but have seen that we added it to the errorlist elsewhere. What's the right place to put it?

phillimorland: @fuzzylogic2000 on the error list as the containing element, will also tell screen reader number of items on list instead of just list item

fuzzylogic2000:
Alright. I think this should be fine now?! I couldn't think of an example for a non_field_error, though and have not tested these, but some of the form field errors.

fuzzylogic2000:
@phillimorland Could you have another look at what I did? And then, we can merge? I like what you did! :)

phillimorland (commented):
yey! so awesome you found the aria invalid attribute!! :)

PR #4638, opened 2022-11-01
[6609 ]Ks 2022 10 support btw phases

### LINKED ISSUE #4109: budgeting: think about moving phase information to api
**URL:** 
**user:** 
**expected behaviour:** phase information is coming from api like all other data
**behaviour:** is_voting_phase is passed in the template tag
**important screensize:**
**device & browser:** 
**Comment/Question:** We need to know if voting phase is active to decide whether to show the ratings or not: https://github.com/liqd/a4-meinberlin/blob/d69a4dd624705c7f28a053e123e0dcdf847fbe3d/meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx#L23

Screenshot?

### LINKED ISSUE #4597: comment count missing on react list
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerhaushalt-3-phasen-testprojekt/?initialSlide=1&mode=list
**user:** any
**expected behaviour:** comment count is visible
**behaviour:** comment count is missing 
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2022-10-21 um 13 46 38

Rineee:
This is supposed to implement the following behaviour:
- show support (on, list, map and detail) during support phase and btw support and voting phase
- support button only active during support phase 
- add 'most support' to ordering filter during support phase and btw support and voting phase
- make 'most support' default ordering filter only btw support and voting phase

Notes: 
- do not test as admin
- old participatory budgeting with rating phase doesnt fully work anymore (map pop up does not show rating), but ok with Caro

I did not add the tests yet as did not want to test unapproved code..

fixes #4109
fixes #4597

### DIFF
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -13,6 +13,7 @@ from adhocracy4.categories import has_icons
 from adhocracy4.categories.models import Category
 from adhocracy4.labels.models import Label
 from adhocracy4.modules.predicates import module_is_between_phases
+from adhocracy4.phases.predicates import has_feature_active
 from meinberlin.apps.contrib.filters import IdeaCategoryFilterBackend
 from meinberlin.apps.contrib.filters import OrderingFilterWithDailyRandom
 from meinberlin.apps.contrib.templatetags.contrib_tags import \
@@ -184,6 +185,9 @@ class PermissionInfoMixin:
         )
         permissions['view_comment_count'] = (
             self.module.has_feature('comment', Proposal)
+        )
+        permissions['view_vote_count'] = has_feature_active(
+            self.module, Proposal, 'vote'
         )
 
         response = super().list(request, args, kwargs)
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -144,16 +144,14 @@ class PermissionInfoMixin:
         fetches the module
         """
         permissions = {}
-
-        permissions['view_support_count'] = has_feature_active(
-            self.module, Proposal, 'support'
-        )
+        user = request.user
+        permissions['view_support_count'] = user.has_perm(
+            'meinberlin_budgeting.view_support', self.module)
         permissions['view_rate_count'] = self.module.has_feature(
             'rate', Proposal
         )
         permissions['view_comment_count'] = (
             self.module.has_feature('comment', Proposal)
-            and not has_feature_active(self.module, Proposal, 'vote')
         )
 
         response = super().list(request, args, kwargs)
--- a/meinberlin/apps/budgeting/api.py
+++ b/meinberlin/apps/budgeting/api.py
@@ -12,7 +12,7 @@ from adhocracy4.categories import get_category_icon_url
 from adhocracy4.categories import has_icons
 from adhocracy4.categories.models import Category
 from adhocracy4.labels.models import Label
-from adhocracy4.phases.predicates import has_feature_active
+from adhocracy4.modules.predicates import module_is_between_phases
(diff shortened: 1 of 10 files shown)


meinberlin/apps/budgeting/api.py
```diff
@@ -56,36 +56,19 @@ def list(self, request, *args, **kwargs):
...
-            for category in categories:
-                category_choices += (str(category.pk), category.name),
-                if has_icons(self.module):
-                    icon_name = getattr(category, 'icon', None)
-                    icon_url = get_category_icon_url(icon_name)
-                    category_icons += (str(category.pk), icon_url),
-
+        category_choices, category_icons = \
+            self.get_category_choices_and_icons()
+        if category_choices:
```
Rineee, line 61: needed to split this into different functions as linter told me this has gotten too complex :see_no_evil:

meinberlin/apps/budgeting/api.py
```diff
@@ -185,6 +217,14 @@ class ProposalViewSet(ModuleMixin,
...
     search_fields = ('name', 'ref_number')
 
+    @property
+    def ordering(self):
+        if module_is_between_phases('meinberlin_budgeting:support',
+                                    'meinberlin_budgeting:voting',
+                                    self.module):
+            return ['-positive_rating_count']
+        return ['dailyrandom']
+
```
Rineee, line 231: this is adding the default ordering, as it is dynamic in our case, had to do it as property

tests/budgeting/test_proposals_api_filtering.py
```diff
@@ -373,7 +373,9 @@ def test_proposal_ordering_filter(
                   kwargs={'module_pk': module.pk})
 
     # queryset is ordered by created
-    response = apiclient.get(url)
+    querystring = '?ordering=-created'
+    url_tmp = url + querystring
+    response = apiclient.get(url_tmp)
```
Rineee, line 378: before created unintentionally was default :see_no_evil:

meinberlin/apps/budgeting/views.py
```diff
@@ -20,6 +20,8 @@ def get_ordering_choices(view):
     choices = (('-created', _('Most recent')),)
     if view.module.has_feature('rate', models.Proposal):
         choices += ('-positive_rating_count', _('Most popular')),
+    elif view.module.has_feature('support', models.Proposal):
+        choices += ('-positive_rating_count', _('Most support')),
```
fuzzylogic2000, line 24: This is for the list, we will not show later, right? But good to have it here, analogous to the rating!

tests/budgeting/test_proposals_api_filtering.py
fuzzylogic2000: but that's still the case, or did I miss the fix?

fuzzylogic2000 (approved):
Supernice! And works perfectly!

meinberlin/apps/budgeting/views.py
Rineee: No, it is for the pop up to work, because if the most support filter is set in list and it doesnt know it here, it breaks..

tests/budgeting/test_proposals_api_filtering.py
Rineee: No, it should not (without the ?ordering=-created). The fix is the ordering property in the ProposalViewSet, which sets the default ordering if the ordering filter hasnt been applied.. https://www.django-rest-framework.org/api-guide/filtering/#specifying-a-default-ordering

fuzzylogic2000: Right, you also order them here by created on purpose for the test... :+1:

fuzzylogic2000:
Cool, should we merge?

PR #4678, opened 2022-11-11
WIP: deps: disable dependency zeep as only used in servicekonto

fuzzylogic2000:
renovate didn't want to let me do that in it's PR


fuzzylogic2000:
Well, can't get it to run the tests and not fail today. Will try again on Monday.

fuzzylogic2000:
too complicated, maybe later

PR #4679, opened 2022-11-11
Ks 2022 11 ampel wording changes

Rineee:
Not sure about the translations files ..


locale/de_DE/LC_MESSAGES/djangojs.po
```diff
@@ -1365,6 +1292,18 @@ msgstr ""
...
+
+#~ msgid "Submit"
+#~ msgstr "Einreichen"
+
+#~ msgid "Other"
+#~ msgstr "Sonstige"
+
+#~ msgid "New answer"
+#~ msgstr "Neue Antwort"
+
```
Rineee, line 1306: Is this that bug again or are they really deleted?

fuzzylogic2000: I don't know, there were string changes, but we should also probably update a4 again. 
But let's only update the django.po files for now and we are safe and can handle everything a4 later!

fuzzylogic2000 (approved):
Perfect! 
As I have checked it out, I will remove the js.po files and then merge!

Rineee:
> Right, the test! Will also fix that!

Thank you for fixing everything! :heart:

PR #4682, opened 2022-11-14
Ks 2022 11 add moderator tasks

Rineee:
The tests are broken, but I did not succeed in fixing them and have to leave now.. it is strange, the attempt to fix is in the second commit.

Also, I only added very basic testing for moderation tasks (copied from labels), probably more to come? But wanted to push now so you can go on.


meinberlin/apps/moderationtasks/templates/meinberlin_moderationtasks/moderation_tasks_form.html
```diff
@@ -0,0 +1,26 @@
+{% load i18n %}
+
+<p class="form-hint">{% blocktrans %}Moderation tasks appear in the detailed view
```
fuzzylogic2000, line 3: detail view, not detailed view. Even though, it's probably like this in the wording file.

fuzzylogic2000: Right, and `blocktranslate` when we add new templates. Should probably change that in all templates, otherwise I also keep forgetting that.

meinberlin/apps/moderationtasks/templates/meinberlin_moderationtasks/moderation_tasks_form.html
```diff
@@ -0,0 +1,26 @@
...
+        {% include 'a4forms/includes/inline_form.html' with form=moderation_task_form can_delete=form.can_delete %}
+    {% endfor %}
+    <div class="js-form-template category-formset__template">
+        {% include 'a4forms/includes/inline_form.html' with form=form.empty_form can_delete=form.can_delete %}
+    </div>
+
+    <div class="category-formset__btn">
+        <button class="btn btn--secondary btn--small btn--full u-spacer-bottom js-add-form" type="button">
+            <i class="fa fa-plus" aria-hidden="true"></i>
+            {% trans 'Add moderation tasks' %}
```
fuzzylogic2000, line 22: `translate`

fuzzylogic2000 (approved):
Works nicely!
Only small wording things.

PR #4683, opened 2022-11-14
[issues] toggle filters btn behavior 

phillimorland:
~~@fuzzylogic2000 or @goapunk I am very confused with this, so the behavior we wish can be seen in the filter toggle for the comments, so I copied that code to make a component, planning to use it in a4 but wanted to test it first and even with exact classes it still shows being active after filters are opened, even though in comments it does not and I have wasted too much time on this by now so any ideas would be great!?~~


fuzzylogic2000:
My guess its' that it is the prevent default in the filter button click? https://github.com/liqd/adhocracy4/blob/main/adhocracy4/comments_async/static/comments_async/comment_box.jsx#L261

phillimorland:
I actually did some other fixing, the button focus is not that same as on the comments but I read that we should have btn randomly loose focus so will leave that, still think having a separate component is good though as will add to a4 next time we do stuff there

fuzzylogic2000 (approved):
Very nice! But the tests are failing

phillimorland:
when fixing tests I realised that we shouldn't show the search string when we are ordering list so fixed that

fuzzylogic2000 (approved):
Cool!

PR #4689, opened 2022-11-15
apps/votes/views: use gettext_lazy instead of gettext to fix tests

Rineee:
Without the laziness :), the translation.override does not work.. https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#lazy-translation

### DIFF
--- a/meinberlin/apps/votes/dashboard.py
+++ b/meinberlin/apps/votes/dashboard.py
@@ -4,6 +4,7 @@ from django.utils.translation import gettext_lazy as _
 from adhocracy4.dashboard import DashboardComponent
 from adhocracy4.dashboard import components
 
+from . import exports
 from . import views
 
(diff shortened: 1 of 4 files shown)


goapunk:
@Rineee tests are still failing

Rineee:
> @Rineee tests are still failing

Wtf, before they were only failing locally and now they are only failing in the ci?

Rineee:
> > @Rineee tests are still failing
> 
> Wtf, before they were only failing locally and now they are only failing in the ci?

Ok, so in the export we need to use gettext, so I moved it to its own file now  :woman_shrugging:

fuzzylogic2000 (approved):
Cool!
You could also only import one of these as `_` and use the other one just with it's name. But this is also fine!

PR #4693, opened 2022-11-16
notifications/notify_contact_on_moderator_feedback: change wording as…

Rineee:
… in notify_creator email

This belongs to story #6704, I only changed the wording in the notify_creator_on_moderator_feedback there..

I think those 2 emails are a bit confusing, maybe we should get rid of the notify_creator one (and instead add the creator to receivers if there is no contact)? We would just need a solution for the footer then.. :/

### DIFF
--- a/meinberlin/apps/notifications/templates/meinberlin_notifications/emails/notify_contact_on_moderator_feedback.de.email
+++ b/meinberlin/apps/notifications/templates/meinberlin_notifications/emails/notify_contact_on_moderator_feedback.de.email
@@ -10,7 +10,7 @@
 {% block content %}
 {{ object.module.project.organisation.name }} hat auf Ihren Beitrag reagiert.
 
-{% if object.moderator_feedback %}Ihr Beitrag {{ object.get_moderator_feedback_display }}.
+{% if object.moderator_feedback %}Der Status Ihres Beitrags hat sich geändert und steht nun auf “{{ object.get_moderator_feedback_display }}”.
 
 {% endif %}
 {% if object.moderator_statement.statement %}Offizielle Rückmeldung: {{ object.moderator_statement.statement | safe }}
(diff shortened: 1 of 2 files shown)


fuzzylogic2000 (approved):
Cool!

fuzzylogic2000:
> I think those 2 emails are a bit confusing, maybe we should get rid of the notify_creator one (and instead add the creator to receivers if there is no contact)? We would just need a solution for the footer then.. :/

We could instead add some context to the email and then use that in the template to determine which footer to show. I can't think of a variable that would be less confusing than the emails are now, though. :(

Rineee:
> > I think those 2 emails are a bit confusing, maybe we should get rid of the notify_creator one (and instead add the creator to receivers if there is no contact)? We would just need a solution for the footer then.. :/
> 
> We could instead add some context to the email and then use that in the template to determine which footer to show. I can't think of a variable that would be less confusing than the emails are now, though. :(

@fuzzylogic2000 Oh yes, we could do that! Would you be fine with that? To me it was just very confusing that there are the two e-mails that are exactly the same and sent on the same occasion.. I would probably forget again to change both next time :sweat_smile: 
We could add both send_to_contact and send_to_creator in the context and check what is true?

fuzzylogic2000:
> > > I think those 2 emails are a bit confusing, maybe we should get rid of the notify_creator one (and instead add the creator to receivers if there is no contact)? We would just need a solution for the footer then.. :/
> > 
> > 
> > We could instead add some context to the email and then use that in the template to determine which footer to show. I can't think of a variable that would be less confusing than the emails are now, though. :(
> 
> @fuzzylogic2000 Oh yes, we could do that! Would you be fine with that? To me it was just very confusing that there are the two e-mails that are exactly the same and sent on the same occasion.. I would probably forget again to change both next time sweat_smile We could add both send_to_contact and send_to_creator in the context and check what is true?

@Rineee Yes, we probably have to look at the emails anyway, because they are sent when the "rückmelden"-form is saved and we don't want to send them when we change the tasks or the remark.

PR #4696, opened 2022-11-17
Ks 2022 11 moderation tasks filter

Rineee:
@fuzzylogic2000 I already pushed this in case you need to go on with it..

The completed tasks can be added in the proposals admin. Although one can choose from any moderation task (not only the ones for the module), same for labels and categories, we should probably add a custom admin there.

I did not handle yet that only mod/initiators can view the filter.


fuzzylogic2000:
> @fuzzylogic2000 I already pushed this in case you need to go on with it..
> 
> The completed tasks can be added in the proposals admin. Although one can choose from any moderation task (not only the ones for the module), same for labels and categories, we should probably add a custom admin there.
> 
> I did not handle yet that only mod/initiators can view the filter.

Yes, thank you! I will rebase on this branch. :)

meinberlin/apps/budgeting/api.py
```diff
@@ -75,6 +77,14 @@ def list(self, request, *args, **kwargs):
                 'choices': label_choices,
             }
 
+        # moderation task filter
+        moderation_task_choices = self.get_moderation_task_choices()
```
Rineee, line 81: I first tried with adding a permission to the api, but that gets more complicated in frontend, so I thought it is easiest to simply only add the filter for moderators?

meinberlin/apps/budgeting/api.py
```diff
@@ -75,6 +78,15 @@ def list(self, request, *args, **kwargs):
                 'choices': label_choices,
             }
 
+        # moderation task filter, only show to moderators
+        if is_allowed_moderate_project(request.user, self.module):
```
Rineee, line 82: I first tried with adding a permission to the api, but that gets more complicated in frontend, so I thought it is easiest to simply only add the filter for moderators?

fuzzylogic2000: Oh yes, I think that's much more elegant than only showing it to moderators in the frontend. :)

meinberlin/apps/ideas/admin.py
```diff
@@ -18,5 +23,25 @@ class IdeaAdmin(module_admin.ItemAdmin):
     def has_add_permission(self, request, obj=None):
         return False
 
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
```
fuzzylogic2000, line 26: supercool! The whole admin thing!

fuzzylogic2000 (approved):
Supercool! 
Just one tiny thing: Can we move the task filter to the end of the filters? Or was it in the design like that?

Rineee:
> Supercool! Just one tiny thing: Can we move the task filter to the end of the filters? Or was it in the design like that?

No, its at the end in the design! In my head it was somehow connected to categories and labels, but thats a total dev perspective :sweat_smile: Changed it now!

fuzzylogic2000 (approved):
Perfect!

PR #4688, opened 2022-11-15
[6668] display mod things in proposal detail and form

fuzzylogic2000:

PR #4704, opened 2022-11-21
Kl 2022 11 moderator emails

fuzzylogic2000:
depending on #4688 which needs to be merged first (and this rebased), so only last two commits are for this PR.

second commit was because of discussion here: https://github.com/liqd/a4-meinberlin/pull/4693


fuzzylogic2000:
@Rineee It's rebased and ready for review now!

Rineee (approved):
Supercool! Thank you for merging the 2 e-mails, much better now!

PR #4749, opened 2022-12-08
deps: adding missing deps from a4

khamui:


khamui:
makes sense:
https://github.com/liqd/adhocracy4/compare/kt-2022-12-deps-as-peerdeps?expand=1

PR #4752, opened 2022-12-12
[issues/updates] small fixes still from planifolia rm

### LINKED ISSUE #4641: #6460 No color change background last/first arrow button
**URL:** (https://meinberlin-dev.liqd.net/dashboard/modules/umfrage-6/poll/)
**user:** Initiator
**expected behaviour:** While creatig a poll project, When there is only one question or when I have reached the first or last question, the non clickable moving up/down arrow buttons have a grey background
**behaviour:** The background of both buttons stays white all the time, while the non clickable arrow itself is lighter than the other
**important screensize:**
**device & browser:** 
**Comment/Question:** Same thing happens with the creating text commenting module

Screenshot?
dev:
Bildschirmfoto 2022-11-02 um 05 56 52

stage:
Bildschirmfoto 2022-11-02 um 05 56 45

phillimorland:
fixes #4641
also fixes: 

- [ ] appended button heights
- [ ] image upload preview structure
- [ ] search button no border

### DIFF
--- a/meinberlin/apps/budgeting/assets/ControlBarSearch.jsx
+++ b/meinberlin/apps/budgeting/assets/ControlBarSearch.jsx
@@ -41,17 +41,15 @@ export const ControlBarSearch = (props) => {
         onChange={handleChange}
         value={value}
       />
-      <div className="input-group-append">
-        <button
-          className="btn btn--light btn--attached-right"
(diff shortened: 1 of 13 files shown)


fuzzylogic2000 (approved):
Nice!

PR #4759, opened 2022-12-13
[#6667] swiper phases

khamui:


meinberlin/templates/a4modules/includes/module_detail_phase.html
```diff
@@ -19,25 +19,36 @@
...
-            </div>
-            <div class="phase-info__item__subtitle">
-                {% html_date phase.start_date 'DATETIME_FORMAT' %}
-                &ndash; {% html_date phase.end_date 'DATETIME_FORMAT' %}
-            </div>
-            <div class="phase-info__item__description">
-                {{ phase.description }}
-            </div>
+    <!-- Start Swiper Template -->
+    <div class="swiper-phases swiper">
```
khamui, line 23: only using `swiper-phases` caused problem, so in addition it needs `swiper` to make it work properly.

phillimorland: @khamui having spoken to @fuzzylogic2000 we do not want to load this when it is a single phase module so we should use a separate include for single phase modules and have this as an include for multi phase modules

khamui: yes makes sense, will do that

fuzzylogic2000 (changes requested):
When there are multiple modules (in timeline and parallel module view), I get none - none as phase start and end dates. And for the other ones, all phases are still shown on the same swiper tile, but that could be my cache somehow?

khamui:
> When there are multiple modules (in timeline and parallel module view), I get none - none as phase start and end dates. And for the other ones, all phases are still shown on the same swiper tile, but that could be my cache somehow?

@fuzzylogic2000 indeed both bugs were real bugs, so no caching issues. In case of only one phase I am passing `phase` now and to make swiper appear correctly for multiple phases, the styling files have to be loaded before. That should work now.

fuzzylogic2000 (approved):
Nice! It always works now!

PR #4772, opened 2022-12-15
Pm 2022 12 release issues

phillimorland:


meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -175,42 +175,42 @@ <h3>{% trans 'Responsible body' %}</h3>
...
-            <div class="offset-lg-3 col-lg-6">
-                {% block project_action %}{% endblock %}
-            </div>
-        </div>
-        {% block phase_content %}{% endblock %}
-    </section>
-
+            {% block phase_content %}{% endblock %}
+        </section>
+    </div>
```
fuzzylogic2000, line 213: Oh my, this file is so crazy. Maybe we should clean it up ba having includes for the different tab content? But another time...

fuzzylogic2000 (approved):
Very nice!

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
phillimorland: That would probably be a nice little clean up job over xmas :)

PR #4782, opened 2022-12-20
apps/budgeting: add anchor link and use in list

fuzzylogic2000:


phillimorland (commented):
the page anchor should work looking at the code but it's not and I don't know why? tested on ff and chrome
also we can add `scroll-behavior: smooth;` to html to ensure it's not stuttery

fuzzylogic2000:
> the page anchor should work looking at the code but it's not and I don't know why? tested on ff and chrome also we can add `scroll-behavior: smooth;` to html to ensure it's not stuttery

@phillimorland It did work in Chromium. Will try again and add the smooth-thingy though!

fuzzylogic2000 (commented):
@khamui Nice! For me it only works when there are multiple proposals in the list? 
Would be cool, if you could make one commit of the two.
And maybe @phillimorland can also review?

phillimorland (commented):
yep looks good, it works but the scroll to view almost scrolls too far so the top of the content is already going off screen

khamui:
@phillimorland about the offset (top edge) we decided to go with a small offset for now. In case there needs to be a positioning in the center, this would be a story for later with a design.

fuzzylogic2000:
Alright! Talked to @phillimorland and wil merge!

PR #4790, opened 2022-12-27
do not explicitly install py requirements only used in a4

fuzzylogic2000:
depending on https://github.com/liqd/adhocracy4/pull/1303
with that, we can delete the packages only usede in a4 from the requirements. I am not entirely sure that's how it should be done? It says while installing: 
`Using legacy 'setup.py install' for adhocracy4, since package 'wheel' is not installed.`

The second commit re,oved the setup.py from mB. We used to install mB in a+, but as we don't do that anymore, I think we don't need the setup.py anymore?


goapunk (approved):
looks good to me!

PR #4795, opened 2022-12-29
remove pollyfills for fetch and promise

goapunk:
* breaks compatibility with older browsers (especially IE)
* makes maplibre-gl work again

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
     "css-loader": "6.7.3",
     "datepicker": "git+https://github.com/liqd/datePicker.git",
     "dsgvo-video-embed": "git+https://github.com/liqd/dsgvo-video-embed.git",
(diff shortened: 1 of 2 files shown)


goapunk:
if we want to use maplibre-gl 2.x we will have to remove the polyfill for fetch (and promise) and therefore lose support for older browsers.

khamui:
does it mean, aplus does not have this polyfill and therefore is not supporting IE? afaik the poll module uses `fetch`.
in case we want to keep the polyfills, would it make sense to shift those into a4 then?

goapunk:
@khamui aplus doesn't use a polyfill for fetch, so we don't support old browsers there already

phillimorland (approved):
cool!

PR #4801, opened 2023-01-02
[6876] js packages

khamui:


fuzzylogic2000 (approved):
It looks good to me, but I think @phillimorland or @goapunk should have a look.

fuzzylogic2000:
@khamui This has a conflict and needs to be rebased

package.json
```diff
@@ -47,22 +44,15 @@
     "react-flip-move": "3.0.5",
     "react-markdown": "8.0.4",
     "react-router-dom": "6.6.1",
-    "react-slick": "0.29.0",
```
phillimorland, line 50: don't we need this for poll?

phillimorland (commented):
still needs a rebase and the polls are not loading, but don't think it's to do with the package as the error is `Uncaught TypeError: window.adhocracy4.getCurrentPath is not a function
`

phillimorland:
@khamui you can rebase on main now and check it

package.json
khamui: I checked it and for the results it works (I guess because a4 installs it). Or is there another location where we use it and I am not aware of?

fuzzylogic2000:
Sorry, there is a conflict again....
But it is all fine now, right? So, just merge after rebasing to not have another conflict before doing so.

PR #4805, opened 2023-01-03
[6876] remove django-capture-tag

fuzzylogic2000:
There are also emails, we don't use?! But this adds all the emails to the management command, so @CarolingerSeilchenspringer can check and we can remove them later. Already took so long as it is...

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,3 +3,4 @@ source =
     meinberlin/apps
 omit =
     */migrations/*
+    */apps/notifications/management/commands/send_test_emails.py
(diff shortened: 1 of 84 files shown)


phillimorland (approved):
looks much cleaner! :) and passes everything, only tested 1 email but looks good, i'm not sure on the test code, maybe @Rineee wants to take a quick look?

Rineee:
Yes cool! And coverage up 1.3 % :grin: Do we also want to exclude all other management commands or are they testable?

fuzzylogic2000:
> Yes cool! And coverage up 1.3 % grin Do we also want to exclude all other management commands or are they testable?

@Rineee Yeah, the others are tested or should be... Only for this it didn't make any sense, I thought.

PR #4814, opened 2023-01-05
WIP: [6800] item_detail.html: add prev and next

khamui:


fuzzylogic2000 (approved):
Nice, looks good. So, I will just go on in this branch, right?!

fuzzylogic2000:
@goapunk The prev and back buttons are already here. So, you can use these.

fuzzylogic2000:
not needed anymore. :(

PR #4818, opened 2023-01-06
[6669] apps/contrib/item-detail: restyle moderator feedback, move status dis…

fuzzylogic2000:
…play away from badges

This updates the labels, so that they are still all in the list, but the feedback one not in the detail view. It does also do the permissions and displays the box as we decided (only if there's content and the upper bit always to moderators). A proper styling is still missing, though:

- [x] style the status with icons and couloured lines
- [x] rename the moderation-info styling to sth. more general to use in both cases
- [ ] ~~think about the headline and its a11y for the actual feedback (we didn't want it twice, so what we call it in the form and so on is the headline to the new box and the organisation name and date are in the place of the headline now)~~
- [x] remove old css classes
- [x] make mobile version match design
- [x] add to component lib

I will keep this WIP and hope @khamui or @phillimorland can take over?


khamui:
@fuzzylogic2000 there is one task I am not really following, so this is probably not done yet:

-[ ] think about the headline and its a11y for the actual feedback (we didn't want it twice, so what we call it in the form and so on is the headline to the new box and the organisation name and date are in the place of the headline now)

fuzzylogic2000:
> @fuzzylogic2000 there is one task I am not following. So I kept the wording for the headlines for now.
> 
> > * [ ]  think about the headline and its a11y for the actual feedback (we didn't want it twice, so what we call it in the form and so on is the headline to the new box and the organisation name and date are in the place of the headline now)

@khamui I think, we can leave that out for now. I will add a task!

meinberlin/assets/scss/style.scss
```diff
@@ -62,7 +62,7 @@
 @import "components/lr_bar";
 @import "components/map";
 @import "components/maplist";
-@import "components/moderation_info";
+@import "components/detail_info";
```
fuzzylogic2000, line 65: They are ordered alphabetically, so this has to go up.

meinberlin/assets/scss/utility.scss
```diff
@@ -27,6 +27,10 @@
     margin-top: $em-spacer;
 }
 
+.u-spacer-top-half {
+    margin-top: ($em-spacer * 0.5);
```
fuzzylogic2000, line 31: Do we still use the em-spacer? Or should that be replaced by a rem.one?

fuzzylogic2000 (commented):
NIce, looks good! Just one nit pick.

meinberlin/assets/scss/utility.scss
khamui: good point! I just used what is currently used to avoid messing around all spacings. @phillimorland what is your take on this?

phillimorland: em-spacer is old and has not been replaced everywhere, going forward we should use spacer

fuzzylogic2000:
@khamui I guess with this not being your branch, you didn't see the answer. :)

fuzzylogic2000:
@khamui Because of all the changing back and forth, I added everything in one commit. I think, it is fine like that!

PR #4819, opened 2023-01-06
[6669] apps/moderatorfeedback: change wording

fuzzylogic2000:


meinberlin/apps/moderatorfeedback/models.py
```diff
@@ -18,11 +18,12 @@
...
-        verbose_name=_('Feedback on the contribution (public)'),
+        verbose_name=_('Feedback (public)'),
         help_text=_(
-            'The official feedback will appear below the idea, '
-            'indicating your organisation. The idea provider receives '
-            'a notification.')
+            'The feedback appears below the idea or proposal, indicating '
+            'your organisation. The editor of the contribution receives a '
+            'notification.'
+        )
```
phillimorland, line 26: I tried searching the story number but can't find it in the sprint so can't check it but the wording here in english is a little unclear but not sure if that matters as it's only in german? i wasn't really sure when 'editor' was used if we mean the contributor/owner of the contribution or some other role.

phillimorland (commented):
Looks good, just one question on the wording but I think it's good to merge

meinberlin/apps/moderatorfeedback/models.py
fuzzylogic2000: It's in the wording here: https://taiga.liqd.net/project/liqd-liquid-software/us/6669?milestone=337
But you are right! It sounds a bit weird. So question for @CarolingerSeilchenspringer : should we change that?

CarolingerSeilchenspringer: @phillimorland I meant the person who has written the idea/proposal, so in German I translate "Verfasser*in". Do we have a different wording for this already on the platform in en/dt? I can't think of any in dashboard/frontend. But if so we should stick to the existing wording to be consistent. Just thought that "user" or "participants" is too general here.

fuzzylogic2000: @CarolingerSeilchenspringer I think elsewhere we say "creator". Would that be a better wording @phillimorland ?
"The creator of the contribution receives a notification." Ok, yeah, but then it doesn't work with the contact. Maybe "the idea giver receives a notification."?

CarolingerSeilchenspringer: @fuzzylogic2000 works both for me, you can decide.

PR #4829, opened 2023-01-10
WIP: proposal detail view prev next button

goapunk:


meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
```diff
@@ -9,7 +9,7 @@ const updatedOnStr = django.gettext('updated on')
 const createdOnStr = django.gettext('created on')
 
 export const BudgetingProposalListItem = (props) => {
-  const { proposal, permissions, tokenvoteApiUrl } = props
+  const { idx, proposal, permissions, queryParams, tokenvoteApiUrl } = props
```
fuzzylogic2000, line 12: Ah! Good idea to use the index instead of searching for the idea in question in the api.
But doesn't this break with the other pagination? So, on page 2 of the list, it starts at the beginning again?

goapunk: ahh, I didn't take that into account, but I guess we can do `page * 15 + idx` ?

meinberlin/apps/budgeting/views.py
```diff
@@ -125,6 +126,31 @@ def get_back(self):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['back'] = self.get_back()
+        page = int(self.request.GET.get('id', -1))
+        # TODO: if no page/params: find page with default filter
+        if page >= 0:
+            params = self.request.GET.copy()
+            params['id'] = int(params['id']) + 1
```
goapunk, line 133: hm, this is actually bs, should be params['page'] = int(params['id') + 1

fuzzylogic2000:
not needed anymore. :(

PR #4838, opened 2023-01-12
[issues] notificatio issue and rating unifying

phillimorland:


meinberlin/apps/budgeting/assets/ListItemStats.jsx
```diff
@@ -19,19 +18,21 @@ export const ListItemStats = (props) => {
           >
             <i
               className="fa fa-chevron-up"
-              aria-label={positiveRatingsStr}
+              aria-hidden="true"
             />
-            {wrapSpaces(props.positiveCount)}
+            {props.positiveCount}
+            <span className="visually-hidden">{positiveRatingsStr}</span>
```
phillimorland, line 24: aria-label is intended for interactive elements (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) so adding a sr span instead

meinberlin/apps/livequestions/assets/Filters.jsx
```diff
@@ -73,6 +74,7 @@ export default class Filter extends React.Component {
                   checked={this.props.displayNotHiddenOnly}
                   onChange={this.props.toggledisplayNotHiddenOnly} // eslint-disable-line react/jsx-handler-names
                 />
+                <span className="visually-hidden">{displayNotHiddenText}</span>
```
phillimorland, line 77: added here as sr will read out content of label by default so seemed better then aria-label

meinberlin/apps/livequestions/assets/QuestionUser.jsx
```diff
@@ -69,18 +69,23 @@ export default class QuestionUser extends React.Component {
         <div className="live-question__action-bar">
           {this.props.hasLikingPermission
             ? (
-              <button type="button" className={(this.state.session_like ? 'u-success ' : '') + 'btn'} onClick={this.handleLike.bind(this)}>
-                <span>{this.state.likes} </span>
+              <button
+                type="button"
+                className={'rating-button rating-up ' + (this.state.session_like ? 'u-success ' : 'u-muted')}
+                onClick={this.handleLike.bind(this)}
+                aria-label={this.state.session_like ? addLikeTag : undoLikeTag}
```
phillimorland, line 76: we should maybe consider using a component for our like button as we now have it twice

meinberlin/apps/budgeting/assets/ListItemStats.jsx
fuzzylogic2000: Oh! I always thought, the icons either need aria-hidden or aria-label. So, will remember this way to do it. And there might be more places in the code where that's wrong?

meinberlin/apps/livequestions/assets/QuestionUser.jsx
fuzzylogic2000: Is this like the support button that we also use for the 3 pahse budgeting?

fuzzylogic2000 (approved):
Nice!

meinberlin/apps/budgeting/assets/ListItemStats.jsx
phillimorland: @fuzzylogic2000 what do you mean? this has an aria hidden, also if you want to have an aria-label on an element, like in a button you can add role=img

PR #4839, opened 2023-01-12
rename moderatorfeedback

khamui:
depends on https://github.com/liqd/adhocracy4/pull/1321


meinberlin/apps/ideas/templates/meinberlin_ideas/idea_moderate_form.html
```diff
@@ -27,8 +27,8 @@ <h1>{% translate 'Feedback on this idea' %}</h1>
                 {{ form.media }}
             {% endfor %}
 
-            {% include 'meinberlin_contrib/includes/form_field.html' with field=forms.moderateable.moderator_feedback %}
-            {% include 'meinberlin_contrib/includes/form_field.html' with field=forms.statement.statement %}
+            {% include 'meinberlin_contrib/includes/form_field.html' with field=forms.moderateable.moderator_status %}
+            {% include 'meinberlin_contrib/includes/form_field.html' with field=forms.moderator_feedback_text.feedback_text %}
```
fuzzylogic2000, line 31: The `moderator_feedback_text` is the form name, so it should be `feedback_text`.

fuzzylogic2000 (changes requested):
Yes, nice renaming!
There are some issues to solve, though:

- [x] fix the moderator forms by using the form name
- [x] do a PR in a4, so that we can merge that first and update here, so that the export does not break
- [x] rebase to remove the first three commits that shouldbn't be here
- [x] redo the last commit message: the modified files are unnecessarry

fuzzylogic2000 (commented):
I did the changes, so that we can also add black today.

fuzzylogic2000 (approved):

meinberlin/apps/moderatorfeedback/models.py
```diff
@@ -27,23 +27,23 @@ class ModeratorStatement(UserGeneratedContentModel):
...
         verbose_name=_('Status (public)'),
         help_text=_(
             'The status appears below the idea or proposal in red, yellow or '
             'green. The creator of the contribution receives a notification.'
         )
     )
 
-    moderator_statement = models.OneToOneField(
-        ModeratorStatement,
+    moderator_feedback_text = models.OneToOneField(
```
Rineee, line 45: To me `moderator_feedback` makes more sense here? And then the text is `moderator_feedback.feedback_text` .. why did you name it `moderator_feedback_text`?

Rineee: Ah, but that is because you wanted a new name, right? Hm, alright.. still confuses me that its not named like the model.

fuzzylogic2000: Oh yes, I thought it would be confusing as the moderator_status was formerly called moderator_feedback. And if sth. goes wrong during renaming or was forgotten, it would be worse?

Rineee:
Looking good!

PR #4859, opened 2023-01-18
[issues] Poll styling issues and dep updates

phillimorland:
when testing polls I found issues with the buttons and that led to finding some dropdowns being width of the page so it's a bit all over the place, sorry!

### DIFF
--- a/meinberlin/assets/scss/components/_tabs.scss
+++ b/meinberlin/assets/scss/components/_tabs.scss
@@ -15,16 +15,6 @@
     padding-top: 2em;
 }
 
-.tablist--mobile {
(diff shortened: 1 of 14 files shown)


meinberlin/assets/scss/components/_tabs.scss
```diff
@@ -15,16 +15,6 @@
     padding-top: 2em;
 }
 
-.tablist--mobile {
-    min-width: 100%;
```
phillimorland, line 19: didn't find a reason for it?!

khamui:
nice! working fine and looking good

PR #4862, opened 2023-01-19
[#6402] apps/budgeting: add vote count to serializer and tile and detail view

phillimorland:


phillimorland:
@Rineee It's not a lot but this is what i've done, was going to try adding the map popups but didn't start yet

meinberlin/apps/budgeting/api.py
```diff
@@ -243,12 +244,14 @@ def list(self, request, *args, **kwargs):
...
+        permissions["view_votes_left"] = has_feature_active(
+            self.module, Proposal, "vote"
         )
-        permissions["view_votes_left"] = self._has_valid_token_in_session(
-            response
-        ) and user.has_perm("meinberlin_budgeting.add_vote", self.module)
+        permissions["view_vote_count"] = self.module.module_end < timezone.now()
+
+        response = super().list(request, args, kwargs)
         response.data["permissions"] = permissions
```
Rineee, line 255: @phillimorland So I added these permissions in my PR... Probably went wrong when rebasing?

phillimorland: @Rineee err rebasing, i really have to get better at that, nice catch, will clean it up :)

meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
```diff
@@ -116,14 +116,11 @@ <h1 class="item-detail__title">{{ object.name }}</h1>
                 </div>
             </div>
         </div>
-    </article>
```
Rineee, line 119: Was the removal here on purpose? If yes, probably also need to delete opening tag?

Rineee (approved):
Yeah, supercool!

PR #4863, opened 2023-01-19
[#4664]apps/budgeting/SupportBox: Accessibility "Support-Button" (Voi…

### LINKED ISSUE #4664: Accessibility "Support-Button" (Voice-Reader)
**URL:** https://meinberlin-dev.liqd.net/projekte/module/burgerinnenhaushalt-3-phasen-21/?mode=list
**user:** any
**expected behaviour:** For accessibility the voice reader should tell the users how many people support a proposal in the detailed proposal view, the list view and the map view. In the proposal view the voice reader should tell the users, if they already clicked the support-button or not.
**behaviour:** The voice reader (any) doesn't read out loud the number of supporters in the proposal tiles nor in the detailed proposal view. In the detailed proposal view, the voice reader just reads out loud "click to support", but not how many people already support the proposal. In the detailed proposal view, the voice reader cannot tell the user whether the user has already clicked the support-button or not.
**important screensize:** any
**device & browser:** any
**Comment/Question:** It wasn't explicitly in a story and I don't know if it's important right now, just realized that this is a problem for accessibility.

Screenshot?
Bildschirmfoto 2022-11-08 um 11 41 30

Bildschirmfoto 2022-11-08 um 11 42 11

Bildschirmfoto 2022-11-08 um 11 42 30

khamui:
…ce Reader)

The corresponding A4 PR has to be merged first.

https://github.com/liqd/adhocracy4/pull/1334/commits/5b3e0281116f7fb7f02b0c8e145ed4b1377d92ee

closes #4664

### DIFF
--- a/meinberlin/apps/budgeting/assets/ListItemStats.jsx
+++ b/meinberlin/apps/budgeting/assets/ListItemStats.jsx
@@ -1,13 +1,19 @@
 import React from 'react'
 import django from 'django'
 
-const positiveRatingsStr = django.gettext('Positive Ratings')
-const negativeRatingsStr = django.gettext('Negative Ratings')
-const supportStr = django.gettext('Support')
-const commentsStr = django.gettext('Comments')
-
 export const ListItemStats = (props) => {
   const { permissions } = props
+
+  const positiveRatingsStr = django.gettext('Positive Ratings')
+  const negativeRatingsStr = django.gettext('Negative Ratings')
+  const supportStr = django.gettext('Support')
+  const commentsStr = django.gettext('Comments')
+  const supportText = django.ngettext(
+    'person supports this proposal.',
+    'persons support this proposal.',
+    props.positiveCount
+  )
+
   return (
     <div className="list-item__stats">
       {permissions.view_rate_count && (
@@ -48,7 +54,9 @@ export const ListItemStats = (props) => {
               aria-hidden="true"
             />
             {props.positiveCount}
-            <span className="visually-hidden">{supportStr}</span>
+            <span className="visually-hidden">
+              {supportText}
+            </span>
           </span>
         </span>
       )}
--- a/meinberlin/apps/budgeting/assets/SupportBox.jsx
+++ b/meinberlin/apps/budgeting/assets/SupportBox.jsx
@@ -4,14 +4,26 @@ import django from 'django'
 const api = require('adhocracy4').api
 const config = require('adhocracy4').config
 
-const translations = {
-  support: django.gettext('Click to support')
-}
-
 export const SupportBox = (props) => {
   const [support, setSupport] = useState(props.support)
   const [userSupported, setUserSupported] = useState(props.userSupported)
   const [userSupportId, setUserSupportId] = useState(props.userSupportId)
+
+  const translations = {
+    support: django.ngettext(
+      'person supports this proposal.',
+      'persons support this proposal.',
+      props.support
+    ),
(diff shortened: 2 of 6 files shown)


goapunk (approved):
Looks good to me but will leave for @phillimorland as I my a11y knowledge is still quite limited

meinberlin/apps/budgeting/assets/SupportBox.jsx
```diff
@@ -66,7 +78,7 @@ export const SupportBox = (props) => {
   return (
     <div className="rating">
       <button
-        aria-label={translations.support}
+        aria-label={support + ' ' + translations.support + getSupportClickStr()}
```
phillimorland, line 81: so i'm a little torn here, I'm not sure if it's better to use a visually hidden span here instead of aria label? either way as it is this will read the count twice so we should try to not confuse the sr user so we should ensure it's read correctly

phillimorland (approved):
some comments

meinberlin/apps/budgeting/assets/ListItemStats.jsx
```diff
@@ -47,8 +53,10 @@ export const ListItemStats = (props) => {
               className="far fa-thumbs-up"
               aria-hidden="true"
             />
-            {props.positiveCount}
-            <span className="visually-hidden">{supportStr}</span>
+            <span aria-hidden="true">{props.positiveCount}</span>
+            <span className="visually-hidden">
+              {props.positiveCount + ' ' + supportText}
```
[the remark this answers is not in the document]
khamui: this way it reads the number and stops. then you have to navigate to the next item then it will read out the sentence (without the number). At least in Voiceover.

meinberlin/apps/budgeting/assets/SupportBox.jsx
khamui: i See, in Voiceover it does not read out the number twice. But added so it will definitely skip the second number to be read out.

phillimorland: I just added a comment in a4 about aria-hidden on spans but have not tested any of these as still can't install so maybe see if its working in Narrator, JAWs and NVD as they are the most used SRs, if it's fine we can leave it as is but if not then we should re-asses aria-hidden on non-semantic elements

phillimorland:
also just checking dev, the info is read before the idea title, while we are dealing with a11y we should ensure title then info is read

meinberlin/apps/budgeting/assets/ListItemStats.jsx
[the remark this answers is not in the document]
phillimorland: sorry there should have been no span around the positiveCount

meinberlin/apps/budgeting/assets/SupportBox.jsx
khamui: @phillimorland good point as I never tried another SR before. I did now and for JAWs and NVDA it works the set aria-hidden on a <kbd><span></kbd>.

Also on MDN they use a `span` as an example (even if they use the span for an icon):
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden#example

khamui:
Okay tested on Voiceover, JAWs and NVDA @phillimorland

phillimorland (approved):
awesome! added the hash, will merge once database sorted

PR #4865, opened 2023-01-20
Ks 2023 01 filter issues

Rineee:
Different issues I tried to fix, see commit messages.. the 2nd is solving that invalid filter choices throw an error (now they are just ignored) and adds default ordering which actually wasnt working before.. I also switched from many different filter backends for the model fields to one filterset that gathers all the model field filters.

### DIFF
--- a/meinberlin/apps/moderationtasks/filters.py
+++ b/meinberlin/apps/moderationtasks/filters.py
@@ -1,25 +1,27 @@
-from rest_framework.filters import BaseFilterBackend
+from django_filters import rest_framework as rest_filters
 
 from meinberlin.apps.moderationtasks.models import ModerationTask
 
 
-class ModerationTaskFilterBackend(BaseFilterBackend):
-    """Filter out proposals that have the moderation tasks completed."""
+class OpenTaskFilter(rest_filters.ModelChoiceFilter):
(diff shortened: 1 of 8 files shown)


meinberlin/apps/moderationtasks/filters.py
```diff
@@ -1,25 +1,21 @@
...
-                    "{app_label}_{model}_completed".format(
-                        app_label=queryset.model._meta.app_label,
-                        model=queryset.model.__name__.lower(),
-                    ),
-                ).all()
-                return queryset.exclude(id__in=proposals_completed)
-            except ModerationTask.DoesNotExist:
-                pass
+class OpenTaskFilter(rest_filters.ModelChoiceFilter):
+    def filter(self, queryset, value):
```
fuzzylogic2000, line 7: I am a bit sad that the whole doc string is gone. Maybe we can change it to sth like "Filter out propsals with the field associated to it. Currently used to filter out proposals that have the moderation tasks completed."

fuzzylogic2000 (commented):
Looks good! I still have to and will try it out, though.

fuzzylogic2000 (approved):
Yes, nice! It also works.

PR #4866, opened 2023-01-23
[6963] VotingToken: use and store hashes instead of codes

goapunk:


goapunk:
warning: has migrations and breaks db

meinberlin/apps/votes/models.py
```diff
@@ -67,6 +72,44 @@ def project(self):
...
         return item.module == self.module
 
+    @staticmethod
+    def get_hash_for_token(token, module):
+        params = argon2.profiles.RFC_9106_LOW_MEMORY
+        # use module_slug as salt as we can't store per-token salts
+        # TODO: slugs can belong, check max length of salt?
+        token_hash = argon2.low_level.hash_secret(
+            secret=token.encode(),
+            salt=module.slug.encode(),
```
goapunk, line 82: this uses the module slug to add at least a bit more security, however, if the slug is changed it wouldn't work anymore. It's not editable from the djang-admin afaics, so it should only happen if we ever do it manually?

meinberlin/apps/votes/templates/meinberlin_votes/token_export_dashboard.html
```diff
@@ -5,7 +5,8 @@
     <h1 class="u-first-heading">{% translate 'Download voting codes' %}</h1>
 
     <p class="form-hint u-bottom-divider">{% blocktranslate %}For voting in the third phase, all participants need their own voting code. Here you can download the codes in packages of {{ export_size }} per excel file. To generate the codes, please contact {{ contact_email }}.{% endblocktranslate %}</p>
-    <p class="u-bottom-divider">{% blocktranslate %}Number of generated codes: {{ number_of_module_tokens }}.{% endblocktranslate %}</p>
+    <p class="form-hint u-bottom-divider">{% blocktranslate %}Note that codes can only be downloaded once and then will be deleted from our database!{% endblocktranslate %}</p>
```
goapunk, line 8: proper wording & design still needed

meinberlin/apps/votes/models.py
```diff
@@ -29,23 +32,25 @@ class VotingToken(models.Model):
...
-                return
-            except IntegrityError as e:
-                if "UNIQUE constraint failed" in e.args[0]:
-                    if i == 3:
-                        raise
-                    else:
-                        self.token = get_token()
-                else:
-                    raise
+        # if self._state == 'adding':
```
goapunk, line 38: do we still need this? codes are only generated using the batch_create, which doesn't call save(), right?

meinberlin/apps/votes/models.py
```diff
@@ -67,6 +71,48 @@ def project(self):
     def is_valid_for_item(self, item):
         return item.module == self.module
 
+    @staticmethod
+    def get_hash_for_token(token, module):
+        params = argon2.profiles.RFC_9106_LOW_MEMORY
+        salt = module.slug
```
goapunk, line 77: this uses the module slug as salt for a little bit of extra security, however this assumes that the slug is never changed. Is that ok? otherwise we would need a hashing algorithm without salt like sha512 (which is less suitable for storing confidential things)

meinberlin/apps/votes/models.py
```diff
@@ -67,6 +71,48 @@ def project(self):
     def is_valid_for_item(self, item):
         return item.module == self.module
 
+    @staticmethod
+    def get_hash_for_token(token, module):
```
fuzzylogic2000, line 112: The naming is already very on point and understandable and total nitpick, but to me the `get` is for stuff that already exists, so this could be `create_hash_for_token` or just `hash_token`?

meinberlin/apps/votes/templates/meinberlin_votes/token_export_dashboard.html
fuzzylogic2000: Maybe you could already add a screenshot to the story to find out where we should put the extra info?

meinberlin/apps/votes/views.py
```diff
@@ -29,10 +29,11 @@ class ExportTokenDashboardView(
     template_name = "meinberlin_votes/token_export_dashboard.html"
 
     def _get_number_of_tokens(self):
-        return VotingToken.objects.filter(module=self.module, is_active=True).count()
+        tokens = VotingToken.objects.filter(module=self.module, is_active=True)
```
fuzzylogic2000, line 32: Maybe a docstring """Return total and downloadable token count for module.""" or sth. like that? I mean, it is quite obvious now, but might not be next time we look at it?

meinberlin/apps/votes/migrations/0003_voting_token_add_token_hash.py
```diff
@@ -1,52 +1,51 @@
 # Generated by Django 3.2.16 on 2023-01-24 11:01
-
+import argon2 as argon2
```
fuzzylogic2000: why not just `import argon2`?

meinberlin/apps/votes/templates/meinberlin_votes/token_generation_dashboard.html
```diff
@@ -4,7 +4,7 @@
 {% block dashboard_project_content %}
 <h1 class="u-first-heading">{% translate 'Generate voting codes' %}</h1>
 
-<p class="form-hint u-bottom-divider">{% blocktranslate %}For voting in the third phase, all participants need their own voting code. Here you can generate up to {{ tokens_per_module }} codes. Depending on the number this may take a few minutes.{% endblocktranslate %}</p>
+<p class="form-hint u-bottom-divider">{% blocktranslate %}For voting in the third phase, all participants need their own voting code. Here you can generate up to {{ tokens_per_module }} codes. Depending on the number this may take up to a day.{% endblocktranslate %}</p>
```
fuzzylogic2000: We could also say "this may take quite a few minutes". :rofl:

meinberlin/apps/votes/models.py
```diff
@@ -110,19 +110,8 @@ def is_valid_for_item(self, item):
...
-            secret=token.encode(),
-            salt=salt.encode(),
-            time_cost=params.time_cost,
-            memory_cost=params.memory_cost,
-            parallelism=params.parallelism,
-            hash_len=params.hash_len,
-            type=params.type,
-            version=params.version,
-        ).decode()
+        token_hash = hashlib.sha3_512((salt + token).encode()).hexdigest()
```
fuzzylogic2000, line 114: Is that what we already use for the passwords?

meinberlin/apps/votes/migrations/0003_voting_token_add_token_hash.py
goapunk: pycharm did that! :D also I removed argon and moved to sha3 which is much faster and should be fine since the tokens are already high-entropy

meinberlin/apps/votes/templates/meinberlin_votes/token_generation_dashboard.html
goapunk: forgot to change that back to minutes with sha3 :D

meinberlin/apps/votes/models.py
goapunk: for passwords we use PBKFD2 which uses sha2 I think, but hashlib is part of python which is nice , in contrast to the argon2 implementation which has not many stars on github

meinberlin/apps/votes/admin.py
```diff
@@ -14,6 +14,10 @@ class VotingTokenAdmin(admin.ModelAdmin):
     def module_name(self, token):
         return token.module.name
 
+    def save_model(self, request, obj, form, change):
```
goapunk, line 17: tokens created via the admin will get a new package number (i.e. the excel file will only contain one code), I don't see any other good way to do this. Appending it to an existing package is rather confusing? (especially when it has been downloaded already).

goapunk:
ok, this is ready for another round of reviewing

meinberlin/apps/budgeting/templatetags/react_proposals_vote.py
```diff
@@ -29,17 +29,16 @@ def react_proposals_vote(context, module, proposal):
...
-                session_token_voted = TokenVote.objects.filter(
-                    token__pk=token.pk, content_type=proposal_ct, object_pk=proposal.pk
-                ).exists()
-            except VotingToken.DoesNotExist:
-                pass
+            token = VotingToken.get_voting_token_by_hash(
+                token_hash=request.session["voting_tokens"][module_key], module=module
+            )
+            if not token:
+                return
```
Rineee, line 36: what does this return do?

goapunk: I can also flip it to `if token:` if that's better to read

Rineee: yeah maybe, then its also more consistent (like in VotingTokenInfo).. thanks!

meinberlin/apps/votes/admin.py
```diff
@@ -6,14 +6,28 @@
...
 @admin.register(VotingToken)
 class VotingTokenAdmin(admin.ModelAdmin):
-    fields = ("token", "module", "is_active", "allowed_votes")
-    readonly_fields = ("token",)
+    fields = (
+        "token",
+        "token_hash",
+        "module",
+        "is_active",
+        "package_number",
```
Rineee, line 14: maybe also add package_number to read_only fields? if someone changes that manually, that breaks stuff, no?

Rineee (approved):
Yeah, cool, it is all working fine!

PR #4872, opened 2023-01-25
Ks 2023 01 several tokens in session

Rineee:


meinberlin/apps/budgeting/serializers.py
```diff
@@ -85,23 +85,27 @@ def get_session_token_voted(self, proposal):
...
+                            module=module,
+                        )
+                    except VotingToken.DoesNotExist:
+                        pass
 
                 vote = TokenVote.objects.filter(
                     token=token,
                     content_type=ContentType.objects.get_for_model(proposal.__class__),
                     object_pk=proposal.pk,
                 )
```
fuzzylogic2000, line 107: This block can be indented one level more, right? If the module is not in the voting_tokens, then we will also not find a vote anyway.

meinberlin/apps/votes/filters.py
```diff
@@ -9,25 +9,23 @@ class OwnVotesFilterBackend(BaseFilterBackend):
...
 
                     if token:
-                        own_votes = TokenVote.objects.filter(
-                            token=token
-                        )
-                        return queryset.filter(
-                            id__in=own_votes.values("object_pk")
-                        )
+                        own_votes = TokenVote.objects.filter(token=token)
+                        return queryset.filter(id__in=own_votes.values("object_pk"))
```
fuzzylogic2000, line 29: This if can also be indented one more

fuzzylogic2000 (approved):
Nice! Just two nit-picks

meinberlin/apps/budgeting/serializers.py
Rineee: Ah yes, true!

fuzzylogic2000 (approved):
Nice!

PR #4874, opened 2023-01-26
README: update installation instruction

goapunk:


goapunk:
we don't want anyone to install fixtures on a prod server :face_with_spiral_eyes:

PR #4878, opened 2023-01-27
WIP: [6951] apps/budgeting: end session on button click

fuzzylogic2000:
Oh, this feels very unrefined and clunky. For the logout, we do a post, so maybe we should also do a post instead of the get here after all? And maybe some different permissions? (@Rineee @goapunk) 
And in this version the button works, but you can only see that after reloading the page. I think, there was more to fix also when the session ends, so that the form field to fill in returns? Buuut we currently need a reload of the whole page for that anyway. :cry: The React is also of course missing the modal inbetween the first click and ending the session and doesn't look like it should (link). (@khamui @phillimorland )


khamui:
link comes with modal in this PR:
https://github.com/liqd/a4-meinberlin/pull/4871/commits/847fd7688dd04ff83dc64f126007b1682f35baa5

meinberlin/apps/budgeting/assets/BudgetingProposalList.jsx
```diff
@@ -16,6 +16,7 @@ export const BudgetingProposalList = (props) => {
     return django.interpolate(countText, [votes])
   }
   const noResults = django.gettext('Nothing to show')
+  const endSessionString = django.gettext('Are you ready? End session')
```
phillimorland, line 19: this should be are you finished right? are you ready doesn't really make sense in the context

phillimorland:
There is no way to load the django token form other then a page reload which I have added. 
However the session ending does not work in firefox, only in chrome, will take a look tomorrow but @goapunk or @Rineee i'm not sure on the get or post stuff if either of you have an opinion i can implement while trying to fix the firefox caching.

Also due to the url changing when adding a token on ending the session the page returns to map but this is also the behavior if you refresh the page so not sure we care?!

goapunk:
I'm fine with `get`, the session ending works for me in firefox, maybe there's a difference between mac and linux?

phillimorland:
@goapunk im getting these errors:
```
Cookie “sessionid” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
```
`Cookie “sessionid” has been rejected because it is already expired.`

phillimorland:
@goapunk oh, was really hoping it was that error :(

khamui:
@phillimorland tested in FF 108.0.1 and it does not work. 
In chrome (109) it does work

phillimorland:
@khamui dam was really hoping was just me :p ok will take a look, cheers! :)

phillimorland:
@goapunk no I think this should be merged then @khamui can rebase and add modal stuff

PR #4879, opened 2023-01-27
[6958] apps/budgeting: display map or list as back link text depending on mo…

fuzzylogic2000:
…de to get back to


khamui:
very nice!

PR #4884, opened 2023-01-30
[4810]apps/budgeting/controlbar: fix horizontal spacing of filter dro…

### LINKED ISSUE #4810: #6799 [mB] pkg 3 (navigation): back-button jumps to filtered proposal list
**URL:** https://meinberlin-dev.liqd.net/projekte/burgerinnenhaushalt-3-phasen/?mode=list&moderator_feedback=CONSIDERATION
**important screensize:** *Desktop*
**device & browser:** *Safari Version 16.1 (18614.2.9.1.12) and Chrome Version 108.0.5359.94 (Official Build) (x86_64)*
**Comment/Question:**wow, this go back thing was a life changing thing. I doesn't feel so lost when I come back to the list haha Thanks!

The bottom part of the dropdown seems a bit off.
Screenshot 2023-01-04 at 16 44 23

Im never fully happy with all these Dropdowns when we have too many of them, but I guess that's for now what we use. Would it make sense to reduce the spacing of the filters only a bit? So we can have a clearer division. 
Screenshot 2023-01-04 at 16 50 23

This part of the Module seems also a bit off.
Screenshot 2023-01-04 at 16 02 03

**

khamui:
…pdowns

closes #4810

### DIFF
--- a/meinberlin/assets/scss/components/_control-bar.scss
+++ b/meinberlin/assets/scss/components/_control-bar.scss
@@ -12,7 +12,7 @@
         & > button,
         & > div.btn-group,
         & > form {
-            margin-bottom: $spacer;
+            @extend .u-spacer-bottom-half;
         }
     }
 


goapunk (approved):
cool, looks much better!

PR #4888, opened 2023-01-31
budgeting: prevent auto-filling of voting token

goapunk:


goapunk:
tested on chrome and firefox, not safari

phillimorland (approved):
LGTM

PR #4901, opened 2023-02-02
Jd 2023 01 store codes as hash

goapunk:


phillimorland (approved):
LGTM second attempt :D

PR #4902, opened 2023-02-02
Jd 2023 01 rename accordion

goapunk:


webpack.common.js
```diff
@@ -86,6 +86,10 @@ module.exports = {
     swiper_phases: {
       import: './meinberlin/assets/js/swiper_phases.js'
     },
+    init_dashboard_accordion: {
```
goapunk, line 89: separate entry point as done in aplus

meinberlin/templates/a4dashboard/base_dashboard_project.html
```diff
@@ -61,4 +61,5 @@
 
 {% block extra_js %}
     <script src="{% static 'unload_warning.js' %}"></script>
+    <script src="{% static 'init_dashboard_accordion.js' %}"></script>
```
goapunk, line 64: this should cover all the places it's used I think?

goapunk:
this will break the existing cookie but that's ok I guess?

phillimorland (approved):
very nice!

Issue #4900, opened 2023-02-02
Breadcrumb in modul "Priorisierung mit Karte" is saying always "list" 


jostertag:
**URL:** https://meinberlin-dev.liqd.net/maptopicprio/2022-00910/
**user:** any
**expected behaviour:** When coming from map I don't want to get confused by a button which says "list" even though you going back to the map.
**behaviour:** In modul "Priorisierung mit Karte" in the detail view the breadcrumb says always list. When you for example come from the map, you go back to map but breadcrumb says "list".
**important screensize:** -
**device & browser:** mac ff
**Comment/Question:** in the other moduls with ideas/proposals (except Bürger*innenhaushalt), it's saying "Back" ("Zurück") and neither "list" nor "map". Probably related to this issue: https://github.com/liqd/a4-meinberlin/issues/4816

Screenshot?

Rineee:
@jostertag That should be solved with [this story](https://taiga.liqd.net/project/liqd-liquid-software/us/6958?milestone=340), right? Please reopen, if not!

Rineee:
Oh, this is the priorisierung, that should not happen..

Rineee:
@fuzzylogic2000 it was changed to list [here](https://github.com/liqd/a4-meinberlin/commit/5502c3e2b3227decc9e5035b86a99e6f04cf9bec), do you remember what was the idea there?

fuzzylogic2000:
@Rineee Yes, it was in the first story about the back-link to change the wording to list, so I changed it everywhere. And then, I didn't change it back in the second story. So, still needs to be changed back.

fuzzylogic2000:
But maybe it would be good to have decision about the back/map/list wording for the default for the budgeting first. That is currently also "back" when coming from outside directly to an idea. Should that change @CarolingerSeilchenspringer ?

Also, I just saw that changing back for the item detail, I used "back" instead of the "go back" that was there before. I guess it doesn't matter shich one we use, we should just stick to the same for the item and the topic deatil views.

PR #4906, opened 2023-02-03
budgeting/templates/proposal_list: keep url params when token is ente…

### LINKED ISSUE #4897: Filters disapear after code entry and filter selection
**URL:** https://meinberlin-dev.liqd.net/projekte/burgehaus-testing/?own_votes=true
**user:** user with token
**expected behaviour:** filters remain visible
**behaviour:** after a user submits a voting token and uses a filter option that will show no results, all the react filters disappear and can not be viewed again unless user goes to map and back 

**device & browser:** firefox and safari so far but possibly others

Screenshot?

Rineee:
…red, fixes #4897


phillimorland:
also seems to fix https://github.com/liqd/a4-meinberlin/issues/4900

phillimorland (approved):
sweet!

PR #4910, opened 2023-02-06
[6958] apps/budgeting: display "map" as default back-link text, apps/proriza…

### LINKED ISSUE #4900: Breadcrumb in modul "Priorisierung mit Karte" is saying always "list"

fuzzylogic2000:
…tions: display "back" as back-link text - fixes #4900

For budgeting always display "map" when not coming from list view

### DIFF
--- a/meinberlin/apps/budgeting/views.py
+++ b/meinberlin/apps/budgeting/views.py
@@ -140,21 +140,22 @@ class ProposalDetailView(idea_views.AbstractIdeaDetailView):
         are remembered when going back, we check if the referer is a
         module or project detail view and add the appropriate back url.
         """
+        back_link = self.module.get_detail_url
+        back_string = _("map")
         if "Referer" in self.request.headers:
             referer = self.request.headers["Referer"]
             parsed_url = urlparse(referer)
             match = resolve(parsed_url.path)
             if match.url_name == "project-detail" or match.url_name == "module-detail":
                 back_mode = None
-                back_string = _("map")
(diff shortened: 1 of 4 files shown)


khamui (commented):
nice! lgtm..
gh actions is failing, apparently a pytest.

question: is there a reason why we do not apply the "list/map/back" as we have it for budgeting?

fuzzylogic2000:
> nice! lgtm.. just actions is failing, apparently because of a pytest.
Oh, right, I did also write tests for the backlink and didn't consider these. :blush: Will fix!

> 
> question: is there a reason why we do not apply the "list/map/back" as we have it for budgeting?
The view for the topics doesn't have that, so we cannot use it in the template.

khamui:
very nice!

PR #4916, opened 2023-02-06
[4907]apps/votes: helptext and package size set to 1 mio, instead of …

### LINKED ISSUE #4907: #6963 Wrong Number in Wording of Helptext
**URL:** https://meinberlin-dev.liqd.net/dashboard/modules/burgerinnenhaushalt-3-phasen-21/download-codes/
**user:** admin, initiator
**expected behaviour:** Each code-package should contain a max. of 1.000.000 codes. The wording of the helptext should have also the right number of 1.000.000 codes per package ~~as each package should contain a maximum of 1.000.000 codes per excel-file.~~
**behaviour:** the number in the wording of the helptext is "10.000.000" ~~and the packages can contain more than 1.000.000 codes. ~~
**important screensize:** -
**device & browser:** mac ff
**Comment/Question:** I tried it with generating two mill codes and the codes were put in only one code-package. I also couldn't download the package probably because it was too big.

Package size issue: https://github.com/liqd/a4-meinberlin/issues/4911

Bildschirm­foto 2023-02-03 um 17 21 37

Screenshot?

### LINKED ISSUE #4911: #6963 Too many codes in 1 package
URL: https://meinberlin-dev.liqd.net/dashboard/modules/burgerinnenhaushalt-3-phasen-21/download-codes/
user: admin, initiator
expected behaviour: Each code-package should contain a max. of 1.000.000 codes. ~~The wording of the helptext should have also the right number of 1.000.000 codes per package as each package should contain a maximum of 1.000.000 codes per excel-file.~~
behaviour: ~~the number in the wording of the helptext is "10.000.000" and~~ the packages can contain more than 1.000.000 codes.
important screensize: -
device & browser: mac ff
Comment/Question: I tried it with generating two mill codes and the codes were put in only one code-package. I also couldn't download the package probably because it was too big.

Linked: https://github.com/liqd/a4-meinberlin/issues/4907

khamui:
…10 mio

fixes #4907 
fixes #4911

### DIFF
--- a/meinberlin/apps/votes/tasks.py
+++ b/meinberlin/apps/votes/tasks.py
@@ -5,9 +5,9 @@ from meinberlin.apps.votes.models import VotingToken
 from meinberlin.apps.votes.models import get_token_12
 
 # Number of tokens to insert into database per bulk_create
-BATCH_SIZE = 1000000
+BATCH_SIZE = 100000
 # Max number of tokens in one download / package
-PACKAGE_SIZE = 10000000
+PACKAGE_SIZE = 1000000
 
 
 def generate_voting_tokens(module_id, number_of_tokens, existing_tokens):


meinberlin/apps/votes/tasks.py
```diff
@@ -7,7 +7,7 @@
 # Number of tokens to insert into database per bulk_create
 BATCH_SIZE = 1000000
```
Rineee, line 8: I believe also this has one zero too many, wasnt that 100.000 before @fuzzylogic2000 ?

fuzzylogic2000: No, it used to be one million: the PAGE_SIZE in the exports: https://github.com/liqd/a4-meinberlin/commit/eacd22c795353a6a5a0ac9773f74dba6ae533bc5#diff-ef2af1f23503a2f006e8048ad40751390211bb74ca18c6882369e290b66f022bL10
And yeah, we could use one variable as they are the same size, but on the other hand, they are for different things, so if we ever want to change only one (like if we change the background tasks somehow) it does make sense to keep them seperated. So, I think, we should keep two numbers.

fuzzylogic2000: @Rineee

fuzzylogic2000: Right, @Rineee was right! The BATCH_SIZE is the other number... That is also wrong and should be one zero shorter.

Rineee (approved):
Yes!

PR #4924, opened 2023-02-07
WIP: [6962] apps/budgeting: shorten expiry age for session when token is entered,…

fuzzylogic2000:
… but only when user didn't log in with remember me

When 'remember' is checked during login, allauth sets the expiry to the set expiry age, while otherwise it is set to 0, causing Django to end the session once the browser is closed. https://github.com/pennersr/django-allauth/blob/30a3473b8010264143f031bb7565eb26faa143e4/allauth/account/forms.py#L206 https://docs.djangoproject.com/en/4.1/topics/http/sessions/#using-sessions-in-views

This longer explanation is also in the commit message. Do you think, we will find it there?

~~Should we merge before I add the tests, so that it can already be tested? edit: well, no, writing tests, I have to investigate if it only works for logged in users if we do it like that. :roll_eyes:~~ Tests are there, should be all fine now. Could you review @Rineee ?

### DIFF
--- a/meinberlin/apps/budgeting/views.py
+++ b/meinberlin/apps/budgeting/views.py
@@ -21,6 +21,8 @@ from meinberlin.apps.votes.models import VotingToken
 from . import forms
 from . import models
 
+SESSION_EXPIRY_AGE = 10 * 60  # 10 minutes
+
 
 def get_ordering_choices(view):
     choices = (("-created", _("Most recent")),)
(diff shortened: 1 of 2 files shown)


Rineee:
@fuzzylogic2000 Really cool!
But I have a question for the following case: When 2 different tokens are entered, it now does not renew the expiry age when 2nd token is entered, but still will end the session after the 12 hours counting from when 1st token was entered.. do we maybe want to reset the expiry age there? Not sure, and also this is of course an edge case..

fuzzylogic2000:
> @fuzzylogic2000 Really cool! But I have a question for the following case: When 2 different tokens are entered, it now does not renew the expiry age when 2nd token is entered, but still will end the session after the 12 hours counting from when 1st token was entered.. do we maybe want to reset the expiry age there? Not sure, and also this is of course an edge case..

Yes! Good catch! That would be especially annoying if you came back after like 11 hours and 45 minutes for the other round of voting. Argh, actually we need to talk about that. It's also not great if I come back after a few hours and then the session is ended even though I vote or read ideas. I mean for reading it would be hard to reset, but for voting, we probably should do that. @Rineee

fuzzylogic2000:
closing in favour of #4966

Issue #4717, opened 2022-11-23
think about moderation rules


fuzzylogic2000:
**URL:** 
**user:** 
**expected behaviour:** ?? To be discussed
**behaviour:** admins and group members are not allowed to moderate (see tests https://github.com/liqd/a4-meinberlin/pull/4707/)
**important screensize:**
**device & browser:** 
**Comment/Question:** This will normally not be noticed during testing because when the project is created, the creator will be added as moderator. And I didn't check where we do use these rules. Still, it was a bit unexpected while writing the tests and we should think about it.

Screenshot?

CarolingerSeilchenspringer:
@fuzzylogic2000 does anything speak against group members and admins to have moderator rights? I would not think so, from my point of view we can change it.

fuzzylogic2000:
For group members I agree as they are basically initiators for certain projects and should have the same right as the initiators on these projects. 
For admins I am not sure and would have to check where we actually use these permissions and what that would mean. And as that is not urgent, I would go back to that in the next mB sprint.

fuzzylogic2000:
https://github.com/liqd/a4-meinberlin/pull/4926#discussion_r1100489548

PR #4926, opened 2023-02-07
tests/account//votes: testing story #6951, test that tokens are delet…

Rineee:
…ed when session ends

### DIFF
--- a/tests/budgeting/rules/test_rules_view_vote_count.py
+++ b/tests/budgeting/rules/test_rules_view_vote_count.py
@@ -7,12 +7,13 @@ from freezegun import freeze_time
 from adhocracy4.projects.enums import Access
 from adhocracy4.test.helpers import freeze_phase
 from adhocracy4.test.helpers import freeze_post_phase
+from adhocracy4.test.helpers import freeze_pre_phase
 from adhocracy4.test.helpers import setup_phase
 from adhocracy4.test.helpers import setup_users
 from meinberlin.apps.budgeting import phases
 from meinberlin.test.helpers import setup_group_members
 
-perm_name = "meinberlin_budgeting.view_support"
+perm_name = "meinberlin_budgeting.view_vote_count"
 
 
 def test_perm_exists():
@@ -20,10 +21,10 @@ def test_perm_exists():
 
 
 @pytest.mark.django_db
-def test_module_no_support_phase(
+def test_module_no_voting_phase(
     phase_factory, proposal_factory, user, admin, user_factory, group_factory
 ):
-    phase, module, project, item = setup_phase(
+    collect_phase, module, project, item = setup_phase(
         phase_factory, proposal_factory, phases.CollectPhase
     )
     anonymous, moderator, initiator = setup_users(project)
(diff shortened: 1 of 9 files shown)


tests/budgeting/rules/test_rules_view_support.py
```diff
@@ -268,7 +268,7 @@ def test_phase_active_project_draft(
 
 
 @pytest.mark.django_db
-def test_post_phase_project_archived(
+def test_post_voting_phase_project_archived(
```
Rineee, line 271: @fuzzylogic2000 I realized that we do not consider is_archived anywhere in the rule, so this tests actually is the same as test_voting_phase_active.. same for testing rule view_vote_count, also there is_archived isnt considered... do you think we nee to add it somewhere to the rule or does this test maybe just not make sense?

tests/budgeting/rules/test_rules_view_vote_count.py
```diff
@@ -0,0 +1,335 @@
...
+    participant = user_factory()
+    project.participants.add(participant)
+
+    assert project.access == Access.PRIVATE
+    with freeze_post_phase(voting_phase):
+        assert not rules.has_perm(perm_name, anonymous, module)
+        assert not rules.has_perm(perm_name, user, module)
+        assert not rules.has_perm(perm_name, group_member_out, module)
+        assert not rules.has_perm(perm_name, group_member_in_org, module)
+        assert not rules.has_perm(perm_name, group_member_in_pro, module)
```
Rineee, line 237: Not sure about this one.. this is false, because we replaced the `is_allowed_moderate_project` by `is_superuser` here. Same for initiators below.. I guess both of them should see the vote count in a private project though, because they are allowed to see the private project, right? @fuzzylogic2000 But not sure how to change that (in a consistent way).. :exploding_head:

tests/budgeting/rules/test_rules_view_support.py
fuzzylogic2000: Haaa, don't know. We do consider the archived for things like vote or support, so that you cannot do that when the proposal is archived, right? Not showing sth. anymore when the porposal was archived does seem a bit intransparent? So, I think: let's keep it like that for now and adapt if needed.
Oh, but the test still makes sense, so that we notice when we do change that and only do that on purpose. I also like, that you can see in the test that it is really meant like that and we didn't just forget about it.

tests/budgeting/rules/test_rules_view_vote_count.py
fuzzylogic2000: Ohhhhh, very right. Because we use that for determining if we want to show the moderation stuff, right? Ah. Should we look at that together?

fuzzylogic2000 (approved):
It's great! Do you want to have a look at the private project thing or should we just merge?

tests/budgeting/rules/test_rules_view_vote_count.py
fuzzylogic2000: I added that comment to an issue we have about that: https://github.com/liqd/a4-meinberlin/issues/4717
It's also in the prod release milestone, so we sill have to come back to it. :) So, with that, we are good to merge! But I am not sure, I am meant to? I wasn't assigned a s rewiewer, but found this through your comments. So, feel free to merge if it is ready!

Rineee: @fuzzylogic2000 I wanted to add some more tests, so lets wait for that? Or do you want to have it in the release?

fuzzylogic2000: Cool! No, waiting is fine!

fuzzylogic2000 (approved):
Perfect!

PR #4927, opened 2023-02-07
[#7008 #6402 tests] list view fixes and tests

phillimorland:


khamui:
> @khamui I have kinda doubled up on tests here but I wanted to kind of use them like documentation but not sure if it's better to do that in item or list view or we should just leave it as is but then we have more tests to maintain?! what do you think?

So from checking the code it looks good to me and it also works fine!
I am also unsure if we should test it twice. I think `BudgetingProposalList` is the right spot to test this.

`BudgetingProposalListItem` is good covered already so, probably we could remove the redundant part here.


Bildschirm­foto 2023-02-07 um 18 20 05

khamui:
looking good! nice tests!

PR #4929, opened 2023-02-07
[6669]apps/budgeting/moderator-feedback: detail-info title fontsize s…

khamui:
…maller

we tried the option to use h1 (with "original" font-size) and h2 for the boxes' titles. But they appear very big. So we decided to overwrite this h2 font size.

fixes #4818


phillimorland (approved):
cool!

PR #4936, opened 2023-02-08
apps/votes: increase token batch sizes again, batch update the tokens…

fuzzylogic2000:
… in download


Rineee (approved):
Supercool, working great now!

PR #4937, opened 2023-02-08
apps/ideas/views: add all missing annotations in AbstractIdeaListView

Rineee:
Hopefully this finally fixes all missing annotations :see_no_evil:


fuzzylogic2000 (approved):
Looks good!
But why do we need to test if the annotations are there? Can we add a comment like "normally annotated in filters (or wherever) and needs to be annotated here when...."?

Rineee:
> Looks good! But why do we need to test if the annotations are there? Can we add a comment like "normally annotated in filters (or wherever) and needs to be annotated here when...."?

Well most of the time they are already annotated in the filters and I would not want to annotate them again if they are already there, because thats quite some database interactions?

PR #4940, opened 2023-02-09
apps/cms: use model fields for storefront counts

### LINKED ISSUE #4358: reevaluate how we load the data on the ladning page
**URL:** mein.berlin.de
**user:** 
**expected behaviour:** 
**behaviour:** The landing page on the prod site is loading pretty slowly. 
**important screensize:**
**device & browser:** 
**Comment/Question:** I guess, it is because we count all items and projects for the tiles. For dev it fairly quick. Maybe instead of calculating every time a user accesses the page, we can run a cron job every night and save these numbers in wagtail or sth. like that?

Screenshot?

fuzzylogic2000:
fixes #4358 

still missing: 

- [x] management command to use as hourly cron job


meinberlin/apps/cms/management/commands/update_storefront_counts.py
```diff
@@ -0,0 +1,58 @@
...
+    help = "Update item and project counts for the storefront."
+
+    def handle(self, *args, **options):
+        for storefront in storefronts.Storefront.objects.all():
+            storefront.num_entries = self.get_num_entries_count()
+            storefront.num_projects = self.get_num_projects_count()
+            storefront.save()
+        for item in storefronts.StorefrontItem.objects.filter(district__isnull=False):
+            district = item.district
+            item.district_project_count = self.get_district_project_count(district)
```
Rineee, line 22: Couldn't you just do `item.district_project_count = item.get_district_project_count()`?

meinberlin/apps/cms/management/commands/update_storefront_counts.py
```diff
@@ -0,0 +1,58 @@
...
+            item.save()
+
+    def get_num_entries_count(self):
```
Rineee, line 25: Also here, couldn't we use the methods from models/storefronts.py here? So either import them here or move them to the Storefront class?

meinberlin/apps/cms/models/storefronts.py
```diff
@@ -62,28 +63,55 @@ def project_url(self):
...
+        )
         active_project_count = plans.count()
         for project in projects:
             if project.active_phase or project.future_phases:
                 active_project_count += 1
         return active_project_count
 
+    def save(self, *args, **kwargs):
+        if not self.district_project_count:
+            self.district_project_count = self.get_district_project_count()
```
Rineee, line 86: Could we also do this with `default =` or does that not work with the district here?

fuzzylogic2000: No, it doesn't work because we do not have the district when we would use the default.

meinberlin/apps/cms/management/commands/update_storefront_counts.py
fuzzylogic2000: Haha, yeah, I wondered how to do that and use the same. :face_palm:
And then I also already wrote this answer this morning and never submitted. :eyes:

Rineee (approved):
yeah, cool!

Rineee:
After deploying on dev (and running the management command), the order of the tiles changed from

image


to this

image


Does this mean anything @fuzzylogic2000 ?

fuzzylogic2000:
@Rineee That is fine as we choose random tiles for the three tiles: you can add up to six (I think), and we show three random ones in random order.

PR #4952, opened 2023-02-13
[#6402] tests: refactor budget list to add test for uncovered lines

phillimorland:


meinberlin/apps/budgeting/assets/__tests__/EndSessionLink.jest.jsx
```diff
@@ -0,0 +1,38 @@
...
+  // overwrite global.fetch with mock function
+  global.fetch = jest.fn().mockRejectedValue('testing: expected network error')
+
+  render(<EndSessionLink />)
+  // toggle modal
+  const modalBtn = screen.getByRole('button', { name: 'End session' })
+  fireEvent.click(modalBtn)
+
+  // get click 'End Session' btn in modal (2nd instance of btn)
+  const endSessionBtn = await screen.getAllByText('End session', { selector: 'button' })[1]
```
phillimorland, line 29: @goapunk do you know why`find/query/getAllByRole('button', { name: 'End session' })[1]` wouldn't work and returns undefined when console.logged? it's more a point of annoyance but role is a good option to check if something is there

package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "6.3.0",
     "@maplibre/maplibre-gl-leaflet": "0.0.19",
+    "@testing-library/user-event": "^14.4.3",
```
fuzzylogic2000: We don't use the ^, but renovate will also remove that once it#s merged.

fuzzylogic2000: You remove it again in the second commit. :)

meinberlin/apps/budgeting/assets/__tests__/EndSessionLink.jest.jsx
goapunk: It works if you also include the  hidden option:
```
screen.getAllByRole('button', {name:'End session', hidden: true})[1]   
```

goapunk: the `await` here is unnecessary, it's only needed if you use `findBy*`

phillimorland: @goapunk so if it is only finding the hidden button and not rendering the modal and finding the button we should delete this right? my understanding is that a test should mimic user behavior, and a user cannot click a hidden button therefore we shouldn't have the test?

goapunk (approved):
cool!

PR #4953, opened 2023-02-13
[6562]apps/plans/export: add 'draft' field to export

khamui:
depending on https://github.com/liqd/adhocracy4/pull/1355


meinberlin/apps/plans/exports.py
```diff
@@ -19,7 +19,6 @@ class DashboardPlanExportView(
     export_views.BaseExport,
     export_views.AbstractXlsxExportView,
 ):
-
```
khamui, line 22: not sure what removes this line automatically. But restoring it does not work (it considers this as empty commit). Shall I have another look?

fuzzylogic2000: I think black does that. I don't know why it does do that now and not earlier, but it's fine, that it does. So, leave it!

meinberlin/apps/plans/exports.py
```diff
@@ -66,6 +65,7 @@ def get_virtual_fields(self, virtual):
         virtual["projects_links"] = gettext("Project Links")
         virtual["created"] = gettext("Created")
         virtual["modified"] = gettext("Modified")
+        virtual["draft"] = gettext("Draft")
```
fuzzylogic2000, line 68: Did you try to ust add it to the fields (line 25 - 39)? That should be enough, I think.

khamui: @fuzzylogic2000 it actually works, but would use "is_draft" as column title. And it does not append to the end of the list. Shall I keep it as virtual field then?

fuzzylogic2000: Ah no, I haven't seen that we do have wording here.

tests/plans/test_export.py
```diff
@@ -156,3 +157,8 @@ def test_reply_to_mixin(plan_factory, project_factory, administrative_district):
 
     assert plan.created.astimezone().isoformat() == export.get_created_data(plan)
     assert plan.modified.astimezone().isoformat() == export.get_modified_data(plan)
+    if plan.is_draft:
```
fuzzylogic2000, line 160: As we create the plan above, we also know if it is draft or not. The usual factory one isn't. https://github.com/liqd/a4-meinberlin/blob/58aa85d2e5fea07841b63ab6cbb11f98e6c384b9/meinberlin/test/factories/plans.py#L29

You could test another one that is and have both cases tested. But let's ask @CarolingerSeilchenspringer first how important the yes/no is, when we do use the False/True for other bools.

fuzzylogic2000 (commented):
Yes, it does work! But it still seems a bit excessive to me to add a whole method because we want to have yes/no instead of True/False. If it stays like that, you should change the test.

tests/plans/test_export.py
fuzzylogic2000: Well, @Rineee also looked at it and is right: we do already add a lot of complicated export stuff here, so one more and exactly as the story says is also fine. I will quickly look at the test, so that both cases are tested.

fuzzylogic2000:
@Rineee I wanted your opinion here. I used some things that were already there and added other ones. But we can also talk about it!
Depending on https://github.com/liqd/adhocracy4/pull/1355

meinberlin/apps/plans/exports.py
```diff
@@ -37,8 +35,13 @@ class DashboardPlanExportView(
         "participation",
         "participation_explanation",
         "organisation",
+        "created",
+        "modified",
+        "is_draft",
     ]
-    html_fields = ["description"]
+    html_fields = ["description", "contact_address_text"]
```
fuzzylogic2000, line 42: This was already in a4 and does what we did below: unescape_and_strip_html

fuzzylogic2000:
@khamui I did clean up the whole export a bit and did move the yes/no thing to a4, so that bool-exports are always exported as yes/no. But your solution also worked fine, I just got a bit annoyed by the code we added before...

Rineee (approved):
Yeah cool, nice cleanup!

The only thing is, that now the draft field is not at the end of the export (as requested in the story), we could achieve it by removing it from the ``fields``` list, but dont know if we want to do that?

Rineee:
I wanted to update A4 to fix tests, but did not manage to do so (see dev channel), sorry :/

fuzzylogic2000:
> Yeah cool, nice cleanup!
> 
> The only thing is, that now the draft field is not at the end of the export (as requested in the story), we could achieve it by removing it from the `fields` list, but dont know if we want to do that?

@Rineee @CarolingerSeilchenspringer said that the order wasn't suuuperimportant. So, let's merge and have it tested and then react to complaints if there are any. :)

I did the update and migrations now and hope this is finally ready!

Rineee (approved):
Yes!

Issue #4963, opened 2023-02-15
check code download on dev


philli-m:
check code download

PR #4965, opened 2023-02-15
Jd 2023 02 codes delete background task

goapunk:
* delete codes in background tasks
* fix some tests
* update translations


goapunk:
maybe fixes #4963

Rineee (approved):
Lgtm! Although locally it does take the same amount of time for me (sth like 30 sec), but I guess worth trying on dev. 
Do you also want to have a look @fuzzylogic2000 ?

goapunk:
> Lgtm! Although locally it does take the same amount of time for me (sth like 30 sec), but I guess worth trying on dev. Do you also want to have a look @fuzzylogic2000 ?

Maybe we first deploy this branch and test it?

Rineee:
> > Lgtm! Although locally it does take the same amount of time for me (sth like 30 sec), but I guess worth trying on dev. Do you also want to have a look @fuzzylogic2000 ?
> 
> Maybe we first deploy this branch and test it?

Good idea, yes!

fuzzylogic2000 (commented):
My thought when setting out to do this the first time was to have an extra package model instead of the number. That way when the package is downloaded, we only need to update that in the package model and can delete the plain tokens whenever it happens.
But let's try if this works first! When the actual downloading is the problem, that also won't help.

fuzzylogic2000 (commented):
Is this ready to be reviewed?
And do we still need the package number here or is a package enough?

meinberlin/apps/votes/views.py
```diff
@@ -26,26 +27,15 @@ class ExportTokenDashboardView(
     permission_required = "a4projects.change_project"
     template_name = "meinberlin_votes/token_export_dashboard.html"
 
-    def _get_number_of_tokens(self):
-        """Return total and downloadable token count for module."""
-        return VotingToken.objects.filter(module=self.module, is_active=True).count()
```
goapunk, line 31: by removing this and using the size we save in the package, codes which have `is_active=False` will still be included in the size (but won't be included in the download). Is that something we need to fix? It would require more db queries, which is why I didn't do it for now

goapunk:
> Is this ready to be reviewed? And do we still need the package number here or is a package enough?

This is ready now I think. I'm not sure what you mean with the package_numbers?

meinberlin/apps/votes/views.py
fuzzylogic2000: I think it is fine like that. We want to use the `is_active` to deactivate tokens that might have been meddled with, right? So, that's not important for the download, as this normally happens before a token is changed to `is_active=False`.

fuzzylogic2000 (commented):
The code looks good, but I haven't tried it out yet. 
I still don't get why we need the package number and can not just use the package without the number. But I will have another look tomorrow morning with a fresh brain.
Maybe @Rineee can also check it out?

meinberlin/apps/votes/models.py
```diff
@@ -56,14 +58,59 @@ def get_or_create_salt_for_module(module):
...
+            .aggregate(Sum("size"))
+            .get("size__sum")
+        )
+        if num_tokens is None:
+            num_tokens = 0
+        return num_tokens
+
+    @staticmethod
+    def create_next_package(module, size):
+        for i in range(100):
```
Rineee, line 83: I dont really get why we need this here? Is it because it might happen that between calculating package_number and actually creating it, there might have already been another one created?

goapunk:
@Rineee @fuzzylogic2000 this is ready for another round of reviews. But don't merge yet, needs squash and translations

meinberlin/apps/votes/exports.py
```diff
@@ -22,35 +22,36 @@ class TokenExportView(
...
             .order_by("pk")
         )
 
     def get_base_filename(self):
-        package_number = self.get_package_number()
-        if not package_number:
-            package_number = 0
-        package_number += 1
+        package = self.get_package()
+        package_number = (
```
fuzzylogic2000, line 51: Cool, you also found a solution for the naming!

meinberlin/apps/votes/models.py
```diff
@@ -60,12 +59,17 @@ def get_or_create_salt_for_module(module):
...
     downloaded = models.BooleanField(default=False)
 
     class Meta:
-        unique_together = ["module", "package_number"]
+        ordering = ["pk"]
+
+    def get_num_tokens(self):
+        return VotingToken.objects.filter(package=self).count()
+
+    def created(self):
```
fuzzylogic2000, line 71: `is_created` as that is what we call bools elsewhere. Like `is_draft`. Ok, we don't always do that. But `created` is still already taken in the TimeStampedModel as a time field. So, I would prefer `is_created`.
And you could add that as a property. Or maybe not. Does that result in that being done every DB call then? Have to read up on that.

fuzzylogic2000 (commented):
Yeah! It reads well. And it nearly works. But when I make a big batch, the "created" is False and not shown in the download section.
```
>>> TokenPackage.objects.all()
<QuerySet [<TokenPackage: TokenPackage object (1)>, <TokenPackage: TokenPackage object (2)>, <TokenPackage: TokenPackage object (3)>]>
>>> TokenPackage.objects.all()[0].module
<Module: Bürger*innenhaushalt (3 Phasen) - new bürherhaushalt s (1)>
>>> TokenPackage.objects.all()[1].module
<Module: Bürger*innenhaushalt (3 Phasen) - new bürherhaushalt s (1)>
>>> TokenPackage.objects.all()[2].module
<Module: Bürger*innenhaushalt (3 Phasen) - new bürherhaushalt s (1)>
>>> TokenPackage.objects.all()[0].size
10
>>> TokenPackage.objects.all()[1].size
1000000
>>> TokenPackage.objects.all()[2].size
200
>>> TokenPackage.objects.all()[0].created()
True
>>> TokenPackage.objects.all()[1].created()
False
>>> TokenPackage.objects.all()[2].created()
False
```

fuzzylogic2000:
> Yeah! It reads well. And it nearly works. But when I make a big batch, the "created" is False and not shown in the download section.

Ah! I didn't wait long enough and forgot about the batch size of 100.000. So, it was simply that not all tasks were run yet. 🤦 It works perfectly fine!

meinberlin/apps/votes/models.py
goapunk: did add the property, also to num_tokens now

meinberlin/apps/votes/models.py
```diff
@@ -56,14 +57,42 @@ def get_or_create_salt_for_module(module):
         return token_salt.salt
 
 
+class TokenPackage(models.Model):
+    module = models.ForeignKey(Module, on_delete=models.CASCADE)
+    size = models.PositiveIntegerField()
```
Rineee, line 62: Didnt we want to remove the size and only use num_tokens instead??

Rineee: > Didnt we want to remove the size and only use num_tokens instead?

Ah, but we need it to check if it can be downloaded already, right?

meinberlin/apps/votes/templates/meinberlin_votes/token_export_dashboard.html
```diff
@@ -8,26 +8,26 @@ <h1 class="u-first-heading">{% translate 'Download voting codes' %}</h1>
     <p class="u-bottom-divider">{% blocktranslate %}Number of generated codes: {{ number_of_module_tokens }}.{% endblocktranslate %}</p>
     <ul class="u-list-reset">
         {% if token_export_url %}
-            {% for number, downloadable in token_packages %}
+            {% for package in token_packages %}
```
Rineee, line 11: maybe we could also do ```for package, downloadable, num_tokens``` here? I find it a bit hard to read..

meinberlin/apps/votes/models.py
goapunk: yes, exactly, and to check if we reached the limit already

meinberlin/apps/votes/templates/meinberlin_votes/token_export_dashboard.html
goapunk: yes, much better, thanks

meinberlin/apps/votes/admin.py
```diff
@@ -11,19 +12,27 @@ class VotingTokenAdmin(admin.ModelAdmin):
...
+        obj.package.save()
+        super().delete_model(request, obj)
+
     def module_name(self, token):
         return token.module.name
 
     def save_model(self, request, obj, form, change):
-        if obj.package_number is None:
-            obj.package_number = VotingToken.next_package_number(obj.module)
+        if obj.package is None:
```
Rineee, line 31: We need ```not hasattr(obj, 'package')``` here, otherwise it throws an error when creating a token..

meinberlin/apps/votes/admin.py
```diff
@@ -11,19 +12,27 @@ class VotingTokenAdmin(admin.ModelAdmin):
...
+            obj.package = TokenPackage.create(module=obj.module, size=1)
+        else:
+            obj.package.size += 1
```
Rineee, line 34: It also does that when editing the token, e.g. setting it to inactive.. hm, not sure how to fix this?

Rineee (commented):
The admin part does not really work, it throws an error when creating a token and also when changing a token, it adds 1 to the package size..

Also it did not show any packages in the dashboard, but that is most likely a me problem, I will check again tomorrow with fresh brain!

meinberlin/apps/votes/admin.py
goapunk: nice catch, fixed

meinberlin/apps/votes/admin.py
goapunk: I just removed it, the package field is read-only anyway, so it will always create a new one

Rineee:
I managed again to produce an error :see_no_evil: I had a package in which all codes were inactive and downloading that throws an error.. I dont know if that is too much of an edge case, but probably we should fix that anyway? Or what do you think @fuzzylogic2000 ?

Apart from that it works perfectly and from side good to merge!

fuzzylogic2000:
> I managed again to produce an error see_no_evil I had a package in which all codes were inactive and downloading that throws an error.. I dont know if that is too much of an edge case, but probably we should fix that anyway? Or what do you think @fuzzylogic2000 ?
> 
> Apart from that it works perfectly and from my side good to merge!

@Rineee Yeeees, I didn't try that, but will also do that. We should probably fix that even though it is an edge case.

Rineee:
> > I managed again to produce an error see_no_evil I had a package in which all codes were inactive and downloading that throws an error.. I dont know if that is too much of an edge case, but probably we should fix that anyway? Or what do you think @fuzzylogic2000 ?
> > Apart from that it works perfectly and from my side good to merge!
> 
> @Rineee Yeeees, I didn't try that, but will also do that. We should probably fix that even though it is an edge case.

@goapunk actually thought of it and raises the bad request here: https://github.com/liqd/a4-meinberlin/pull/4965/files#diff-ef2af1f23503a2f006e8048ad40751390211bb74ca18c6882369e290b66f022bR67
But I guess we should probably not show packages that have only inactive tokens?

fuzzylogic2000:
> > > I managed again to produce an error see_no_evil I had a package in which all codes were inactive and downloading that throws an error.. I dont know if that is too much of an edge case, but probably we should fix that anyway? Or what do you think @fuzzylogic2000 ?
> > > Apart from that it works perfectly and from my side good to merge!
> > 
> > 
> > @Rineee Yeeees, I didn't try that, but will also do that. We should probably fix that even though it is an edge case.
> 
> @goapunk actually thought of it and raises the bad request here: https://github.com/liqd/a4-meinberlin/pull/4965/files#diff-ef2af1f23503a2f006e8048ad40751390211bb74ca18c6882369e290b66f022bR67 But I guess we should probably not show packages that have only inactive tokens?

@goapunk @Rineee I agree. Let's fix that by not showing these packages. I will merge anyway and hope that noone tests that. :p

```
Traceback (most recent call last):
  File "/home/katharina/a4-meinberlin/venv/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/katharina/a4-meinberlin/venv/lib/python3.10/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/katharina/a4-meinberlin/venv/lib/python3.10/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/katharina/a4-meinberlin/venv/lib/python3.10/site-packages/django/contrib/auth/mixins.py", line 104, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/home/katharina/a4-meinberlin/venv/lib/python3.10/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/katharina/a4-meinberlin/venv/lib/python3.10/site-packages/adhocracy4/exports/views.py", line 38, in get
    for rownum, row in enumerate(self.export_rows(), start=1):
  File "/home/katharina/a4-meinberlin/meinberlin/apps/votes/exports.py", line 68, in export_rows
    raise BadRequest

Exception Type: BadRequest at /dashboard/modules/burgerinnenhaushalt-3-phasen-3/download-codes/export/
Exception Value: 
```

PR #4968, opened 2023-02-16
add debug log for bplan api

goapunk:


fuzzylogic2000 (approved):
Cool! Could you also add FIXME? That's what I search for when I search for stuff to do. :)
Right, and that can be found in the access logs, then?

PR #4969, opened 2023-02-16
[6968] votes: increase token length to 16

goapunk:
note: contains migration


meinberlin/apps/votes/migrations/0004_alter_votingtoken_token.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 3.2.17 on 2023-02-16 13:45
+
+from django.db import migrations, models
+import meinberlin.apps.votes.models
+
+
+def delete_all_voting_token(apps, schema_editor):
+    VotingToken = apps.get_model("meinberlin_votes", "VotingToken")
```
goapunk, line 8: this again is slow with a lot of tokens but hopefully we shouldn't have that many?

Rineee: Did you check if it is maybe faster to just drop the whole table (like in the migration before) and then recreate it?

goapunk: no, not yet, but dropping and recreating adds even more code :/ the migrations are already not very nice for the votes app. I'll give it a try / copy it from the previous

fuzzylogic2000: The dropping turned out to be very easy, it's just the first bit with the delete from the migration before. And then I added the recreation to the same table, so that even if the old migration that was in main before would have been run in the meantime, it would all still be fine. here we can also just do two migrations:
1. create one with `--empty` and add the delete
2. just make new migrations recreating the tables with the new state of the model

goapunk:
@Rineee @fuzzylogic2000 I changed the migration now

fuzzylogic2000 (changes requested):
It does work!
But the exported codes and the ones in the list view of the django-admin are still only 12 digits long, because it's using the string method.

meinberlin/apps/votes/forms.py
```diff
@@ -23,21 +23,20 @@ def __init__(self, placeholder=None, *args, **kwargs):
...
-
         # ensure no spaces or dashes
         value = value.replace(" ", "").replace("-", "")
 
         # check the value is not too short or too long
-        if len(value) < 12:
+        if len(value) < 16:
             raise forms.ValidationError(self.error_messages["invalid_short"])
-        elif len(value) > 12:
+        elif len(value) > 19:
```
fuzzylogic2000, line 32: Shouldn't this also be 16 as the dashes and spaces are already removed?

fuzzylogic2000 (approved):
Now it's all good! Just one more question...

meinberlin/apps/votes/forms.py
goapunk: yes, good catch!

fuzzylogic2000 (approved):
Nice!

PR #4976, opened 2023-02-20
Kl 2023 02 moderation rules

fuzzylogic2000:


meinberlin/apps/budgeting/rules.py
```diff
@@ -45,7 +46,10 @@
 
 rules.add_perm(
     "meinberlin_budgeting.moderate_proposal",
-    module_predicates.is_context_moderator | module_predicates.is_context_initiator,
+    is_superuser
+    | module_predicates.is_context_moderator
+    | module_predicates.is_context_initiator
+    | module_predicates.is_context_group_member,
```
Rineee, line 52: Couldnt we use ```is_allowed_moderate_project``` here? https://github.com/liqd/adhocracy4/blob/ca0331aa7612304691fe637903fec4f6936a1f9f/adhocracy4/modules/predicates.py#L87

Rineee:
Yes, cool, just one question!

meinberlin/apps/budgeting/rules.py
fuzzylogic2000: Yes! It does exactly do that. Will do.

meinberlin/apps/budgeting/rules.py
```diff
@@ -45,7 +46,7 @@
 
 rules.add_perm(
     "meinberlin_budgeting.moderate_proposal",
-    module_predicates.is_context_moderator | module_predicates.is_context_initiator,
+    is_superuser | module_predicates.is_allowed_moderate_project,
```
Rineee, line 49: But the ```is_allowed_moderate_project``` already includes the superusers, right?

Rineee:
@fuzzylogic2000 Another question, soorry..

meinberlin/apps/budgeting/rules.py
fuzzylogic2000: How did I not see that?

Issue #4992, opened 2023-02-24
proposal list - url params are not remembered correctly after adding a token


Rineee:
**URL:** https://meinberlin-dev.liqd.net/projekte/module/burgerinnenhaushalt-3-phasen-23/ 
**user:** 
**expected behaviour:** filtering is kept after posting a token
**behaviour:** filtering gets lost if it was applied on react list
**important screensize:**
**device & browser:** 
**Comment/Question:** The filters/query params are passed via a tempatetag [here](https://github.com/liqd/a4-meinberlin/blob/7c6f2c3add38af5a4383a871a1d04d3c3adb3c63/meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_list.html#L26). The problem is that request.GET does not get updated unless the page is reloaded, so it will not reflect any filters that have been changed by react. Instead, it will keep the filters from last page reload (e.g. when coming from map with filters, those filters will be passed, no matter what was changed in the meantime).

Screenshot?

PR #4993, opened 2023-02-24
budgeting: pass mode=list as url param when posting a token, fixes  #…

Rineee:
…4989

When fixing this issue, unfortunately I found another issue: https://github.com/liqd/a4-meinberlin/issues/4992
Not really sure how to solve this one, any ideas @fuzzylogic2000 @goapunk @philli-m @khamui ? I guess we would need to reload the page before posting the token or something?


fuzzylogic2000 (approved):
Cool!

PR #5003, opened 2023-02-27
[#5000]apps/contrib/helpers: input date string has to be modified (re…

### LINKED ISSUE #5000: /projekte broken on safari
**URL:** https://meinberlin-dev.liqd.net/projekte/
**user:** any
**expected behaviour:** 
**behaviour:** cannot see list, map and filter (only header/footer)
**important screensize:**
**device & browser:** safair on mac
**Comment/Question:** 

Screenshot?
Bildschirmfoto 2023-02-27 um 10 39 50

khamui:
…placing whitespace with T) in order to make older safari working closes #5000 

apparently the Date class is differently implemented for older Safari versions (<16).

### DIFF
--- a/meinberlin/apps/contrib/assets/helpers.js
+++ b/meinberlin/apps/contrib/assets/helpers.js
@@ -21,6 +21,7 @@ export const toLocaleDate = (
   locale = 'de-DE',
   formatStyle = { dateStyle: 'long' }
 ) => {
-  const date = new Date(isodate)
+  const normalizedDate = isodate.replace(/ /g, 'T')
+  const date = new Date(normalizedDate)
   return new Intl.DateTimeFormat(locale, formatStyle).format(date)
 }


meinberlin/apps/contrib/assets/helpers.js
```diff
@@ -21,6 +21,7 @@ export const toLocaleDate = (
   locale = 'de-DE',
   formatStyle = { dateStyle: 'long' }
 ) => {
-  const date = new Date(isodate)
+  const normalizedDate = isodate.replace(/ /g, 'T')
+  const date = new Date(normalizedDate)
```
philli-m, line 25: nice! but can you add to the comment to explain why we add this, we will definitely forget :p

philli-m (approved):
lgtm just add little documentation

goapunk:
one think @Rineee and I have been discussing is that instead of doing this in the frontend the backend should probably just send the correct string already?

philli-m:
@goapunk I do remember there was a reason for doing it in the front end, maybe because a+, mB, Opin use different date formats for their project info? not sure if that would affect it

khamui:
@goapunk @philli-m @Rineee yes that would be the best solution, as we did it already somewhere (I think for the app if I am not wrong, maybe @fuzzylogic2000 remembers?).

We use this helper function in some of our components, so I thought the backend solution would exceed this issue's scope.

philli-m:
will close PR as decided to go with backend solution

Issue #4989, opened 2023-02-24
breadcrumb navigation does not work properly


CarolingerSeilchenspringer:
**URL:**https://meinberlin-dev.liqd.net/budgeting/2023-00998/
**user:** user with code
**expected behaviour:** breadcrumb navigation works properly
**behaviour:** project page, enter code above map, get redirected to list, vote on list, go into proposal detail -> breadbrumb is on "map". This does not happen when: project page, enter code above map, get redirected to list, go back to map, go back to list, vote on list, go into proposal detail
**important screensize:**
**device & browser:** tested firefox and safari
**Comment/Question:** 

Screenshot?

PR #5005, opened 2023-02-27
apps/budgeting: redirect to referer after token post to keep filters …

### LINKED ISSUE #4992: proposal list - url params are not remembered correctly after adding a token

fuzzylogic2000:
…from url - fixes #4992

Well, this does work on the list. But I would like to have another look tomorrow morning and try out more cases. I am not supersure it's good. Also, the tests are failing and might give me more hints if it's a good idea. 
So, will WIP it and assign you for review. In case you already have ideas or opinions. :)

### DIFF
--- a/meinberlin/apps/budgeting/views.py
+++ b/meinberlin/apps/budgeting/views.py
@@ -3,6 +3,7 @@ from urllib.parse import parse_qs
 from urllib.parse import urlparse
 
 import django_filters
+from django.shortcuts import redirect
 from django.urls import resolve
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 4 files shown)


meinberlin/apps/budgeting/views.py
```diff
@@ -122,10 +123,8 @@ def post(self, request, *args, **kwargs):
             request.session["token_expire_date"] = (
                 datetime.datetime.now() + TOKEN_SESSION_EXPIRE
             ).timestamp()
-            kwargs["valid_token_present"] = True
-        kwargs["token_form"] = token_form
```
Rineee, line 126: I think we still need to pass the token form here in case it has errors, currently they dont show anymore... not sure about the valid_token_present, we check again what that this is used for

Rineee:
> …from url - fixes #4992
> 
> Well, this does work on the list. But I would like to have another look tomorrow morning and try out more cases. I am not supersure it's good. Also, the tests are failing and might give me more hints if it's a good idea. So, will WIP it and assign you for review. In case you already have ideas or opinions. :)

Hm, but this now reintroduces https://github.com/liqd/a4-meinberlin/issues/4989, right?

Rineee (approved):
Supercool, this is finally working :tada: 

Should we maybe add some comment so we will remember what we did there? But then feel free to merge!

PR #5011, opened 2023-03-02
apps/budgeting: remove own_votes filter from url when session is ende…

### LINKED ISSUE #5010: 3 phase budgeting: own votes filter remains when session is ended
**URL:** 
**user:** 
**expected behaviour:** own filter is removed when the session is ended and thus token is removed
**behaviour:** own votes filter remains and thus an empty list is shown
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

fuzzylogic2000:
…d - fixes #5010

### DIFF
--- a/meinberlin/apps/budgeting/assets/EndSessionLink.jsx
+++ b/meinberlin/apps/budgeting/assets/EndSessionLink.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import django from 'django'
 import Modal from 'adhocracy4/adhocracy4/static/Modal'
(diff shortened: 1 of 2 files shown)


fuzzylogic2000:
@philli-m I do need help here. It had to be a frontend solution and it works, but now the frontend tests are failing.

fuzzylogic2000:
@goapunk @Rineee It does pass now, so could you review? :tada:

Rineee (approved):
:partying_face:

PR #5059, opened 2023-05-15
install urllib3 < 2 as requests doesn't support it yet

goapunk:


fuzzylogic2000 (changes requested):
NIce! It should be alphabetically ordered into the list, though.

fuzzylogic2000 (approved):

PR #5099, opened 2023-06-08
[7409] budgeting 1+2 support

goapunk:


Rineee:
@goapunk This should be ready, could not request review as its your PR

Rineee:
will fix tests

Rineee:
I am still trying to test the new `ProposalDashboardExportView` in views.py, but could not make it work so far, so I thought we could already merge this, so we can go on?

Rineee:
> @Rineee the other stories don't depend on this one, do they? So we don't really need to rush this. In any case it looks good to me! I added a commit to update the fixtures so they use the new phases. Maybe you can have a quick look and then we merge

Well the next one in backlog does and as the other one in the sprint depends on the form story, I thought we might need it! But turns out no :stuck_out_tongue_closed_eyes: 

Maybe @fuzzylogic2000 or @philli-m can also have a quick look as we both worked on this?

Fixture looks good to me!

philli-m:
@Rineee it's looking good to me, think @fuzzylogic2000 is a better bet to check the code though as I don't get all of it

fuzzylogic2000 (approved):
Cool!

PR #5121, opened 2023-06-14
[7412] images/fields: add alt text image field to bplan, container and extproject form

philli-m:
dependent on https://github.com/liqd/adhocracy4/pull/1425


meinberlin/apps/bplan/serializers.py
```diff
@@ -42,6 +43,7 @@ class BplanSerializer(serializers.ModelSerializer):
         required=False,
         write_only=True,
     )
+    image_alt_text = ImageAltTextField(image_name=_("Tile image"))
     image_copyright = serializers.CharField(
         required=False,
         write_only=True,
```
philli-m, line 49: why do we not use the copyright field from a4 here?

meinberlin/apps/plans/models.py
```diff
@@ -115,6 +116,7 @@ class Plan(ProjectContactDetailMixin, UserGeneratedContentModel):
         blank=True,
         help_prefix=_("Visualize your plan with an image underneath the description."),
     )
+    description_image_alt_text = ImageAltTextField(image_name=_("Header image"))
     description_image_copyright = models.CharField(
         verbose_name=_("Header image copyright"),
         blank=True,
```
philli-m, line 122: again we are making a new field instead of importing from a4?

philli-m:
@fuzzylogic2000 I haven't added any migrations as wasn't sure if this was correct way to do it but i added it like the copyright field in a4 then used it in the different apps, is that right? if so will add migrations ect.

meinberlin/apps/bplan/serializers.py
```diff
@@ -66,6 +68,7 @@ class Meta:
             "start_date",
             "end_date",
             "image_url",
```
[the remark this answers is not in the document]
fuzzylogic2000: Because the serializer gets the url and has to get the image from that. This is used via imperia, not the dahsboard, so it's a bit different.

meinberlin/apps/bplan/serializers.py
fuzzylogic2000: Maybe it's different in the serializer?

meinberlin/apps/plans/models.py
fuzzylogic2000: Hmmja, we should have used the one from a4. Should still do that and migrate the data, then. But we do use it for the tile image, so maybe there is some reason? I don't know.

fuzzylogic2000 (commented):
Hmm, I cannot answer your questions. Maybe it would be best to add the image field to the project model and thus also to external projects, bplans and containers and the dashboard forms for these, but do the bplan serializer and the whole plan in a next step?

philli-m:
@fuzzylogic2000 @Rineee so I removed the plan stuff so it's just the apps that use the project model, this needs linking with the a4 PR to work, have not added migrations or to template till i know its correct, plans have added as additional task in taiga

meinberlin/apps/bplan/serializers.py
```diff
@@ -42,6 +43,7 @@ class BplanSerializer(serializers.ModelSerializer):
         required=False,
         write_only=True,
     )
+    image_alt_text = ImageAltTextField(image_name=_("Tile image"))
```
Rineee, line 46: Ahh, now I get why the copyright is not using the ImageCopyrightField.. that is a model field, but we need a serializer field here, so probably best to also add serilalizer.CharField here like copyright field. So sorry I did not get that before!

philli-m: done

meinberlin/apps/bplan/serializers.py
```diff
@@ -42,6 +42,16 @@ class BplanSerializer(serializers.ModelSerializer):
         required=False,
         write_only=True,
     )
+    # can't use a4 field as must be serilizer field
```
goapunk, line 45: typo in serializer

goapunk (approved):
looks good to me. only changelog needs rebasing (not sure what happened with the formatting there

goapunk (approved):
really nice

PR #5131, opened 2023-06-20
makefile: update make commands and descriptions to include js test in…

philli-m:
… make test and stick to naming convention for rest

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,30 +40,31 @@ help:
 	@echo
 	@echo usage:
 	@echo
-	@echo "  make install         				-- install dev setup"
-	@echo "  make clean           				-- delete node modules and venv"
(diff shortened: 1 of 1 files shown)


philli-m:
to ensure all tests are run with `make test` also should we double check our make lint as python lint is now done with black but we don't seem to run that with our lint command?

fuzzylogic2000:
> to ensure all tests are run with `make test` also should we double check our make lint as python lint is now done with black but we don't seem to run that with our lint command?

I looked that up, because I was also confused. But black only does the formatting. It's good to still and also have a linter because that picks up on more things.

fuzzylogic2000 (approved):
The indentation bugs me a bit, will quickly fix that. :)

PR #5150, opened 2023-06-26
[7411] budgeting: disable support button when supporting is not possible

goapunk:


philli-m:
Ready for review however cannot add changelog without refactoring file so it doesn't play nice with others so didn't

meinberlin/apps/budgeting/templatetags/react_support.py
```diff
@@ -42,6 +42,7 @@ def react_support(context, obj):
         "isReadOnly": (
             not has_support_permission and not would_have_support_permission
         ),
+        "isArchived": obj.is_archived,
```
philli-m, line 45: added extra prop as 'isReadOnly' also includes when user not logged in

meinberlin/apps/budgeting/assets/SupportBox.jsx
```diff
@@ -75,13 +83,23 @@ export const SupportBox = (props) => {
     }
   }
 
+  const getSupportClass = () => {
+    if (props.isArchived && props.isReadOnly) {
```
Rineee, line 87: I think you can remove the `&& props.isReadOnly`, right? Because when it is archived, it will always be readonly

meinberlin/apps/budgeting/assets/SupportBox.jsx
```diff
@@ -63,7 +67,11 @@ export const SupportBox = (props) => {
       })
   }
 
-  const handleSupport = () => {
+  const handleSupport = (e) => {
+    if (props.isReadOnly) {
+      e.preventDefault()
```
Rineee, line 72: So now the button does also not react when not logged in and the isReadOnly is true. Is that intended or should we maybe change the order of the if statements, so if not logged in, it always redirects to the log in?

Rineee: But yeah, on the other hand it is annoying if you then log in and cant do anything anyway afterwards..

Rineee (approved):
Really nice! There is one statement, that I think can be deleted and the other comment about redirecting to log in, you can decide, fine with both, so good to merge then!

meinberlin/apps/budgeting/assets/SupportBox.jsx
philli-m: I feel this is more consistent behavior and would leave it up to Marcelo to say what is better for the user

Rineee (approved):
Supercool!

PR #5153, opened 2023-06-27
Jd 2023 6 fix drf char limit translation

### LINKED ISSUE #4973: Wording: no formal adress regarding notice when comment is too long (missing String on Weblate)
**URL:** https://meinberlin-demo.liqd.net/budgeting/2023-00049/
**user:** any
**expected behaviour:** as a user on mein Berlin I want to be adressed in a formal way (Sie) 
**behaviour:** When I write a comment which is too long, I get a notice with an informal adress that the textfield should not contain more than 4000 characters.
**important screensize:** - 
**device & browser:** - 
**Comment/Question:** there is no string on weblate yet, so I cannot translate it correctly

Screenshot?
Bildschirmfoto 2023-02-13 um 10 27 18

goapunk:
fixes #4973


Rineee (approved):
Nice!

PR #5180, opened 2023-07-06
[7421 7412 issues] story related issues

philli-m:
a4: https://github.com/liqd/adhocracy4/pull/1455


goapunk:
@philli-m the categories and labels in the control bar shouldn't be ellipsized (according to the issue)

Issue #5178, opened 2023-07-06
Home page no tiles shown for finished projects, just empty space


maxliqd:
**URL:** mein.berlin.de
**user:** visitor, CMS editor
**expected behaviour:** When a project is finished and was selected to be shown on the start page, a tile is still shown on the start page. You can only select running projects from the list of eligible projects, because the list would otherwise be too long.
**behaviour:** Once a project is finished, no tile is shown. Instead, there is an empty space. 
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?
<img width="686" alt="image" src="https://github.com/liqd/a4-meinberlin/assets/94360651/a3f74a8a-90d1-404d-aa87-8e6c347bd54b">

philli-m:
NOTES:
We limit the list of projects permitted for the storefront to just running ones:
https://github.com/liqd/a4-meinberlin/blob/01b1725d04113a83df80a13906e90808db87ffee/meinberlin/apps/cms/models/storefronts.py#L20
As long as we don't edit the store front items after a selected project is finished it will display the project, this is the intended behavior.
If we edit the storefront after one of the selected projects has finished the projects update and there is now an empty item in the random_items list

Rineee:
@philli-m @maxliqd So the desired behaviour would be that finished projects are still shown, right? You can only not add an already finished project to the storefront, but if it finishes later, it should still show?

maxliqd:
@Rineee Yes, finished projects shall still appear in the storefront if they were added before as a running project. But you cannot add an already finished project, because otherwise the list would be way too long. That's how I understood @philli-m

philli-m:
@Rineee also to note in case it wasn't clear, finished projects are still shown as finished tiles until the list is edited in wagtail then they become blank entries

PR #5188, opened 2023-07-11
cms/models/storefronts: allow already selected projects as possible c…

Rineee:
…hoice for StoreFrontItem to avoid that they disappear once finished, fixes # #5178


philli-m (approved):
perfect! :)

PR #5233, opened 2023-09-14
Berlin redesign 23 09

philli-m:
first round of design updates, due to uncertainty with some elements such as breadcrumbs and header I added fixme's and commented out code that we may still need but have no design for yet


meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -193,13 +207,19 @@ <h3>{% translate 'Responsible body' %}</h3>
...
-                    {% html_date event.date 'DATETIME_FORMAT' %}
+                    <div class="item-detail">
+                        <h2 class="item-detail__title">{{ event.name }}</h2>
+                        <div class="item-detail__meta">
+                            {% html_date event.date 'DATETIME_FORMAT' %}
+                        </div>
+                        <div class="item-detail__content">
+                            {{ event.description | safe | transform_collapsibles | richtext }}
+                            {% block additional_content %}{% endblock %}
+                        </div>
```
philli-m, line 221: added styling from the template which appears not to be used

meinberlin/apps/offlineevents/templates/meinberlin_offlineevents/offlineevent_detail.html
```diff
@@ -20,4 +20,4 @@ <h2 class="item-detail__title">{{ object.name }}</h2>
             </article>
         </div>
     </div>
-{% endblock %}
+{% endblock %} -->
```
philli-m, line 23: We do not appear to use this template but have only commented out for now for testing

meinberlin/apps/budgeting/templates/meinberlin_budgeting/proposal_create_form.html
```diff
@@ -2,6 +2,23 @@
...
+                    <li><a href="{{ project.get_url }}">{% if project.name|truncatechars:50 %}{{ project.name|truncatechars:50 }}{% else %}{{ project.name }}{% endif %}</a></li>
+                    {% if module.is_in_module_cluster  %}
+                        <li><a href="{{ module.get_detail_url }}">{% if module.name|truncatechars:50 %}{{ module.name|truncatechars:50 }}{% else %}{{ module.name }}{% endif %}</a></li>
+                    {% endif %}
+                    <li class=”active” aria-current="page">{% translate 'Submit a new proposal for this project' %}</li>
+                </ol>
+            </nav>
+        </div>
+    </div>
+{% endblock %}
```
philli-m, line 21: @m4ra I ended up just adding the breadcrumb code to all required templates as it's different for each page but i'm not sure if this is the best django/pythony way to do it?

meinberlin/config/settings/base.py
```diff
@@ -577,3 +577,8 @@
 CELERY_RESULT_BACKEND = "redis://localhost:6379"
 CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP = True
 CELERY_RESULT_EXTENDED = True
+
+# BO external footer
+BERLIN_FOOTER_URL = (
+    "https://www.berlin.de/rbmskzl/aktuelles/__i9/std/landesfooter.inc?js=0"
+)
```
philli-m, line 584: @goapunk can I just use the url here?

goapunk: @philli-m why not? that's also how we do it in gd I think

PR #5235, opened 2023-09-18
FIX: 2023 09 add captcha url in base settings

m4ra:
fix issue with no captcha url 
ref: https://github.com/liqd/adhocracy-plus/pull/2470


meinberlin/templates/account/signup.html
```diff
@@ -45,9 +45,11 @@ <h1>{% translate "Register" %}</h1>
             {% endif %}
         </div>
 
-        {% with tabindex="0" %}
-            {% include 'meinberlin_contrib/includes/form_field.html' with field=form.captcha %}
-        {% endwith %}
+				{% if form.captcha %}
```
goapunk, line 48: nitpick: indentation is off

goapunk (changes requested):
LGTM, just a small note

goapunk (approved):
LGTM

PR #5241, opened 2023-09-26
wagtail: make toolbar sticky by default

goapunk:


philli-m (commented):
Nice, I would add a comment to explain why we have this, either in js or webpack? not sure where is best?! Also we only currently use Pascal case for jsx files, should we also start using it for vanilla js too? i'm not sure on why we have the current set up just want to check before we change naming convention

philli-m (approved):
yey!

PR #5243, opened 2023-09-26
[7275] improve performance of project overview

hklarner:


meinberlin/apps/plans/models.py
```diff
@@ -208,6 +208,25 @@ def published_projects(self):
             is_archived=False,
         )
 
+    @cached_property
+    def published_projects_count(self) -> int:
+        projects_are_prefetched = (
```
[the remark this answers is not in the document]
hklarner: I've just fixed the attribute error - i.e. if projects are not prefetched, then the attribute `_prefetched_objects_cache` is missing.

meinberlin/apps/plans/models.py
```diff
@@ -208,6 +208,28 @@ def published_projects(self):
...
+                if project.is_draft or project.is_archived:
+                    continue
+                if project.access in [Access.PUBLIC, Access.SEMIPUBLIC]:
+                    count += 1
+
+            return count
+
+        else:
+            return self.published_projects.count()
+
```
m4ra, line 232: why not move this to the plan serializer method? we can't use the cached_property decorator, but we won't need to check for the prefetch cache name.

hklarner: Yeah, I don't think the `cached_property` is important for this story. I just added it because why not.

> why not move this to the plan serializer method?

Do you mean move the whole function to the serializer? If so, my thinking was that there may be other users of this code (outside of serializing). Like the property `published_projects` just above it is also defined for the model.

But I am pretty indifferent here. Shall I go ahead and move it to the serializer?

meinberlin/apps/plans/models.py
```diff
@@ -208,6 +208,28 @@ def published_projects(self):
             is_archived=False,
         )
 
+    @cached_property
+    def published_projects_count(self) -> int:
+        try:
+            prefetch_enabled = (
+                self.projects.prefetch_cache_name in self._prefetched_objects_cache
+            )
+        except AttributeError:
```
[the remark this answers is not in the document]
hklarner: You mean `self._prefetched_objects_cache`? That's not a function, it's a dict. I've checked that it is working by fixing the value of `prefetch_enabled` and noting that the response time changes significantly. And I have checked that `self._prefetched_objects_cache` is a dict by debugging with a break point.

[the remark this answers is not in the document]
m4ra: to recap, if we don't want to use this hack, we can move the calculation with python to the serializer method, because we know it is only called from the view, and the view does a prefetch. @goapunk?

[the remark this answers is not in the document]
hklarner: I have added a function `is_prefetched(..)` to a new file `plans/helpers.py` to separate the way we count from testing the prefetch.

I guess if we move the code to the serializer we can silently assume it has been prefetched, but for clarity, I would leave the test in to be explicit about why we suddenly count in python. But either way works.

So two decisions pending: 1) move code to serializer? and 2) test whether projects has been prefetched?

meinberlin/apps/plans/models.py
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/plans/models.py
hklarner: decision see next post

meinberlin/apps/plans/serializers.py
```diff
@@ -7,95 +7,72 @@
...
-            return instance.tile_image_copyright
-        elif instance.image_copyright:
-            return instance.image_copyright
-        else:
-            return None
+    def get_tile_image_alt_text(self, plan: Plan):
+        if plan.tile_image:
+            return plan.tile_image_alt_text
+        elif plan.image_alt_text:
+            return plan.image_alt_text
```
goapunk, line 72: reading this function makes it quite easy to think there are only two cases (e.g. tile_image or image_alt_text), however returning None is also possible. While it's implicitly there I find it easier to read with the explicit `else`. Or at least have an `Optional` type hint so it's clear?

goapunk (commented):
Looks mostly good to me!

meinberlin/apps/plans/serializers.py
```diff
@@ -7,95 +7,74 @@
...
-    def get_tile_image_alt_text(self, instance):
-        if instance.tile_image:
-            return instance.tile_image_alt_text
-        elif instance.image_alt_text:
-            return instance.image_alt_text
+    def get_tile_image(self, plan: Plan) -> str:
+        if plan.tile_image:
+            return get_thumbnailer(plan.tile_image)["project_tile"].url
+        elif plan.image:
+            return get_thumbnailer(plan.image)["project_tile"].url
```
m4ra, line 64: keep `instance` here instead of plan, as this is the standard naming in serializers for the model instance

meinberlin/apps/plans/serializers.py
```diff
@@ -7,95 +7,74 @@
...
+            return plan.image_alt_text
         else:
-            return None
+            return ""
+
+    def get_tile_image_copyright(self, plan: Plan) -> str:
+        if plan.tile_image:
+            return plan.tile_image_copyright
+        elif plan.image_copyright:
+            return plan.image_copyright
```
m4ra, line 80: missing an else here too

m4ra (changes requested):
the helper is_prefetched is only used once, so is not justified as a helper yet. since it also has a call to private attributes of django, is better to move the python counting in the serializer with a comment that explains why we do so.

meinberlin/apps/plans/serializers.py
hklarner: done

meinberlin/apps/plans/serializers.py
```diff
@@ -1,101 +1,111 @@
...
+                projects_are_prefetched = bool(
+                    projects.prefetch_cache_name in instance._prefetched_objects_cache
+                )
+            except AttributeError:
+                pass
 
-    def get_url(self, instance):
-        return instance.get_absolute_url()
+        if projects_are_prefetched:
+            count = 0
```
m4ra, line 71: maybe it wasn't clear, but we don't need the prefetch test, just do the count in python, and update the comment why we do so.

hklarner: ah yes, sry, done.

meinberlin/apps/plans/serializers.py
```diff
@@ -1,101 +1,97 @@
...
+    def get_url(self, instance: Plan) -> str:
+        return instance.get_absolute_url()
 
-    def get_type(self, instance):
-        return "plan"
+    def get_published_projects_count(self, instance: Plan) -> int:
+        """
+        This method counts published projects of `instance`.
+        It assumes that the related field `instance.projects` has been prefetched (see `plans/api.py`)
+        ad counts in Python (which is faster than hitting the db with `projects.count()`).
```
m4ra, line 56: small typo: add counts

hklarner: done

meinberlin/apps/plans/models.py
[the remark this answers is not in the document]
hklarner: done, moved it to serializer

meinberlin/apps/plans/serializers.py
goapunk: @hklarner did you forget this? it still shows `plan` for me :)

meinberlin/apps/plans/serializers.py
```diff
@@ -1,101 +1,97 @@
...
-    def get_tile_image(self, instance):
-        image_url = ""
-        if instance.tile_image:
-            image = get_thumbnailer(instance.tile_image)["project_tile"]
-            image_url = image.url
-        elif instance.image:
-            image = get_thumbnailer(instance.image)["project_tile"]
-            image_url = image.url
-        return image_url
+    def get_tile_image(self, plan: Plan) -> str:
```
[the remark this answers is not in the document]
hklarner: done.

m4ra (approved):
Thank you

PR #5245, opened 2023-09-27
workflows: add dev branch

m4ra:


goapunk (approved):
LGTM in general, the only thing I don't like is that we now run main against jammy/22.04  and postgres 14 although it still runs on 20.04 / 12. One alternative option would be to just change all references from main to dev and merge this into the dev branch instead of adding main and dev to the workflow on main. However I don't really expect any problems with running it against 22.04 / 14 so we can also just merge this. What do you think @m4ra ?

m4ra:
yes i can create one PR for dev only. but need to be careful once we apply dev to main in the future.

PR #5252, opened 2023-09-28
REDESIGN [7626] add BO platform footer

philli-m:


Makefile
```diff
@@ -78,6 +79,7 @@ install:
 	if [ ! -f $(VIRTUAL_ENV)/bin/python3 ]; then python3 -m venv $(VIRTUAL_ENV); fi
 	$(VIRTUAL_ENV)/bin/python3 -m pip install --upgrade -r requirements/dev.txt
 	$(VIRTUAL_ENV)/bin/python3 manage.py migrate
+	if [ ! -d $(MEDIA_ROOT) ]; then mkdir -p $(MEDIA_ROOT); fi
```
philli-m, line 82: if you clone repo and `make install`and `make watch`it won't work as the footer contrib tag has no where to put the downloaded footer, by creating media folder on install it solves this problem and also should solve it for the CI to

.github/workflows/django.yml
```diff
@@ -79,6 +79,7 @@ jobs:
           isort --diff -c meinberlin tests
           python manage.py makemigrations --dry-run --check --noinput
           flake8 meinberlin tests --exclude migrations,settings
+          mkdir -p media
```
philli-m, line 82: so same as local, not sure on this but as it's needed for the local setup it should probably be in the test setup too right?

.github/workflows/django.yml
```diff
@@ -75,6 +75,7 @@ jobs:
           PGHOSTADDR: 127.0.0.1
         run: |
           python manage.py collectstatic > /dev/null
+          mkdir -p media
```
philli-m, line 78: media folder also required for CI as it doesn't run make install as it does not need `migrate` command

Makefile
m4ra: No need for the `if` check, the `-p` option only creates the dir if it doesn't exist.

philli-m: @m4ra fixed :)

meinberlin/apps/contrib/templatetags/contrib_tags.py
```diff
@@ -115,3 +118,24 @@ def get_proper_elided_page_range(p, number, on_each_side=1, on_ends=1):
     return paginator.get_elided_page_range(
         number=number, on_each_side=on_each_side, on_ends=on_ends
     )
+
+
+@register.simple_tag
+def get_external_footer():
+    # The BO footer needs to be included here as we can't include external
+    # html in the browser. Add a cron job to update the file periodically via
+    # the get_footer management command.
```
m4ra, line 127: We can have a scheduled celery task instead of a management command. the task can be called also inline directly as a simple function below in line 134.

m4ra (changes requested):
the management command can become a task in contrib/tasks.py
I can take over this PR, install celery beat, and configure celerybeat in the settings. 
@goapunk what do u think?
ref: https://django-celery-beat.readthedocs.io/en/latest/

meinberlin/apps/contrib/templatetags/contrib_tags.py
goapunk: While I agree, this will need work on the admin site as we currently don't have celery beat running or configured, so maybe this is out of scope of thids PR? We already have everything to run a management command as a cron job in a single line. So I would suggest to have a separate PR which converts periodic tasks to celery beat? What do you think @m4ra @philli-m

philli-m: @goapunk @m4ra it's a bit out of my area so happy to go with whatever you guys think. This is separate to the path fix though right?
However we choose to do it, it should be replicated in GD as that's where I got this code from

goapunk:
> the management command can become a task in contrib/tasks.py I can take over this PR, install celery beat, and configure celerybeat in the settings. @goapunk what do u think? ref: https://django-celery-beat.readthedocs.io/en/latest/

@m4ra sounds good to me!

philli-m:
@m4ra @goapunk whats the status here? was it taken over? 
Also this is also how we might look at collecting the footer and JS files for GD right? maybe I'll ref this branch in the release story so we don't forget

m4ra:
> [@m4ra](https://github.com/m4ra) @goapunk whats the status here? was it taken over?
> Also this is also how we might look at collecting the footer and JS files for GD right? maybe I'll ref this branch in the release story so we don't forget

@philli-m we are finishing with configuring celerybeat in the admin repo. then I need to turn the management command to celery task.
For GD though, since we are not installing celery, need to see how we just call the function from a cron job, as we briefly discussed earlier before. Yes leave the story, I will update it once I start working on convering the mngtm command to celery task

meinberlin/apps/contrib/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def download_footer(footer_url, filepath):
+    try:
+        response = requests.get(footer_url, timeout=2.0)
+        filepath.write_bytes(response.content)
+        # clear cache
+        cache.delete("footer")
```
goapunk, line 17: why delete the cache here? we are updating it in `periodic_footer_update` anyway. And if the download fails we don't want it deleted because otherwise the footer will disappear then?

tests/contrib/test_get_footer.py
```diff
@@ -0,0 +1,66 @@
...
+from django.test import override_settings
+
+from meinberlin.apps.contrib.tasks import periodic_footer_update
+from meinberlin.apps.contrib.templatetags.contrib_tags import get_external_footer
+
+
+def test_get_footer(clear_cache_footer):
+    footer = cache.get("footer")
+    assert footer is None
+    footer = get_external_footer()
```
goapunk, line 13: these tests will fail if the berlin.de servers are not available, which happened more than one would think and sometimes for multiple days. I'm not sure if it's a good idea to rely on them.  I guess we could use something like `request-mocks` or just keep the tests which don't download anything. What do you think?

goapunk (changes requested):
Added some comment

meinberlin/apps/contrib/templatetags/contrib_tags.py
```diff
@@ -122,20 +124,14 @@ def get_proper_elided_page_range(p, number, on_each_side=1, on_ends=1):
...
+    The BO footer needs to be included here as we can't include external
+    html in the browser. Add the celery task 'periodic_footer_update' in
+    django admin periodic tasks.
+    """
+    footer = cache.get("footer", "")
+    if not footer:
+        succeeded = periodic_footer_update()
+        if succeeded:
+            footer = cache.get("footer")
+    return footer
```
goapunk, line 137: the footer needs to be marked as safe with `mark_safe()`, otherwise it will be displayed as string instead of rendered as html

meinberlin/apps/contrib/tasks.py
```diff
@@ -0,0 +1,43 @@
...
+@shared_task
+def periodic_footer_update():
+    footer_url = hasattr(settings, "BERLIN_FOOTER_URL")
+    footer_filename = hasattr(settings, "BERLIN_FOOTER_FILENAME")
+    downloaded = False
+    if footer_url and footer_filename:
+        filepath = Path(settings.MEDIA_ROOT) / settings.BERLIN_FOOTER_FILENAME
+        downloaded = download_footer(settings.BERLIN_FOOTER_URL, filepath)
+        if downloaded:  # set footer in the cache
+            cache.set("footer", filepath.read_bytes())
```
goapunk, line 37: better use `read_text()` here as otherwise it will give weird output in combination with `mark_safe`

tests/contrib/test_get_footer.py
m4ra: I will apply a mocked request

goapunk (changes requested):
very nice, only tests need fixing (currently failing)

m4ra:
> very nice, only tests need fixing (currently failing)
yes because of reading the file before downloading, which another test from another module has downloaded already. am updating it. also adding some docs

m4ra:
@goapunk tests pass now.
And we can add the periodic task in the admin, once dev gets rebased on main with the celerybeat part, @philli-m will do a rebase in one of her outstanding PR.

goapunk (approved):
really nice :)

m4ra (approved):

PR #5255, opened 2023-10-02
config/settings: add assets in staticfiles dirs

m4ra:


meinberlin/config/settings/base.py
```diff
@@ -207,6 +207,7 @@
 
 STATICFILES_DIRS = [
     os.path.join(PROJECT_DIR, "static"),
+    os.path.join(PROJECT_DIR, "assets"),
```
philli-m, line 210: @m4ra so looking at this made me think it actually might be a webpack issue (sorry!!) as webpack should add the files to static, I kind of blanked on that as I thought it was just not collecting the new static files but if they are never in static then webpack has failed to add them, I'll have a look after lunch.

m4ra: there are some other files that are missing from static folders. how does webpack work? Maybe it's fine to have the assets too, for local development.

philli-m: we use something called CopyWebpackPlugin that should copy over all image files from the asset folder into the defined output folder which is static which is why we shouldn't need to add assets to our settings, i'm not sure of the advantages to using webpack over static in these cases though but as we bundle our other assets with webpack, it's probably more consistent?

goapunk: @m4ra during the build on the build server we call `make release`, which in turn runs `npm run build:prod`. Locally I think it happens with `make watch` but don't know from the top of my head

m4ra:
Locally yes, I've seen it in make watch.
Closing this PR.

PR #5256, opened 2023-10-02
[7275] adds cache for api endpoints

hklarner:


meinberlin/apps/projects/api.py
```diff
@@ -39,43 +43,50 @@ def get_queryset(self):
...
+        key = get_projects_cache_key(suffix="private")
+        filter_kwargs = dict(is_draft=False, is_archived=False, access=Access.PRIVATE)
+        projects = cache_add_or_query(
+            key=key, query_set=Project.objects, filter_kwargs=filter_kwargs
         )
-        if private_projects:
-            not_allowed_projects = [
+
+        if projects:
+            forbidden_ids = [
```
m4ra, line 82: `not_accesible` or something less harsh than forbidden

meinberlin/config/settings/production.py
```diff
@@ -3,6 +3,12 @@
 DEBUG = False
 STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+    }
```
m4ra, line 9: use redis here
https://docs.djangoproject.com/en/4.2/topics/cache/#redis

m4ra (changes requested):
looks very good overall

meinberlin/apps/projects/api.py
```diff
@@ -39,43 +43,50 @@ def get_queryset(self):
         )
         return projects
 
+    def list(self, request, *args, **kwargs):
+        key = get_projects_cache_key(suffix=self.request.GET.get("status"))
```
goapunk, line 47: One thing I didn't take into account when I tested this is that if we cache by status here, what happens in the following scenario:
1. running projects get cached (let's say for 2h) at 13:00
2. a project starts at 13:30
Will the project then not be listed until the cache expires at 15:00 ? If so we probably won't be able to do it like this as users should find the project once it started running. What if we cache the unfiltered qs instead? I guess we can't filter it after retrieving it, right? We could use a very short cache time but then the caching only helps with bursts of users, not in general

goapunk (changes requested):

meinberlin/apps/bplan/api.py
```diff
@@ -4,6 +4,8 @@
 from adhocracy4.api.mixins import OrganisationMixin
 from adhocracy4.api.permissions import ViewSetRulesPermission
 
+from ..cache_helpers import cache_add_or_serialize
+from ..cache_helpers import get_bplans_cache_key
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/bplan/signals.py
```diff
@@ -1,6 +1,7 @@
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
+from ..cache_helpers import cache_delete
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/cache_helpers.py
```diff
@@ -0,0 +1,72 @@
...
+from typing import Optional
+
+from django.core.cache import cache
+from django.db.models import QuerySet
+from rest_framework.response import Response
+from rest_framework.viewsets import GenericViewSet
+
+from adhocracy4.organisations.models import Organisation
+
+log = getLogger(__name__)
```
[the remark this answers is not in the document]
hklarner: I always that that it is useful to be able to filter by logger as well, but it depends on how we use logging files.

I have moved it to `app/__init__.py`.

meinberlin/apps/cache_helpers.py
```diff
@@ -0,0 +1,72 @@
...
+
+    return filtered_query
+
+
+def cache_add_or_serialize(
+    key: str, view_set: GenericViewSet, timeout: int = DEFAULT_TIMEOUT
+) -> Response:
+    data = cache.get(key)
+
+    if data is None:
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/projectcontainers/api.py
```diff
@@ -5,13 +5,17 @@
 from meinberlin.apps.projectcontainers.models import ProjectContainer
 from meinberlin.apps.projectcontainers.serializers import ProjectContainerSerializer
 
+from ..cache_helpers import cache_add_or_serialize
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/projects/api.py
```diff
@@ -8,8 +8,12 @@
 from meinberlin.apps.projects import serializers as project_serializers
 from meinberlin.apps.projects.filters import StatusFilter
 
+from ..cache_helpers import cache_add_or_query
+from ..cache_helpers import cache_add_or_serialize
+from ..cache_helpers import get_projects_cache_key
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/projects/api.py
hklarner: done

meinberlin/apps/bplan/api.py
```diff
@@ -31,3 +33,7 @@ def get_serializer_context(self):
             }
         )
         return context
+
+    def list(self, request, *args, **kwargs):
```
goapunk, line 37: the bplan api only allows `create` or `update` , not list - so I think we don't actually need caching here (unless I am missing something, sorry)

meinberlin/apps/cache_helpers.py
```diff
@@ -0,0 +1,72 @@
...
+from rest_framework.response import Response
+from rest_framework.viewsets import GenericViewSet
+
+from adhocracy4.organisations.models import Organisation
+
+log = getLogger(__name__)
+
+
+KEYS = {
+    "projects": set(),
```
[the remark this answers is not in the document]
hklarner: > you can't use (non-constant) globals as they won't be shared between the processes

Yes, but this is true for the the in-memory cache (`LocMemCache`) as well:
- https://docs.djangoproject.com/en/4.2/topics/cache/#local-memory-caching
- quote: `This cache is per-process (see below) and thread-safe.`

> with multiple threads and processes you can't use (non-constant) globals

As far as I know, our project is single threaded. In production we start it with gunicorn which in turn starts multiple workers (processes). The design of this cache is currently per-process, meaning each worker will have to cache a url (meaning up to "number of gunicorn workers" users may request a url before it is "fully" cached). I think in that case this setup works.

But the idea of using Redis sounds exciting:
- processes can share the cache
- we can glob cache keys (we don't have to keep track of them ourselves)

meinberlin/apps/cache_helpers.py
```diff
@@ -0,0 +1,72 @@
+from logging import getLogger
```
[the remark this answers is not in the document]
hklarner: The advantage of having loggers that are not shared is that you can filter more easily (simplest case: you include the logger name in the log message). So you can easily look at all cache logs (without knowing what the messages might be). But it depends on how we want to use log messages.

I have moved it to `apps/contrib/__init__.py`.

meinberlin/apps/bplan/api.py
hklarner: good catch! I will remove it.

meinberlin/apps/projectcontainers/api.py
```diff
@@ -5,13 +5,17 @@
...
+class ProjectContainerViewSet(viewsets.ReadOnlyModelViewSet):
     def get_queryset(self):
-        containers = ProjectContainer.objects.filter(
+        return ProjectContainer.objects.filter(
             is_draft=False, access=Access.PUBLIC, is_archived=False
         )
-        return containers
+
+    def list(self, request, *args, **kwargs):
+        return cache_add_or_serialize(key="containers", view_set=self)
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/cache_helpers.py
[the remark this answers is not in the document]
goapunk: If the cache is per-process, can we reliably clear the cache with signals ? I'm not super familiar with the signals and whether they are thread-safe (probably depending on your code for the signal) but they won't be process-safe, so deleting a cache won't reliable, right?

meinberlin/apps/extprojects/api.py
```diff
@@ -5,8 +5,10 @@
 from meinberlin.apps.extprojects.models import ExternalProject
 from meinberlin.apps.extprojects.serializers import ExternalProjectSerializer
 
+from ..cache_helpers import cache_add_or_serialize
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/plans/api.py
```diff
@@ -3,9 +3,14 @@
 from meinberlin.apps.plans.models import Plan
 from meinberlin.apps.plans.serializers import PlanSerializer
 
+from ..cache_helpers import cache_add_or_serialize
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/cache_helpers.py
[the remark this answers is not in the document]
goapunk: yes, sorry, was a bit unclear!

meinberlin/apps/extprojects/api.py
```diff
@@ -18,3 +20,6 @@ def get_queryset(self):
     def get_serializer(self, *args, **kwargs):
         now = timezone.now()
         return ExternalProjectSerializer(now=now, *args, **kwargs)
+
+    def list(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
hklarner: It seems that the signature is correct the way it it, see e.g.
```
ExternalProjectViewSet
  viewsets.ReadOnlyModelViewSe
    mixins.ListModelMixin
      def list(self, request, *args, **kwargs):
        ..
```
- https://github.com/encode/django-rest-framework/blob/605cc4f7367f58002056453d9befd3c1918f6a38/rest_framework/mixins.py#L39
- I don't have experience with fully / partially supporting this

I changed it so the list is now always `def list(request):`

meinberlin/apps/plans/api.py
```diff
@@ -3,9 +3,14 @@
...
 
-class PlansListViewSet(viewsets.ReadOnlyModelViewSet):
+
+class PlansViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = PlanSerializer
 
     def get_queryset(self):
         return Plan.objects.filter(is_draft=False)
+
+    def list(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
hklarner: done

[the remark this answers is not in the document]
hklarner: EDIT: I think the signature is correct / we should not change it.

meinberlin/apps/extprojects/api.py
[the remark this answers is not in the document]
hklarner: EDIT: I think the signature is correct and we should not change it.

meinberlin/apps/projects/api.py
hklarner: Very good point. At first glance I can't find a signal for projects that started. If we had such a signal we could just hook it to the cache invalidation.

But I isn't there a email notification for projects that started? :thinking:

meinberlin/apps/cache_helpers.py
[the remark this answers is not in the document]
hklarner: > If the cache is per-process, can we reliably clear the cache with signals ?

Very good point! I think you are right. This can happen:
- 3 gunicorn workers are started
- all workers cache the projects
- worker 1 is called to delete a project
  - worker 1 triggers signal and clears its cache
- worker 2 and 3 still have projects cached because they did not delete the project / trigger the signal

meinberlin/apps/projects/api.py
goapunk: there are no signals I think as that would require some background task which constantly checks if projects started/ended. So it's done by the rather complicated logic with different serializers for each status. However, you are right that there are emails if a project starts, so maybe there is something like that ?

meinberlin/apps/contrib/cache_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+    return key
+
+
+def get_projects_cache_key(suffix: Optional[str]) -> str:
+    return add_key(
+        namespace="projects", key=f"projects_{suffix}" if suffix else "projects"
+    )
+
+
+def get_bplans_cache_key(organisation: Organisation) -> str:
```
goapunk, line 31: we don't need this anymore, right?

hklarner: good catch, done.

meinberlin/apps/extprojects/api.py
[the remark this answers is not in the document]
m4ra: Right, my mistake, thought it was a plain viewset.

meinberlin/apps/contrib/cache_helpers.py
```diff
@@ -22,14 +23,19 @@ def get_key(namespace: str, suffix: str) -> str:
...
         REDIS_CLIENT.delete(names=keys)
 
         logger.info(f"cache delete: {namespace=}, {len(keys)=}, {context=}")
 
 
+@shared_task
+def cache_delete_task(namespace: str):
+    cache_delete(namespace=namespace, context=dict(caller="celery"))
+
+
```
m4ra, line 38: tasks should be in a tasks.py file

Now the reason I asked for this task to be in a separate PR is because it needs scheduling with celerybeat. You can start off in a new PR and we can take over if you don't have time to finish by next week.
https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html
@hklarner

hklarner: ok removed it from this PR.

meinberlin/apps/contrib/cache_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+    return "_".join(terms)
+
+
+def cache_delete(namespace: str, context: Dict[str, str] = None):
+    if REDIS_IS_ENABLED:
+        keys = REDIS_CLIENT.keys(pattern=f"*{namespace}*")
+        REDIS_CLIENT.delete(names=keys)
+
+        logger.info(f"cache delete: {namespace=}, {len(keys)=}, {context=}")
+
```
m4ra, line 31: an else here to log that cache is not enabled would be good

hklarner: done

m4ra (commented):
LGTM!
it would be good to have a test for the list views, by checking cache keys inline, before and after calling the list view. Before the GET, cache doesn't have the key, after it does. then remove the key from the cache at the end of test case.

meinberlin/apps/projects/apps.py
```diff
@@ -6,6 +6,8 @@ class Config(AppConfig):
     label = "meinberlin_projects"
 
     def ready(self):
+        import meinberlin.apps.projects.signals  # noqa:F401
+
         from . import overwrites
```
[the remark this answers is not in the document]
hklarner: done

meinberlin/apps/projects/api.py
```diff
@@ -39,43 +41,55 @@ def get_queryset(self):
...
         super().__init__(*args, **kwargs)
         now = timezone.now()
         self.now = now
 
     def get_queryset(self):
-        private_projects = Project.objects.filter(
-            is_draft=False, is_archived=False, access=Access.PRIVATE
+        filter_kwargs = dict(is_draft=False, is_archived=False, access=Access.PRIVATE)
+        projects = caching.add_or_query(
+            namespace=reverse("projects-list"),
```
m4ra, line 78: this should be privateprojects. Better just give hardcoded values everywhere you call these functions, not dependent on url pattern

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,50 @@
...
+        ("plans-list", PlanFactory, {}),
+        ("extprojects-list", ExternalProjectFactory, {"access": Access.PUBLIC}),
+        ("containers-list", ProjectContainerFactory, {}),
+        # ("projects-list", ProjectFactory, {"access": Access.PRIVATE}),
+    ],
+)
+def test_list_url(client, url_name, factory, factory_kwargs):
+    n_objects = 3
+    url = reverse(url_name)
+    cache_key = caching.get_key(namespace=url)
```
m4ra, line 39: oops, now i realize it, the use of the URL includes slashes, and not only the viewset's basename. let's just use hardcoded values.

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,50 @@
...
+    cache_value_before = cache.get(cache_key)
+
+    objects = factory.create_batch(size=n_objects, **factory_kwargs)
+    response = client.get(url)
+    cache_value_after = cache.get(cache_key)
+
+    assert response.status_code == 200
+    assert cache_value_before is None
+    assert len(cache_value_after) == len(objects) == n_objects
```
m4ra, line 48: check also the string pattern of the value

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,50 @@
...
+def test_cache_delete():
+    ...
+
+
+def test_cache_add_or_query():
+    ...
+
+
+def test_cache_add_or_serialize():
+    ...
```
m4ra, line 22: the delete cache you can include directly in the test_list_url.
the other two are indirectly tested with the GET call in the same test, so no need to expand.

meinberlin/apps/cache_helpers.py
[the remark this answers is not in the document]
hklarner: done, we are now only supporting redis if cache is enabled.

meinberlin/apps/projects/api.py
hklarner: done

tests/contrib/test_caching.py
hklarner: done

tests/contrib/test_caching.py
hklarner: that is quite tricky because the serialized object and the original model can be (and are) very different. I would have to pass each serializer to the test data. But the serializers are called differently because the `__init__(..)` has been overridden (e.g. `ProjectsSerializer` has to be called with a `now: datetime` argument, but `PlanSerializer` doesn't.

Even testing a single attribute is not straightforward as the serializer renames practically everything.

So I would suggest to stick with testing the length. Since the DB is empty at test start, it would be surprising if the cache returns the exact same number of objects, but they are different object. Right?

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,50 @@
...
+    "url_name,factory,factory_kwargs",
+    [
+        ("projects-list", ProjectFactory, {}),
+        ("plans-list", PlanFactory, {}),
+        ("extprojects-list", ExternalProjectFactory, {"access": Access.PUBLIC}),
+        ("containers-list", ProjectContainerFactory, {}),
+        # ("projects-list", ProjectFactory, {"access": Access.PRIVATE}),
+    ],
+)
+def test_list_url(client, url_name, factory, factory_kwargs):
```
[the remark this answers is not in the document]
hklarner: done, renamed to `test_calling_list_api_creates_cached_value(..)`

tests/contrib/test_caching.py
m4ra: I mean check that the namespace is part of the key.
But the tests are failing atm @hklarner

hklarner: The cache key is created with
```python
cache_key = caching.get_key(namespace=namespace)
```
i.e. the namespace is an argument that is required to create a key. All API view sets that have caching enabled (`ProjectViewSet`, `PlanViewSet`, etc.) have the name space hard-coded.

The test already makes the assumption that the value is saved in the correct key because it calls
```
cache_value_after = cache.get(cache_key)
```
So it already asserts that each view has the correct hard coded namespace.

Is that what you mean?

hklarner: not sure what to assert here, the code never tells me what key it uses, instead I request the cache value for a key. So the key is part of the input for the test. If I misunderstand then please add explicitly what I should assert and I will. Marking as "done" for now.

docs/caching.md
```diff
@@ -42,8 +42,10 @@ The name space for each API endpoint is hard-coded and prefixed by a global cach
...
 - `api_cache_plans`
 
-In production, we use redis as cache backend (`django_redis.cache.RedisCache`, see `settings/production.py::CACHES`). For development the cache backend is disabled (see `settings/base.py::CACHES`). If you want to enable it locally, then copy the production settings to your `settings/local.py`.
+In production, we use redis as cache backend (`django_redis.cache.RedisCache`, see `settings/production.py::CACHES`). For development the cache backend is disabled (undefined). If you want to enable it locally, then copy the production settings to your `settings/local.py`.
+
+All tests that relate to caching are conditional. If the redis caching backend is not detected (see `caching.py::REDIS_IS_ENABLED`) then tests are skiped.
 
 Some more notes about the caching module:
-- it is indented to be imported as a module instead of importing individual methods so that you call `caching.add_or_query(..)` instead of `add_or_query(..)`, for example
+- it is indented to be imported as a module instead of importing individual methods (e.g. you call `caching.add_or_query(..)` instead of `add_or_query(..)`)
```
m4ra, line 50: it is intended (small typo)

hklarner: done

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,160 @@
...
+
+    cache_value_before = cache.get(keys[0])
+    deleted_keys = caching.delete(keys=keys)
+    cache_value_after = cache.get(keys[0])
+
+    assert cache_value_before == value
+    assert cache_value_after is None
+    assert deleted_keys == keys
+
+    caching.clear()
```
m4ra, line 48: similar

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,160 @@
...
+    cache_value_before = cache.get(key=key)
+    caching.add_or_query(
+        namespace=namespace, query_set=Project.objects, filter_kwargs=filter_kwargs
+    )
+    cache_value_after = cache.get(key=key)
+
+    assert cache_value_before is None
+    assert list(cache_value_after) == list(cache_value_expected)
+
+    caching.clear()
```
m4ra, line 71: call the caching.delete here instead.

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,160 @@
...
+    view_set.request = HttpRequest()
+
+    cache_value_before = cache.get(key=key)
+    caching.add_or_serialize(namespace=namespace, view_set=view_set)
+    cache_value_after = cache.get(key=key)
+
+    assert cache_value_before is None
+    assert cache_value_after == []
+
+    caching.clear()
```
m4ra, line 89: same

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,160 @@
...
+    objects = factory.create_batch(size=n_objects, **factory_kwargs)
+    response = client.get(url)
+    cache_value_after = cache.get(cache_key)
+
+    assert cache_value_before is None
+    assert len(cache_value_after) == len(objects) == n_objects
+    assert response.status_code == 200
+    assert response.data == cache_value_after
+
+    caching.clear()
```
m4ra, line 125: same

tests/contrib/test_caching.py
```diff
@@ -0,0 +1,160 @@
...
+    cache_value_before = cache.get(cache_key)
+
+    factory(**factory_kwargs)
+    cache_value_after = cache.get(cache_key)
+
+    assert response.status_code == 200
+    assert len(cache_value_before) == len(objects) == len(response.data) == n_objects
+    assert cache_value_after is None
+
+    caching.clear()
```
m4ra, line 160: same

m4ra (changes requested):

meinberlin/apps/contrib/caching.py
```diff
@@ -0,0 +1,96 @@
...
+        logger.info(f"cache delete: {namespace=}, {len(keys)=}, {context=}")
+
+    else:
+        logger.warning(f"cache delete failed: {namespace=}, {keys=}, {context=}")
+
+    return keys
+
+
+def clear():
+    cache.clear()
```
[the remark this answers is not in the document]
hklarner: ok removed

tests/contrib/test_caching.py
hklarner: ok removed

tests/contrib/test_caching.py
hklarner: ok removed

tests/contrib/test_caching.py
hklarner: ok removed

tests/contrib/test_caching.py
hklarner: ok removed

tests/contrib/test_caching.py
hklarner: ok removed

meinberlin/apps/projectcontainers/signals.py
```diff
@@ -0,0 +1,10 @@
+from django.core.cache import cache
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import ProjectContainer
+
+
+@receiver(post_save, sender=ProjectContainer)
+def reset_cache(sender, instance, update_fields, **kwargs):
+    cache.delete("containers")
```
m4ra, line 10: same

m4ra (changes requested):

meinberlin/apps/plans/signals.py
```diff
@@ -0,0 +1,10 @@
+from django.core.cache import cache
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import Plan
+
+
+@receiver(post_save, sender=Plan)
+def reset_cache(sender, instance, update_fields, **kwargs):
+    cache.delete("plans")
```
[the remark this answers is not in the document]
hklarner: oh yes, very good catch! done.

meinberlin/apps/projectcontainers/signals.py
hklarner: great catch, done.

PR #5257, opened 2023-10-02
[7697] add extra context in paragraph detail view

m4ra:
Add project and chapter in paragraph context so breadcrumb template block can be abstracted for all templates.


philli-m (approved):
Nice!

PR #5258, opened 2023-10-02
Jd 2023 10 delete logo

goapunk:


philli-m (approved):
cool, good catch with the not deleted file

PR #5260, opened 2023-10-02
[7274] replace django-ckeditor with django-ckeditor-5

goapunk:
needs https://github.com/liqd/adhocracy4/pull/1499

- [ ] find solution to remove old django-ckeditor (broken migrations when removing the package) in a new PR


meinberlin/apps/activities/migrations/0004_alter_activity_description.py
```diff
@@ -0,0 +1,18 @@
...
+        ("meinberlin_activities", "0003_add_helptext"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="activity",
+            name="description",
+            field=django_ckeditor_5.fields.CKEditor5Field(verbose_name="Description"),
+        ),
+    ]
```
m4ra, line 18: maybe first add a new field, e.g `description_ck5`, then do the data migration from (old) `description` to new one, then drop `description` field, and in a new migration file, rename `description_ck5` to `description`

m4ra (commented):
See my comment for how to deal with migrating date from old field to new ckeditor5.

meinberlin/apps/activities/migrations/0004_alter_activity_description.py
goapunk: This migration works fine already I think. I tagged you on the a4 pr as I don't really know what to do there.

m4ra (approved):

philli-m (changes requested):
tested functionality and following errors:
you cannot save a video
when you add an image you get a browser error and img disappears again error:
Couldn't upload file: automne07-56386274.jpg.
weirdly managed to add 1 logo image and another graphic, seems to be an issue with photos
no text wrapping options around images

goapunk:
> tested functionality and following errors: you cannot save a video when you add an image you get a browser error and img disappears again error: Couldn't upload file: automne07-56386274.jpg. weirdly managed to add 1 logo image and another graphic, seems to be an issue with photos no text wrapping options around images

the issues should be fixed now

philli-m (approved):
approved, just need to consult with Caro concerning the old video issue

philli-m (approved):
videos still there so looking good! will give this to testers now

PR #5266, opened 2023-10-10
Celery beat

m4ra:
Celery tasks can become scheduled/periodic. Periodic tasks can be enabled via django admin
@goapunk salt has celerybeat enabled via the settings, so it should work directly without additional celery beat command:
https://github.com/liqd/admin/blob/13c6c912272de9b208a257c735397ac1fbbc38a7/salt/weblate/settings.py#L925


goapunk (approved):
LGTM

PR #5267, opened 2023-10-10
REDESIGN [7701] replace project header with BO

philli-m:
Icon still present in follow button, need a4 refactor to be fixed but that can be done in follow story


meinberlin/templates/base.html
```diff
@@ -23,8 +23,14 @@
...
-    <link rel="stylesheet" type="text/css" href="{% static 'berlin_css.css' %}" />
-    <link rel="stylesheet" type="text/css" href="{% static 'adhocracy4.css' %}" />
+
+    <!-- FIXME this is to ensure the dashboard styles are untouched by BO styles but not sure its the best way? -->
+    <link rel="stylesheet" type="text/css" href="{% static 'common_css.css' %}" />
+    {% if '/dashboard' in request.path %}
+        <link rel="stylesheet" type="text/css" href="{% static 'adhocracy4.css' %}" />
+    {% else %}
+        <link rel="stylesheet" type="text/css" href="{% static 'berlin_css.css' %}" />
+    {% endif %}
```
philli-m, line 33: This is my first attempt to separate out the styles as the dashboard is getting affected, I don't think this is a perfect solution but to keep everything workable and to notice when deleted or altered styles will be problematic in the dashboard I think its a good idea but also happy for other suggestions?

meinberlin/templates/base.html
```diff
@@ -47,7 +53,11 @@
 <body>
     <a href="#main" class="visually-hidden-focusable">{% translate "Skip to content "%}</a>
     {% block header %}
-      {% include 'header.html' %}
+    {% if '/dashboard' in request.path %}
+        {% userindicator %}
+    {% else %}
+        {% include 'header.html' %}
+    {% endif %}
```
philli-m, line 60: it's not confirmed if we need their header and footer in the dashboard yet, if so that makes the solution above non viable but also I don't know why we would need a platform header in a CMS

meinberlin/templates/base.html
m4ra: A more standard way is to work with blocks. E.g here you can make use of the above extra_css block and point to the berlin CSS, and in the meinberlin/templates/a4dashboard/base_dashboard.html you override it with the a4 CSS.

meinberlin/templates/base.html
m4ra: similar here, make use of the blocks for overriding between templates

meinberlin/templates/base.html
```diff
@@ -76,7 +86,9 @@
 
     </main>
     {% block footer %}
+    {% if '/dashboard' not in request.path %}
       {% include 'footer.html' %}
+    {% endif %}
```
m4ra, line 91: same

m4ra (changes requested):
I went through the template blocks mostly. The styling review cannot really do it as am not part of the design stories process.

philli-m:
@m4ra thank you so much! i'll change those out to use blocks :) 
For style review, I think @goapunk has just been looking at the design from the story and comparing but also Marcelo will also check it once it's on dev anyway

m4ra (approved):

m4ra:
@philli-m LGTM, I don't know if you wanna wait for julian's review too. If not go ahead and merge.

PR #5271, opened 2023-10-18
add workaround for macbook m1 sqlite bug for django < 4

goapunk:


goapunk:
Should we merge this workaround for now? We can remove it once we did the update to django 4

m4ra (approved):

PR #5274, opened 2023-10-23
[7618/7701] assets/follow: update classes with a4 prefix

philli-m:
https://github.com/liqd/adhocracy4/pull/1500


goapunk (approved):
LGTM

PR #5275, opened 2023-10-23
deps: add html linter for django templates to ensure consistency 

philli-m:
**Tasks**
- [x] Changelog

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,8 +57,10 @@ help:
 	@echo "  make coverage					-- write coverage report to dir htmlcov"
 	@echo "  make lint						-- lint all project files"
 	@echo "  make lint-quick				-- lint all files staged in git"
-	@echo "  make lint-fix					-- fix linting for all js files staged in git"
-	@echo "  make lint-python-files			-- lint all python files staged in git"
+	@echo "  make lint-js-fix				-- fix linting for all js files staged in git"
+	@echo "  make lint-html-fix				-- fix linting for all html files passed as argument"
+	@echo "  make lint-html-files			-- lint for all html files with django profile rules"
+	@echo "  make lint-python-files			-- lint all python files passed as argument"
 	@echo "  make po						-- create new po files from the source"
 	@echo "  make mo						-- create new mo files from the translated po files"
 	@echo "  make release					-- build everything required for a release"
@@ -186,17 +188,30 @@ lint-quick:
 	$(VIRTUAL_ENV)/bin/python manage.py makemigrations --dry-run --check --noinput || EXIT_STATUS=$$?; \
 	exit $${EXIT_STATUS}
 
-.PHONY: lint-fix
-lint-fix:
+.PHONY: lint-js-fix
+lint-js-fix:
 	EXIT_STATUS=0; \
-	npm run lint-fix ||  EXIT_STATUS=$$?; \
+	npm run lint-fix || EXIT_STATUS=$$?; \
+	exit $${EXIT_STATUS}
+
+# Use with caution, the automatic fixing might produce bad results
+.PHONY: lint-html-fix
+lint-html-fix:
+	EXIT_STATUS=0; \
+	$(VIRTUAL_ENV)/bin/djlint $(ARGUMENTS) --reformat --profile=django || EXIT_STATUS=$$?; \
+	exit $${EXIT_STATUS}
+
+.PHONY: lint-html-files
+lint-html-files:
+	EXIT_STATUS=0; \
+	$(VIRTUAL_ENV)/bin/djlint $(ARGUMENTS) --profile=django --ignore=H030,H031 || EXIT_STATUS=$$?; \
 	exit $${EXIT_STATUS}
 
 .PHONY: lint-python-files
 lint-python-files:
 	EXIT_STATUS=0; \
 	$(VIRTUAL_ENV)/bin/black $(ARGUMENTS) || EXIT_STATUS=$$?; \
-	$(VIRTUAL_ENV)/bin/isort --diff -c $(ARGUMENTS) --filter-files || EXIT_STATUS=$$?; \
+	$(VIRTUAL_ENV)/bin/isort $(ARGUMENTS) --filter-files || EXIT_STATUS=$$?; \
 	$(VIRTUAL_ENV)/bin/flake8 $(ARGUMENTS) || EXIT_STATUS=$$?; \
 	exit $${EXIT_STATUS}
 
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -1,0 +1,112 @@
+# Linting
+
+We use a set of linters to check our code for errors and to provide commands for
+auto formatting.
+
+## Linters
+
+### Python and Django
+
+#### Black
+
+We use black to format all our python code. It is run as part of the commit hook
+and auto formats all staged code. It can also be invoked manually as part of the
+`lint-python-files` command:
+
+```
+make lint-python-files <path>
+# example for linting all python files for mb
+make lint-python-files meinberlin
+```
+
+For details on black see https://github.com/psf/black
+
+#### Djlint
+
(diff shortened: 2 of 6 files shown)


philli-m:
@m4ra @goapunk @hom3mad3 so this is the first round with the template lint and 2 test files that I reformatted with the reformat command. I did not get it to use the config file as the make command overrides it. But maybe you all want to see what you think of the default and we can go from there, it should ensure better a11y according to the dafault rules which is great, heres the defaults: https://www.djlint.com/docs/linter/

m4ra:
Wah looks good indeed :)
Could you add a brief docs for what we use as linters - djlint, flak8 and black?

philli-m:
@m4ra @goapunk @hom3mad3 I have used the reformtter to do simple html formatting fixes and the rest of the errros that come up repeatedly are listed below, I agree with most but not sure on the first as it's a django thing also on the use of double quotes but I can add a config to ignore these, let me know what you think, also you can see errors in more detail if you run 'make lint' on this branch.
cheers! :) 
rules:
D018 35:16 (Django) Internal links should use the {% url ... %} pattern. <a href="/

H005 3:0 Html tag should have lang attribute. <html xmlns="http://
H006 72:40 Img tag should have height and width attributes. <img width="50" src=
H014 31:98 Found extra blank lines. letters %}
H021 72:40 Inline styles should be avoided. <img width="50" src=
H022 6:8 Use HTTPS for external links. <a class="portal-hea
H023 13:79 Do not use entity references. &#38;
H026 15:20 Empty id and class tags can be removed. <div class=""
H029 14:12 Consider using lowercase form method values. <form method="POST"
H030 3:0 Consider adding a meta description. <html xmlns="http://
H031 4:0 Consider adding meta keywords. <html lang="{{ LANGU

T001 7:57 Variables should be wrapped in a whitespace. {{provider.id}}
T002 82:84 Double quotes should be used in tags. {% translate 'Terms 
T003 63:24 Endblock should have name. Ex: {% endblock body %}. {% endblock %}

m4ra:
@philli-m keep double quotes and lower case form methods in templates too.

goapunk:
fyi: config issue will be fixed by https://github.com/Riverside-Healthcare/djLint/pull/785

philli-m:
@goapunk do we want to configure it or are we happy with how it is?

goapunk:
> @goapunk do we want to configure it or are we happy with how it is?

no, it's probably fine, was just saying it should be fixed as we struggled with it initially. Is this waiting for something or why is it marked as WIP?

Makefile
```diff
@@ -57,7 +57,9 @@ help:
 	@echo "  make coverage					-- write coverage report to dir htmlcov"
 	@echo "  make lint						-- lint all project files"
 	@echo "  make lint-quick				-- lint all files staged in git"
-	@echo "  make lint-fix					-- fix linting for all js files staged in git"
+	@echo "  make lint-js-fix					-- fix linting for all js files staged in git"
+	@echo "  make lint-html-fix					-- fix linting for all html files in git only structure"
```
goapunk, line 61: sorry, maybe a bit dense but what does "in git only structure" mean ?

Makefile
```diff
@@ -186,11 +190,20 @@ lint-quick:
...
-	exit $${EXIT_STATUS}
+	npm run lint-fix || EXIT_STATUS=$$?;
+
+.PHONY: lint-html-fix
+lint-html-fix:
+	EXIT_STATUS=0; \
+	djlint $(SOURCE_DIRS) --reformat || EXIT_STATUS=$$?;
+
+.PHONY: lint-html-files
+lint-phtml-files:
```
goapunk, line 204: `phtml` should be `html`, right?

Makefile
```diff
@@ -186,11 +190,20 @@ lint-quick:
...
+lint-js-fix:
 	EXIT_STATUS=0; \
-	npm run lint-fix ||  EXIT_STATUS=$$?; \
-	exit $${EXIT_STATUS}
+	npm run lint-fix || EXIT_STATUS=$$?;
+
+.PHONY: lint-html-fix
+lint-html-fix:
+	EXIT_STATUS=0; \
+	djlint $(SOURCE_DIRS) --reformat || EXIT_STATUS=$$?;
```
goapunk, line 201: needs to be `$(VIRTUAL_ENV)/bin/djlint`

Makefile
```diff
@@ -186,11 +190,20 @@ lint-quick:
...
+
+.PHONY: lint-html-files
+lint-phtml-files:
+	EXIT_STATUS=0; \
+	djlint $(SOURCE_DIRS) --profile=django || EXIT_STATUS=$$?;
```
goapunk, line 206: needs to be $(VIRTUAL_ENV)/bin/djlint

Makefile
philli-m: I assume a typo and I was going for files staged in git like above but honestly not sure because that isn't right

Makefile
philli-m: yes

goapunk:
@philli-m can't tag you but is ready for a review

Makefile
```diff
@@ -186,17 +188,30 @@ lint-quick:
...
+.PHONY: lint-html-fix
+lint-html-fix:
+	EXIT_STATUS=0; \
+	$(VIRTUAL_ENV)/bin/djlint $(ARGUMENTS) --reformat --profile=django || EXIT_STATUS=$$?; \
+	exit $${EXIT_STATUS}
+
+.PHONY: lint-html-files
+lint-html-files:
+	EXIT_STATUS=0; \
+	$(VIRTUAL_ENV)/bin/djlint $(ARGUMENTS) --profile=django --ignore=H030,H031 || EXIT_STATUS=$$?; \
```
goapunk, line 207: added ignore for the following rules because they will produce an error every where and don't necessarily make sense or need feedback from PMs:
- H030 	Consider adding a meta description.
- H031 	Consider adding meta keywords.

Makefile
```diff
@@ -186,17 +188,30 @@ lint-quick:
...
 	exit $${EXIT_STATUS}
 
 .PHONY: lint-python-files
 lint-python-files:
 	EXIT_STATUS=0; \
 	$(VIRTUAL_ENV)/bin/black $(ARGUMENTS) || EXIT_STATUS=$$?; \
-	$(VIRTUAL_ENV)/bin/isort --diff -c $(ARGUMENTS) --filter-files || EXIT_STATUS=$$?; \
+	$(VIRTUAL_ENV)/bin/isort $(ARGUMENTS) --filter-files || EXIT_STATUS=$$?; \
```
goapunk, line 214: I removed `--diff -c` so it tries to autofix the broken ones

philli-m: @goapunk shall we adjust this in our other projects too?

docs/linting.md
```diff
@@ -0,0 +1,112 @@
...
+part of the `lint` command to lint all files:
+
+```
+# lint all python files
+make lint
+# lint  specific python file or folder
+make lint-python-files <path>
+# example for linting all python files for mb
+make lint-python-files meinberlin
+```
```
philli-m, line 69: @goapunk should we also add details on the auto fix function for isort? can also be added later

docs/linting.md
```diff
@@ -0,0 +1,112 @@
...
+
+#### Stylelint
+
+We use stylelint to check and enforce the codestyle for all our (s)css files.
+It is run as part of the commit hook and checks all staged code. It can also be
+invoked manually via `npm run lint` to check the code.
+
+For configuration specifics see the stylelint config file `.stylelintrc.json`.
+
+For details on eslint see https://github.com/stylelint/stylelint
```
philli-m, line 112: is there a reason not to add examples like above?

philli-m (commented):
looks good, happy to merge and add improvements later

Makefile
goapunk: probably yes

PR #5279, opened 2023-10-24
DEV meinberlin/react: move all react components to own folder structure

philli-m:
**Tasks**
- [x] Changelog


webpack.common.js
```diff
@@ -113,7 +116,7 @@ module.exports = {
         'maplibre-gl/dist/maplibre-gl.css',
         'leaflet.markercluster/dist/MarkerCluster.css',
         'react-bootstrap-typeahead/css/Typeahead.css',
-        './meinberlin/apps/plans/assets/react_plans_map.jsx'
+        './meinberlin/react/projectoverview/assets/react_plans_map.jsx'
```
philli-m, line 119: not sure if this is a good idea but it never made any sense to me why the project overview map was in plans?! so i put it in a new 'app' in react folder but can easily move it back, I just have no idea why it is in plans...

philli-m:
@goapunk @m4ra @hom3mad3 so I thought I would just move the react code and check if it would just run smoothly and it does so figured I would commit and ask some questions:
- what do you think? is this what you thought when we discussed it?
- there are other apps with assets files but the files in them are plain js, would you guys still move them to 'react' folder? 
- using folder named `projectoverview`for the project overview map instead of putting in a plans folder, not sure if that is a good idea? happy to not do it, i just never remember its in plans

m4ra:
point 1: looks good!
point 2: maybe move js files to meinberlin/assets/js?
point 3: I'd say keep it as plans, for the shake of mirroring the structure of django apps. And add a docs about react structure with a note on what the mystic plans contain. Itwould be good to have a docs structure for django too at some point.
@philli-m @goapunk @hom3mad3

philli-m:
3. I have changed projectoverview back to plans

philli-m:
@m4ra @hom3mad3 I will update the js structure in a separate PR so this is ready for review. For documentation we already have a wiki doc for that which I will update

philli-m:
@m4ra @

> @philli-m if the goal of a react folder is to make components a little bit less nested and easier to locate, then this is great as it is.
> 
> i was wondering if we should add folders to `react/contrib/assets` in order to make those shared components more organized, imagining contrib is where all the stuff that will be used on various apps go. we could go from very basic like `react/contrib/assets/containers` and `react/contrib/assets/components` all the way to `react/contrib/assets/list-item`...though i'm not sure we are just adding needless complexity right now.
> 
> if we want to go a step further and organize the shared components, then we might need to put more thought into how contrib will evolve. somewhere we will encounter a bottleneck as the library grows. the intention is to keep things organized, especially since contrib is where we gather elements for use across various apps? but as mentioned, I'm hesitant about adding unnecessary complexity.
> 
> how do we want to tackle the challenge of reusability of components and how reusable should they be (styling variations and behavioral differences considered)?

in answer to structure question I was wondering if instead of contrib we just have an assets folder and have components/list-item ect. in there fits with the frontend structure in the djangoi, contrib is a django thing after all so we actually don't need to continue with it here.

as for reusability, we do have some examples (i think CheckboxBtn) where the classes are props that can be changed so style variations of components are possible but I think for behavioral/functional changes a new component should be used, i'm not sure if anything needs to be done in this PR for that though?

goapunk:
Generally looks good to me, the only thing I think we should change is that we don't need the files to be in an `assets` folder anymore (unless there is some standard I'm unaware of). So I would propose `react/<component>/` instead of `react/<component>/assets/`

philli-m:
@hom3mad3 ok I changed my mind about the assets folder as though it is a better idea to have same structure as apps so have kept with contrib for now and I think we add sub folders for components and containers as we go.
@goapunk I have removed the pointless assets folders and updated the paths. 
so ready for review again, sorry for it being so long!

goapunk (changes requested):
LGTM but needs rebase

philli-m:
@goapunk done

m4ra:
@philli-m there is a conflict to fix. And I'd still like to include docs that references the react wiki entry as these things tend to be informed only orally if someone thinks to ask about it, and also what we use as linters.

philli-m:
@m4ra conflict resolved, let me know what you think about docs

m4ra (approved):

m4ra:
I will merge this one. Docs can go into a4 at some point. Reading through frontend documentation is really helpful for also backend devs like me who are not working with react, and I didn't know about this wiki entry till now. That's why I think this wiki entry can also be added in a4. Docs in the repo is what I look for when I start working on code. ANd I feel am not the only one.

m4ra:
@goapunk you need to merge this one, as you had previous requests.

goapunk (approved):

PR #5293, opened 2023-11-01
apps/budgeting: create moderator status component

hom3mad3:
**Tasks**
- [x] Documentation (inline)
- [x] Changelog

### DIFF
--- a/meinberlin/react/budgeting/BudgetingProposalListItem.jsx
+++ b/meinberlin/react/budgeting/BudgetingProposalListItem.jsx
@@ -3,6 +3,7 @@ import django from 'django'
 import { toLocaleDate } from '../contrib/helpers'
 import { ListItemBadges } from '../contrib/ListItemBadges'
 import { ListItemStats } from '../contrib/ListItemStats'
+import { ModeratorStatus } from '../contrib/ModeratorStatus'
 import VoteButton from './VoteButton'
 
 const updatedOnStr = django.gettext('updated on')
@@ -43,6 +44,7 @@ export const BudgetingProposalListItem = (props) => {
         numOfMoreBadges={proposal.additional_item_badges_for_list_count}
         proposalUrl={proposal.url}
       />
+      <ModeratorStatus badges={proposal.item_badges_for_list[0]} />
       <div className="list-item__vote">
         <div>
           <span className="list-item__author">{proposal.creator}</span>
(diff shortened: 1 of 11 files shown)


philli-m (commented):
Initial review with some answers

meinberlin/apps/budgeting/assets/BudgetingProposalListItem.jsx
```diff
@@ -43,6 +44,7 @@ export const BudgetingProposalListItem = (props) => {
         numOfMoreBadges={proposal.additional_item_badges_for_list_count}
         proposalUrl={proposal.url}
       />
+      <ListItemAmpel badges={proposal.item_badges_for_list[0]} />
```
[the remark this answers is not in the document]
m4ra: What do you need as data payload? we can update the one of the serializers here:
https://github.com/liqd/a4-meinberlin/blob/dev/meinberlin/apps/budgeting/serializers.py

philli-m:
@goapunk can you test this please, I have this console error `Uncaught TypeError: modType is undefined` which doesn't make sense as it is defined but maybe because its done oddley?

meinberlin/react/budgeting/BudgetingProposalListItem.jsx
```diff
@@ -44,7 +44,7 @@ export const BudgetingProposalListItem = (props) => {
         numOfMoreBadges={proposal.additional_item_badges_for_list_count}
         proposalUrl={proposal.url}
       />
-      <ModeratorStatus badges={proposal.item_badges_for_list[0]} />
+      <ModeratorStatus modStatus={proposal.moderator_status} modStatusDisplay={proposal.get_moderator_status_display} />
```
philli-m, line 47: looking good, however these items don't always have a modStatus so we will need conditional formatting either here or in the sub-component to stop the code from erroring when there is no status

philli-m (changes requested):
just one little change and fix tests and it's good to go :)

meinberlin/react/budgeting/BudgetingProposalListItem.jsx
philli-m: it isn't shown at all if there is no mod status set, will send you a design

package.json
```diff
@@ -64,6 +64,7 @@
     "@babel/preset-env": "7.22.14",
     "@babel/preset-react": "7.22.5",
     "@testing-library/react": "14.0.0",
+    "@testing-library/jest-dom": "^6.1.4",
```
philli-m, line 67: the caret can be removed, we don't use them, I think it's to do with our renovate config but maybe @goapunk knows better?

philli-m (commented):
just the caret to remove and I think there is a test failing but i'm not sure as code looks good?!

package.json
philli-m: the coveralls small reductions can be a bit mysterious?! this might be adding the mod stuff to the serializer?! although coveralls is not currently loading for me so I can't check, we don't worry until it's more then 0.5 reduction

philli-m (approved):
great! :)

PR #5301, opened 2023-11-06
fix map teaser js include

goapunk:
**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/cms/templates/meinberlin_cms/blocks/map_teaser_block.html
+++ b/meinberlin/apps/cms/templates/meinberlin_cms/blocks/map_teaser_block.html
@@ -1,9 +1,4 @@
-{% load wagtailimages_tags wagtailcore_tags block_tags react_map_teaser static %}
-
-{% block extra_js %}
-    <script type="text/javascript" src="{% static 'react_map_teaser.js' %}"></script>
-    {{ block.super }}
-{% endblock %}
+{% load wagtailimages_tags wagtailcore_tags block_tags react_map_teaser %}
 
 {% image value.image max-1920x700 as image %}
 <div class="block block__homepage block--map-teaser" style="background-image: url({{ image.url }})">
--- a/meinberlin/apps/cms/templates/meinberlin_cms/home_page.html
+++ b/meinberlin/apps/cms/templates/meinberlin_cms/home_page.html
@@ -14,6 +14,11 @@
     {% include 'meinberlin_cms/includes/cms_breadcrumb.html' %}
 {% endblock %}
 
+{% block extra_js %}
+    {{ block.super }}
+    <script type="text/javascript" src="{% static 'react_map_teaser.js' %}" defer></script>
+{% endblock %}
+
 {% block content %}
     {% image self.header_image max-1920x560 as header_image %}
 


philli-m (approved):
cool!

meinberlin/apps/cms/templates/meinberlin_cms/blocks/map_teaser_block.html
```diff
@@ -1,9 +1,6 @@
 {% load wagtailimages_tags wagtailcore_tags block_tags react_map_teaser static %}
 
-{% block extra_js %}
-    <script type="text/javascript" src="{% static 'react_map_teaser.js' %}"></script>
-    {{ block.super }}
-{% endblock %}
+<script type="text/javascript" src="{% static 'react_map_teaser.js' %}"></script>
```
m4ra, line 3: the extra js block not needed here?

goapunk: this template doesn't extent any other template, so there is no super block, and no other templates extends this, so the block is redundant afaic

goapunk: I'm not sure if having a script tag somewhere in the body is good practice though, what do you think @philli-m ?

m4ra: I don't understand where are these blocks like the map_teaser_block here included?

philli-m: @goapunk I guess it would be best to add it in the extra js tag for the home_page, I assume that would work? although django templates are bit weird about whats loaded when?!
@m4ra the mB homepage is created via the editors in wagtail, that is where the map_teaser_block is added

goapunk: the map_teaser_block is a wagtail block (https://github.com/liqd/a4-meinberlin/blob/dev/meinberlin/apps/cms/blocks.py#L117)  which can be added via wagtail. I think it will then be used here https://github.com/liqd/a4-meinberlin/blob/dev/meinberlin/apps/cms/templates/meinberlin_cms/home_page.html#L69. Does that help?

m4ra: @goapunk yes thanks. the extra js block then can definitely go in the bottom of the home_page.

meinberlin/apps/cms/templates/meinberlin_cms/home_page.html
```diff
@@ -14,6 +14,11 @@
     {% include 'meinberlin_cms/includes/cms_breadcrumb.html' %}
 {% endblock %}
 
+{% block extra_js %}
+    <script type="text/javascript" src="{% static 'react_map_teaser.js' %}"></script>
+    {{ block.super }}
```
m4ra, line 19: maybe the block.super should go before the new script we are including, in case of any js overwrites? and usually scripts are going after content.

goapunk: I moved the super above and added defer, as this is the modern way instead of moving it to the end of the body (https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML#applying_css_and_javascript_to_html)

goapunk: I discussed it with @philli-m  and we should start using defer (or async) everyhwere

m4ra (approved):

PR #5308, opened 2023-11-09
[7638] remove background_task tables

m4ra:
**Tasks**
- [x] Changelog


meinberlin/apps/contrib/migrations/0001_drop_background_task_table.py
```diff
@@ -0,0 +1,9 @@
+# Generated by Django 3.2.20 on 2023-11-09 07:46
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = []
+
+    operations = [migrations.RunSQL("DROP TABLE IF EXISTS background_task")]
```
goapunk, line 9: background tasks creates two tables: `background_task` and `background_task_completedtask`, do you want to keep the second one or is it missing ? :)

goapunk (changes requested):
Nice, just a question about the second table

meinberlin/apps/contrib/migrations/0001_drop_background_task_table.py
m4ra: oh no I don't, I didn't add it, slipped me.

goapunk (approved):
LGTM, just one question I haven't really considered before: by deleting the table manually, if we ever want to add background-tasks back the database still thinks the tables exist in the migration history, so it won't create them again. Do you think this is a problem? Or should we just ignore it?

m4ra:
@goapunk background tasks are not supported in Django 4, so we won't go back to these.

goapunk:
> @goapunk background tasks are not supported in Django 4, so we won't go back to these.

yes, true, was more of a general thought that running custom sql will break the migration history.

m4ra:
The migration is not dependent in any other though. So I don't think that is an issue. And the background tasks migrations come from the package, not from our own repo. Once the package is deleted, its history also is gone afaik

m4ra:
Moreover, if the app label is not in the settings. django migrations won't look for these.

goapunk (approved):
cool

PR #5309, opened 2023-11-09
[7777] django upgrade from 3.2.x to 4.2

m4ra:
**Tasks**
- [x] Changelog


meinberlin/apps/dashboard/views.py
```diff
@@ -204,7 +204,7 @@ class ModuleDeleteView(PermissionRequiredMixin, generic.DeleteView):
     model = module_models.Module
     success_message = _("The module has been deleted")
 
-    def delete(self, request, *args, **kwargs):
+    def form_valid(self, request, *args, **kwargs):
```
goapunk, line 207: does the call to `super().delete` also need to be changed to `super().form_valid` ?

goapunk (approved):
LGTM so far, just a small note about the `form_valid` change

meinberlin/apps/dashboard/views.py
```diff
@@ -204,15 +204,16 @@ class ModuleDeleteView(PermissionRequiredMixin, generic.DeleteView):
...
+        print("outside if check")
```
goapunk, line 208: forgot to remove the prints :)

changelog/7777.md
```diff
@@ -7,3 +7,9 @@
   USE_L10N is by default enabled
   ENGINE from postgresql_psycopg2 to postgresql
   new migrations for relations
+
+- Django from 4.0 to 4.1
+
+- Django from 4.1 to 4.2
+  settings/production.py: change to STORAGES 
+  requirements: update psycopg to v3
```
goapunk, line 15: did you forget to commit the update ?

goapunk (changes requested):
small note about psycopg

also: do we need to fix the `save()` method of the models as described here: https://docs.djangoproject.com/en/4.2/releases/4.2/#setting-update-fields-in-model-save-may-now-be-required ?

meinberlin/apps/cms/models/storefronts.py
```diff
@@ -84,10 +84,12 @@ def get_district_project_count(self):
...
         return active_project_count
 
-    def save(self, *args, **kwargs):
+    def save(self, update_fields=None, *args, **kwargs):
         if not self.district_project_count:
             self.district_project_count = self.get_district_project_count()
-        super().save(*args, **kwargs)
+            if update_fields:
+                update_fields = {"district_project_count"}.union(update_fields)
+        super().save(update_fields=update_fields, *args, **kwargs)
```
m4ra, line 92: @goapunk thanks for the custom save() reminder. I am thinking, this one here, doesn't need always to update the `district_project_count` when the instance is saved? and not only if the count is not set?

goapunk: I'm not completely sure, I didn't write this code but I think the idea is that if it is not set (e.g. when creating the storefront the first time), that it is added here and then it is updated periodically via cron and this management command https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/apps/cms/management/commands/update_storefront_counts.py

goapunk:
some notes:
- the save methods in the vote models haven't been updated, is that on purpose?
-  the tests for the servicekonto are now failing. I assume because of the allauth update. We'll have to ask @CarolingerSeilchenspringer about whether we need the servicekonto or nor. Fixing it might be annoying

m4ra:
```
some notes:

    the save methods in the vote models haven't been updated, is that on purpose?
    the tests for the servicekonto are now failing. I assume because of the allauth update. We'll have to ask @CarolingerSeilchenspringer about whether we need the servicekonto or nor. Fixing it might be annoying
```
The votes save() are not called for updating, it's to generate tokens. So they don't need the update_fields.
The servicekonto, took me a while but it's fixed now. I will push soon my changes.

m4ra:
@goapunk now there is an error with duplicate email constrains, from psycopg for the test that checks signup with existing email. I could inherit from allauth SignUp view/form and handle the exception. But I also tried to bump allauth to latest version, and servicekonto gave new errors. So yes good to ask Caroline, and if the servicekonto is still in use, we need to make it work with latest allauth.

CarolingerSeilchenspringer:
@goapunk @m4ra talked to client and the service konto in its current form is not used. It may come back but maybe in a whole different technical setting so we can take it out for now.

goapunk:
@m4ra there is one instance of `index_together` which should be replaced with `indexes` as described here: https://docs.djangoproject.com/en/4.2/releases/4.2/

meinberlin/apps/projects/migrations/0005_topics.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations
+
+
+def add_topics(apps, schema_editor):
+    if hasattr(settings, "A4_PROJECT_TOPICS"):
+        Topic = apps.get_model("a4projects", "Topic")
+        for topic in settings.A4_PROJECT_TOPICS:
+            try:
+                Topic.objects.get(code=topic[0])
+            except ObjectDoesNotExist:
```
goapunk, line 14: suggestion: `get_or_create(code=topic[0], name=topic[1])` instead of try except

meinberlin/apps/projects/serializers.py
```diff
@@ -62,6 +63,7 @@ class ProjectSerializer(serializers.ModelSerializer, CommonFields):
     title = serializers.SerializerMethodField()
     type = serializers.SerializerMethodField()
     url = serializers.SerializerMethodField()
+    topics = serializers.SerializerMethodField()
```
goapunk, line 66: This was all just hacked together to test, I think having a`TopicsSerializer` which we add here instead of the function would probably be nice ?

m4ra: true

meinberlin/apps/plans/migrations/0058_plan_m2mtopics.py
```diff
@@ -0,0 +1,18 @@
...
+        ("meinberlin_plans", "0057_alter_plan_description"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="plan",
+            name="m2mtopics",
+            field=models.ManyToManyField(to="a4projects.topic"),
+        ),
+    ]
```
m4ra, line 18: this one can merge with 0062

meinberlin/apps/plans/models.py
```diff
@@ -195,10 +196,7 @@ def administrative_district(self):
 
     @property
     def topic_names(self):
-        if hasattr(settings, "A4_PROJECT_TOPICS"):
-            choices = dict(settings.A4_PROJECT_TOPICS)
-            return [choices.get(topic, topic) for topic in self.topics]
-        return []
+        return [topic.name for topic in self.topics.all()]
```
m4ra, line 199: mb add the hasattr check back

meinberlin/apps/plans/forms.py
```diff
@@ -58,6 +82,7 @@ def save(self, commit=True):
             plan.group = group
         if commit:
             plan.save()
+            self._save_m2m()
```
m4ra, line 85: @goapunk I am changing `self` to `plan` for better readability

goapunk (approved):
looks good to me!

PR #5311, opened 2023-11-13
[7760] add topic prio list with viewset

philli-m:
**Tasks**
- [x] Changelog

As topics are different to ideas and have less things I didn't extend the idea view set but gave topics it's own that can also be used in topic maps, but if this isn't correct please let me know!

### DIFF
--- a/meinberlin/apps/topicprio/templates/meinberlin_topicprio/topic_list.html
+++ b/meinberlin/apps/topicprio/templates/meinberlin_topicprio/topic_list.html
@@ -1,5 +1,10 @@
 {% extends extends %}
-{% load i18n module_tags rules static maps_tags %}
+{% load i18n module_tags react_topics rules static maps_tags %}
+
+{% block extra_js %}
+    <script type="text/javascript" src="{% static 'topics.js' %}"></script>
+    {{ block.super }}
+{% endblock %}
 
 {% block phase_content %}
     <div class="container u-spacer-top-double">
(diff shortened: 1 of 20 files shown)


meinberlin/apps/topicprio/urls.py
```diff
@@ -1,7 +1,11 @@
 from django.urls import re_path
 
+from adhocracy4.api import routers as a4routers
+
 from . import views
 
+module_router = a4routers.ModuleDefaultRouter()
+
```
philli-m, line 8: not actually sure if this is needed?

meinberlin/apps/topicprio/templates/meinberlin_topicprio/topic_list.html
```diff
@@ -1,5 +1,10 @@
 {% extends extends %}
-{% load i18n module_tags rules static maps_tags %}
+{% load i18n module_tags react_topics rules static maps_tags %}
+
+{% block extra_js %}
+    <script type="text/javascript" src="{% static 'topics.js' %}"></script>
```
goapunk, line 5: should this have the defer tag?

meinberlin/react/topicprio/TopicPrioList.jsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const fetchTopics = () => {
+    const url = props.topics_api_url
+    fetch(url)
+      .then(resp => resp.json())
+      .then(json => {
+        setData(json.results)
+        setMeta({
+          permissions: json.permissions,
+          locale: json.locale
```
goapunk, line 22: we don't use the locale, right? so this can be removed

philli-m:
@goapunk I copied over the tests from idea api and removed code related to date/time and moderator status but have got to the end of my python test knowledge with trying to fix them... could you take a look?

meinberlin/apps/topicprio/templates/meinberlin_topicprio/topic_list.html
goapunk: with the defer tag it will load async though, which I think is still nice

goapunk (approved):
LGTM

PR #5312, opened 2023-11-13
[7833] add IdeaViewSet 

goapunk:
**Tasks**
- [x] Changelog

### DIFF
--- a/meinberlin/apps/ideas/api.py
+++ b/meinberlin/apps/ideas/api.py
@@ -1,0 +1,300 @@
+from typing import Dict
+from typing import List
+from typing import Tuple
+
+from django.db.models import Model
+from django.db.models import QuerySet
+from django.utils.translation import get_language
+from django.utils.translation import gettext_lazy as _
+from rest_framework import mixins
+from rest_framework import viewsets
+from rest_framework.filters import BaseFilterBackend
+from rest_framework.filters import SearchFilter
+from rest_framework.pagination import BasePagination
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.serializers import ModelSerializer
+
+from adhocracy4.api.mixins import ModuleMixin
+from adhocracy4.api.permissions import RulesPermission
+from adhocracy4.api.permissions import ViewSetRulesPermission
+from adhocracy4.categories import get_category_icon_url
+from adhocracy4.categories import has_icons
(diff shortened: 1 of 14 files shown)


philli-m (commented):
I think this is working well, the api response has everything I think that will be needed for any idea list item and I can make a simplified version for topics. I do have a question though, as I need less information in topics should I make a base view set that can be extended or shall I make one and call it topics as it only needs the name and the comments, and the labels and categories (if added) so it's very basic.
Also did you mean to base it on main?
And did you mean to include the linting experiments? 
I am not sure what else I should test other then the api response, should I be looking at something else?

meinberlin/apps/ideas/api.py
```diff
@@ -0,0 +1,327 @@
...
+from meinberlin.apps.contrib.filters import DefaultsRestFilterSet
+from meinberlin.apps.contrib.filters import OrderingFilterWithDailyRandom
+from meinberlin.apps.contrib.templatetags.contrib_tags import (
+    get_proper_elided_page_range,
+)
+from meinberlin.apps.moderationtasks.models import ModerationTask
+from meinberlin.apps.moderatorfeedback.models import (
+    DEFAULT_CHOICES as moderator_status_default_choices,
+)
+from meinberlin.apps.votes.filters import OwnVotesFilterBackend
```
philli-m, line 40: votes are only in budgeting

goapunk: good catch, fixed

meinberlin/apps/ideas/api.py
```diff
@@ -0,0 +1,300 @@
...
+from .serializers import IdeaSerializer
+
+
+class LocaleInfoMixin:
+    """Add the current locale of the user to the API response"""
+
+    def list(self, request: Request, *args, **kwargs) -> Response:
+        response = super().list(request, args, kwargs)
+        response.data["locale"] = get_language()
+        return response
```
philli-m, line 49: I wonder if this would also go int contrib?

philli-m (approved):
looking good, think everything is there that we need but can improve as we go if not!

meinberlin/apps/ideas/api.py
goapunk: @philli-m good point, it definitely could!

PR #5319, opened 2023-11-16
remove warning message from ModeratorStatus

hom3mad3:
removing the warning message, since it gets triggered every time ModeratorStatus receives no props. This will trigger warnings every time a card has no status assigned to it.


philli-m (approved):
nice! :)

PR #5320, opened 2023-11-16
WIP assets: update to folder structure to separate styles more clearly

philli-m:
**Tasks**
- [ ] Changelog


philli-m:
@hom3mad3 @goapunk so this is my first hit at updating the folder structure to have a more definitive separation between dashboard and public/Berlin Online styles, I am very open to criticism and changes but thought it would be good to get the ball rolling before we get more components and it gets more messy

philli-m:
@goapunk @hom3mad3 if there is no resistance to the suggested structure I will continue with this as there are now getting to be a few too many mixed styles in forms ect. with the filter PR

PR #5322, opened 2023-11-20
[7761] budgeting filters

vellip:
**Tasks**
- [x] Changelog


meinberlin/react/contrib/forms/Input.jsx
```diff
@@ -0,0 +1,19 @@
...
+export const Input = ({ label, id, before, children, className, ...rest }) => (
+  <div className="form-group">
+    <label htmlFor={id} className="form-label">
+      {label}
+    </label>
+    <div className="input">
+      {before &&
+        <div className="input__before">{before}</div>}
+      <input
+        className={'form-control input__element ' + (before && 'input__element--before ') + className + (children && ' input__element--after')}
```
goapunk, line 12: the `className` prop is always `undefined` (at least in this PR) and also ends up as `undefined` in the resulting html, is this intended (e.g. for use in other components?).

meinberlin/react/contrib/ControlBarSearch.jsx
```diff
@@ -27,29 +29,18 @@ export const ControlBarSearch = (props) => {
...
-          {translated.startSearch}
-        </span>
-      </button>
+        <button type="submit" title={translated.startSearch}>
+          <i className="fa fa-arrow-right" aria-hidden="true" />
+          <span className="visually-hidden">
+            {translated.startSearch}
+          </span>
+        </button>
+      </Input>
```
philli-m, line 41: I think we should stick to the bo implementation as much as possible for the search as I don't think we had any design changes? also there are other searches on the site so they should all look as much the same as possible. I think main change here is just the icons, will check with Marcelo as they were using font awesome but recently changed to these bicon ones.

meinberlin/assets/scss/_form.scss
```diff
@@ -26,6 +35,16 @@ select {
...
+// FIXME ensure auto added buttons look inline with rest of UI eg:close on alerts
+button {
+    appearance: none;
+    color: inherit;
+    background: none;
+    font-size: inherit;
+    border: 0;
+    cursor: pointer;
+}
+
```
philli-m, line 47: I am not sure if this should be needed for any of the BO elements as this and control base should already be set in bo styles?!

meinberlin/assets/scss/components/_control-bar.scss
```diff
@@ -53,3 +42,79 @@
...
+    display: flex;
+    align-items: center;
+
+    @media screen and (max-width: $breakpoint-tablet) {
+        width: 100%;
+
+        .button {
+            flex: 1 1 100%;
+        }
+    }
```
philli-m, line 119: some of these styles might be set by using the faceting form container classes, have sent a link

philli-m (commented):
this is mostly just comments about use of bo styles, will take a look at react next

meinberlin/react/contrib/ControlBarFilterPills.jsx
```diff
@@ -0,0 +1,39 @@
...
+const translated = { filterRemove: django.gettext('Remove filter %s') }
+
+const getFilterRemoveText = (label) => {
+  return django.interpolate(translated.filterRemove, [label])
+}
+
+export const ControlBarFilterPills = ({ filters: _filters, onRemove }) => {
+  const filters = _filters.filter(f => f.value && f.value !== '')
+  return filters.length
+    ? (
```
goapunk, line 14: more of a personal choice but I find this rather hard to read, wouldn't something like
```
if filters.length == 0 {
  return null
}
return (....)
```
be a bit easier?

meinberlin/react/contrib/forms/Input.jsx
vellip: This was more in case it might come in handy in the future. I haven't needed it here yet, that's true. I was going to use template syntax here first, and then I could have included a check, but apparently eslint doesn't allow that. That's why I left out the check, because I didn't want this line to grow.
If you use something like [classnames](https://www.npmjs.com/package/classnames) or already have similar utility, I will be happy to convert it.

goapunk: template syntax is not allowed because it kept breaking django translations /gettext for us. Might be fixed once the upgrade to django 4.2 is done. I don't think we have anything else in place but @philli-m might know better. If there's no negative side-effect of having `undefined` as  class I'm also fine with leaving it like this

meinberlin/assets/scss/_form.scss
vellip: It wasn't set in bo styles. I took this `button` style over because without it buttons look like native buttons with that grey background and border but our default style is "make them not look like a button". 

regarding the control-base, I can remove the entire mixin and where it was used, i.e. [here](https://github.com/liqd/a4-meinberlin/pull/5322/files#diff-8386b6a134ac758dcc457ca8b8c73525c88299d1ad799353ae2ee39ed2bb0963R13)?

meinberlin/assets/scss/_variables.scss
```diff
@@ -98,3 +98,7 @@ $enable-dark-mode: false !default;
 
 $purple: #9185be !default;
 $pink: #f6b4cc !default;
+$whitesmoke: #f5f5f5 !default;
+$grey-dark: #dddddd !default;
```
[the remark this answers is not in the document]
vellip: I took the colors and breakpoint names from the BO CSS. There are custom properties set within that. I am not sure if we want to name them differently, or even if we should have a separate variables.scss from here on. @philli-m 

```css
:root {
	--breakpoint-smallest: 430px;
	--breakpoint-palm: 600px;
	--breakpoint-tablet: 920px;
	--breakpoint-tabletlandscape: 1181px;
	--breakpoint-desktop: 1480px
}

:root {
	--color-white: #FFFFFF;
	--color-black: #000000;
	--color-red: #b5050e;
	--color-blue: #253276;
	--color-errorRed: #000000;
	--color-blue-light: #049FE1;
	--color-green: #15846E;
	--color-grey: #F5F5F5;
	--color-grey-dark: #DDDDDD;
	--color-grey-darkest: #454545;
	--color-tabfocus: #0076DD;
	--color-border: #DDDDDD;
	--color-hr: #DDDDDD;
	--color-decorative-primary: #000000;
	--color-decorative-secondary: #000000;
	--color-decorative-tertiary: #000000;
	--color-decorative-light: #FFFFFF;
	--color-decorative-positive: #15846E;
	--color-decorative-destructive: #D52F2E;
	--color-text-base: #000000;
	--color-text-negative: #FFFFFF;
	--color-text-lightest: #454545;
	--color-text-link: #0047D3;
	--color-text-disabled: #DDDDDD;
	--color-text-error: #9B0000;
	--color-text-success: #3f6b25;
	--bgcolor-block-light: #FFFFFF;
	--bgcolor-block-heavy: #F5F5F5;
	--bgcolor-block-dark: #000000;
	--bgcolor-block-colored: #F5F5F5;
	--color-block-footer: #F5F5F5;
	--color-object-panel: #E8EDF2;
	--color-message-default: #F9F4E4;
	--color-message-success: #E2F1EB;
	--color-message-info: #EBF1F7;
	--color-message-error: #FDECEE;
	--color-block-seperationheading: #F5F5F5;
	--color-block-seperationheading-dark: #DDDDDD;
	--button--default-bgcolor: #FFFFFF;
	--color-zebra-odd: #F5F5F5;
	--color-zebra-even: #fcfcfc
}
```

meinberlin/react/contrib/ControlBar.jsx
```diff
@@ -24,110 +25,127 @@ const getResultCountText = (count) => {
...
-  const handleToggleFilters = (e) => {
-    e.preventDefault()
-    setExpandFilters(!expandFilters)
+  const clearFilters = () => {
+    setQueryParams('')
+    setFilters(filters.map(f => ({ ...f, value: '' })))
   }
 
   return (
     <nav
```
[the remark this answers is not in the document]
vellip: These days it's more redundant to add that. The W3C html validator even shows a warning that it is.

meinberlin/react/contrib/ControlBar.jsx
```diff
@@ -24,110 +25,127 @@ const getResultCountText = (count) => {
...
+                  filterId={'id_' + filter.type}
+                  onSelectFilter={(choice) => handleFilterChange(filter.type, choice[0])}
+                />
+              ))
+          )}
+        </div>
+        <button onClick={() => setExpandFilters(!expandFilters)} className="control-bar__filter-btn">
+          <i className={'fa fa-chevron-' + (expandFilters ? 'up' : 'down')} aria-hidden="true" />
+          {expandFilters ? translated.hideFilters : translated.showFilters}
+        </button>
```
philli-m, line 128: this is shown when there are no additional filters to be displayed so if we only have `Status`and `Archived`it is uneeded

meinberlin/react/contrib/forms/Input.jsx
philli-m: I guess we use terney operators sometimes to avoid undefined or maybe write a function to keep the render cleaner

meinberlin/assets/scss/_form.scss
philli-m: Ok I need to check their stuff again as I thought they set their bases, I was going to restructure the frontend files and only have the old components, including this, loaded for the dashboard and only load BO and our additional style sheets for user side pages but if they haven't set up base styles that won't work, I also wonder how their styleguide is working then?!

meinberlin/assets/scss/_variables.scss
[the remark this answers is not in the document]
philli-m: I added the pink and purple as they are new for BO, I assume they will be added to their styleguide after we have finished implementation, I think for anything that is already declared we shouldn't double declare, I will try finish the restructure today and that will separate the variables and most of the components into dashboard and user loading

meinberlin/assets/scss/_form.scss
vellip: They only remove the base style for specific variants: https://styleguide.berlin.de/patterns/09-vertical_marketing-page-element-buttons/09-vertical_marketing-page-element-buttons.html look at the clean button or the close button

meinberlin/react/contrib/ControlBarSearch.jsx
philli-m: check `Searchform` in the styleguide

philli-m (commented):
@vellip 
Looking great! 
I should have mentioned that we to keep git history clean we don't have review or merge commits so if you can move changes into the commits they belong to and we rebase instead of merge dev just so the git history hopefully makes sense then it's good to go from me :)

PR #5323, opened 2023-11-20
apps/contrib: fix broken background-tasks migration

goapunk:
**Tasks**
- [ ] Changelog


m4ra (commented):
Thanks! Shall we also add a `revers_sql` that does nothing? In case we ever want to migrate to zero locally?

goapunk:
> Thanks! Shall we also add a `revers_sql` that does nothing? In case we ever want to migrate to zero locally?

Good Idea!

m4ra (approved):

PR #5325, opened 2023-11-20
Pm 2023 11 mapprio list

philli-m:
**Tasks**
- [ ] Changelog


meinberlin/apps/maptopicprio/api.py
```diff
@@ -0,0 +1,25 @@
...
+
+from meinberlin.apps.maptopicprio.models import MapTopic
+from meinberlin.apps.maptopicprio.serializers import MapTopicSerializer
+from meinberlin.apps.topicprio.api import BaseTopicViewSet
+from meinberlin.apps.topicprio.api import ModuleMixin
+from meinberlin.apps.topicprio.api import PermissionInfoMixin
+from meinberlin.apps.topicprio.api import TopicFilterInfoMixin
+
+
+class BaseMapTopicViewSet(BaseTopicViewSet):
```
goapunk, line 12: this is not needed, just add the model and the serializer_class to `MapTopicViewSet`

meinberlin/apps/maptopicprio/api.py
```diff
@@ -0,0 +1,25 @@
...
+    model: Model = MapTopic
+    serializer_class: ModelSerializer = MapTopicSerializer
+
+    def get(self, request):
```
goapunk, line 16: do we need this? this is new, right?

philli-m: I thought that what we needed to initialize the extended model? but I just got it from the internet

goapunk: don't think so unless I'm missing something.

meinberlin/apps/maptopicprio/api.py
```diff
@@ -0,0 +1,18 @@
...
+from meinberlin.apps.topicprio.api import TopicFilterInfoMixin
+
+
+class MapTopicViewSet(
+    ModuleMixin, TopicFilterInfoMixin, PermissionInfoMixin, BaseTopicViewSet
+):
+    model: Model = MapTopic
+    serializer_class: ModelSerializer = MapTopicSerializer
+
+    pass
```
goapunk, line 18: is redundant

philli-m: what is the class?!

goapunk: just the `pass` can be removed, it doesn't do anything

meinberlin/react/maptopicprio/MapTopicPrioList.jsx
```diff
@@ -0,0 +1,53 @@
+import React, { useEffect, useState } from 'react'
+import django from 'django'
+import { Card } from '../contrib/Card'
+
+export const MapTopicPrioList = (props) => {
+  const [data, setData] = useState([])
+  const [meta, setMeta] = useState([])
```
goapunk, line 7: maybe for consistency use `useState[{}]` here, as it's used as a map

meinberlin/react/maptopicprio/MapTopicPrioList.jsx
```diff
@@ -0,0 +1,53 @@
...
+  }, [])
+
+  const renderList = (data) => {
+    return (
+      <>
+        {data && data.map((topic, idx) =>
+          <li id={'item_' + topic.pk} key={'budgeting-topic-' + idx}>
+            <Card
+              item={topic}
+              permissions={meta?.permissions}
```
goapunk, line 38: `meta` can't be undefined, it's either `{}` or `{permissions: something}` so we don't need the `?` I think

goapunk (approved):
Looks good to me, just some small notes

philli-m:
@hom3mad3 @goapunk ModuleList ModuleItemList ?

goapunk:
ModuleList to me sounds like a list which would list modules. So I personally don't think it's very accurate. Maybe something like `ItemList` is more neutral ?

philli-m:
@hom3mad3 @goapunk What do we think of `CardList` as it always contains a card?

PR #5333, opened 2023-11-24
fix ckeditor forms

goapunk:
**Tasks**
- [ ] Changelog


philli-m (approved):
Working! :D good for now but something we should consider for fixes if error handling as there is no indication on the rich text field if the entry is invalid, just an error message under it which is not enough really

PR #5335, opened 2023-11-27
[7809] apps/projectcontainers: delete models and references in preparation of

goapunk:
full removal of the projectcontainers app

**Tasks**
- [x] Changelog


philli-m (changes requested):
Project container is now listed in projects and I can view it, edit view causes an error though

meinberlin/apps/ideas/models.py
```diff
@@ -130,7 +130,9 @@ class AbstractIdea(module_models.Item, Moderateable, ItemBadgesPropertyMixin):
     )
     slug = AutoSlugField(populate_from="name", unique=True)
     name = models.CharField(max_length=120, verbose_name=_("Title"))
-    description = CKEditor5Field(verbose_name=_("Description"))
+    description = CKEditor5Field(
+        verbose_name=_("Description"), config_name="collapsible-image-editor"
+    )
```
philli-m, line 135: should this be here?

meinberlin/apps/projectcontainers/forms.py
```diff
@@ -1,101 +0,0 @@
...
-            )
-            .order_by("name")
-        )
-
-    class Meta:
-        model = models.ProjectContainer
-        fields = ["projects"]
-        required_for_project_publish = ["projects"]
-        widgets = {
-            "projects": Select2MultipleWidget,
```
philli-m, line 100: we are phasing the select2 out anyway as it's not accessible but I think this is last instance of the Select2MultipleWidget so we can also delete any styling that goes with is in the select2.scss but can also be totally deleted when we remove them all

goapunk: it's still used for the categories in all modules which have them and the `map_choose_polygon_with_preset_widget` as far as I understand?

philli-m (approved):
looking great!
@m4ra I've tested it works but not sure on the code so will wait for you to check it

m4ra (approved):
LGTM

PR #5337, opened 2023-11-27
[7829] add map idea list in react

hom3mad3:
**Tasks**
- [x] Changelog


philli-m (approved):
Nice! the pagination include can go but that can happen with code clean up as possible we don't need it anymore on the user side

PR #5338, opened 2023-11-27
[7767] update userside tabs to new style

philli-m:
**Tasks**
- [x] Changelog


meinberlin/apps/projects/templates/meinberlin_projects/includes/timeline.html
```diff
@@ -1,4 +1,4 @@
-{% load contrib_tags i18n offlineevent_tags static %}
+<!-- {% load contrib_tags i18n offlineevent_tags static %}
```
philli-m, line 1: I think this was replaced a few years ago

meinberlin/assets/js/tabs.js
```diff
@@ -0,0 +1,136 @@
+/*
```
philli-m, line 1: I kind of thought this would be the best bet as it is the definition of best practices but i'm not totally sure having been a big bootstrap user previously so adding own functionality for basic things like this is kind of new, maybe you know better @hom3mad3

philli-m:
@hom3mad3 I think I will take another run at this actually as I realised there is no fallback incase of no JS, also not sure in the wcag js I used. But if you do have a suggestion about the bold/button width issue let me know!

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -9,6 +9,7 @@
 {% block extra_js %}
   {{ block.super }}
     <script type="text/javascript" src="{% static 'dsgvo_video_embed.js' %}"></script>
+    <!-- <script defer type="text/javascript" src="{% static 'tabs.js' %}"></script> -->
```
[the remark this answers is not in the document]
philli-m: oupps from the old version sorry

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -74,66 +74,61 @@ <h1 class="title">{{ project.name }}</h1>
...
-                    aria-controls="tabpanel-project-{{ project.pk }}-result"
-                    aria-expanded="false">
-                    {% translate 'Result' %}
-                </a>
-            </nav>
+            </div>
+            {% endif %}
+            <div class="span8">
+                <div role="tablist"
+                    aria-label="{% translate 'Project Details Navigation' %}">
```
philli-m, line 89: @hom3mad3 i read up this and according to mozilla we should have an aria-label for tablists if they don't have a visible heading and this is the best I could come up with?! is it descriptive enough do you think?

PR #5340, opened 2023-11-29
update pr template

hom3mad3:
**Describe your changes**
Updating our PR template for clearer communication and easier review.

At the moment, it's challenging to understand the goal of a PR and how to test it without having to ask the author. The suggested changes aim to improve the workflow for PR reviewers.

**Tasks**
- [x] Changelog

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,9 @@
+**Describe your changes**
+Briefly explain what you did and provide context for a clearer understanding.
+
 **Tasks**
 - [ ] PR name contains story or task reference
+- [ ] Steps to recreate and test the changes
 - [ ] Documentation (docs and inline)
 - [ ] Tests (including n+1 and django_assert_num_queries where applicable)
 - [ ] Changelog
--- a/changelog/_6473.md
+++ b/changelog/_6473.md
@@ -1,0 +1,5 @@
+## Changed
+
+- Updated pull request template for improved clarity and efficiency.
+- Introduced "Describe your changes" section.
+- Added items to "Tasks" checklist.


.github/pull_request_template.md
```diff
@@ -1,5 +1,11 @@
+**Describe your changes**
+Briefly explain what you did and provide context for a clearer understanding.
+
+**How to Reproduce**
+Provide steps to recreate and test the changes.
+
```
philli-m, line 6: I think this should be in the checklist as it's easy to pass over otherwise

.github/pull_request_template.md
```diff
@@ -1,5 +1,11 @@
...
 **Tasks**
-- [ ] PR name contains story or task reference
+- [ ] Short and clear PR title with story or task reference
```
philli-m, line 8: I sometimes use the story reference for searching for all PR's relating to that story and I personally think also needing to check all the task references would be annoying but maybe it makes sense for larger stories?

.github/pull_request_template.md
```diff
@@ -1,5 +1,11 @@
+**Describe your changes**
+Briefly explain what you did and provide context for a clearer understanding.
+
```
philli-m, line 3: this feels like it's doubling up, it's kind of a mix between what the commit messages should be and what information should be in the changelog, maybe we can utilise those 2 spaces better rather then adding an extra thing?  
My suggestion would be add a checklist item with 'all commits briefly describe changes made' (maybe with ref to our git guidelines? 
And maybe either on the checklist item for the changelog or in the changlog readme add a line to include the context of the change.
I'm just not sure of PR descriptions are seen after merge, whereas these things are important and should be easily viewable in either the git history or release notes which are generated from changelog

philli-m (commented):
Love the idea i just added lots of notes on the how, sorry!! also happy to go with majority on this!

.github/pull_request_template.md
goapunk: I'm also in favor of better commit messages. As for the changelog the messages can be improved but they are more of a general overview of what changed (also for other people wanting to use the software) and not a place for implementation details. Those information should either go in the commit messages or in a doc imo.

changelog/_6473.md
```diff
@@ -0,0 +1,5 @@
+## Updated
```
goapunk, line 1: we follow https://keepachangelog.com/en/1.1.0/ which doesn't know `Updated`, in this case it would be `Changed` I'd say

goapunk (approved):
Nice, looks good to me! Will leave for @m4ra to merge

philli-m (approved):
Looking good, can see if additional updates need to be made as we go

PR #5342, opened 2023-11-30
add page size query param to APIPagination

goapunk:
**Tasks**
- [x] Changelog


goapunk:
@vellip this should be enough to fetch the whole list. Let me know if you need anything else

PR #5343, opened 2023-11-30
[7274] fix ckeditor documents

goapunk:
**Tasks**
- [ ] Changelog


meinberlin/apps/documents/templates/meinberlin_documents/react_documents.html
```diff
@@ -1,5 +1,7 @@
 {% load static %}
 
+  <link href="/static/django_ckeditor_5/dist/styles.css" type="text/css" media="all" rel="stylesheet">
```
goapunk, line 3: not sure how to best include this? or should the styles be added to our scss?

meinberlin/apps/documents/templates/meinberlin_documents/react_documents.html
```diff
@@ -1,5 +1,7 @@
...
+
 {{ ckeditor_media.js }}
```
goapunk, line 5: this was here already but means that there will be a script tag just randomly in the body, should we move this somewhere else?

meinberlin/config/settings/base.py
```diff
@@ -571,6 +571,35 @@
...
+                "underline",
+                "bulletedList",
+                "numberedList",
+                "todoList",
+                "link",
+                "blockQuote",
+                "imageUpload",
+                "fileUpload",
+            ],
+            "shouldNotGroupWhenFull": "true",
```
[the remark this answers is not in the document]
goapunk: We could but I'm not sure if we want it. The grouping kind of makes sense I think? More of a ui/ux question so will leave this to you

meinberlin/apps/documents/templates/meinberlin_documents/document_dashboard.html
```diff
@@ -8,7 +8,12 @@ <h1 class="u-first-heading">{% translate 'Edit document' %}</h1>
 
 {% endblock %}
 
+{% block extra_css %}
+    {{ block.super }}
+    {{ ckeditor_media.css }}
+{% endblock extra_css %}
 {% block extra_js %}
   {{ block.super }}
+    {{ ckeditor_media.js }}
```
philli-m, line 17: why not use script tag or add comment to explain?

meinberlin/config/settings/base.py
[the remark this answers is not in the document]
philli-m: the grouping and config is exactly the same isn't it?

philli-m (commented):
can't test but go some questions

meinberlin/config/settings/base.py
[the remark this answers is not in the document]
goapunk: with  grouping I mean `shouldNotGroupWhenFull`, it will look different on small screens when this is set to `True`

meinberlin/apps/documents/templates/meinberlin_documents/document_dashboard.html
goapunk: it already comes as script tag

goapunk: I added a commit to make it clearer

meinberlin/config/settings/base.py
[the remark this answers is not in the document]
philli-m: both these editors are used in the dashboard right? so I don't know how important the usecase of initiators creating projects on mobile is? I also don't mind, just seems a lot of doubling up of code

philli-m (commented):
some comments, it does mostly work though other then www. links and the todos so fine if you want to just merge it

meinberlin/apps/documents/templates/meinberlin_documents/document_dashboard.html
goapunk: I guess mostly because if the `django-ckeditor-5` package changes the name or the path of the js files it will still work. If we hardcode the path it will break

meinberlin/config/settings/base.py
```diff
@@ -571,6 +571,35 @@
...
     },
+    "document-editor": {
+        "toolbar": {
+            "items": [
+                "bold",
+                "italic",
+                "underline",
+                "bulletedList",
+                "numberedList",
+                "todoList",
```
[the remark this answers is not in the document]
goapunk: will remove it

philli-m (approved):
Nice!

philli-m (approved):
very nice! little issue with image formatting but I will open a PR for that and find fix monday

PR #5345, opened 2023-12-04
[7809] apps/projectcontainers: split deletion migration into two separate

goapunk:
This is necessary as we can't delete the projects and the model in the same migration.

**Tasks**
- [ ] Changelog


m4ra (approved):
yeah, missed that b4. we can't know which will happen first, the data migration or the deletion

PR #5357, opened 2023-12-05
[7861] templates/*_detail: cleaning out old code and old classes from detail pages

philli-m:
**Tasks**
- [x] Changelog

Go to module and create an idea if you don't already have one locally then after creation click on the idea to see the detail view, there should only be 1 set of breadcrumbs, the old should be gone, also removed on all other list based modules (check template locations for names). To check the moderate detail view find 3 dots and click bearbeiten in the menu.


goapunk (approved):
LGTM

PR #5360, opened 2023-12-06
topics: refactor migrations to only use the code

goapunk:
- convert A4_PROJECT_TOPICS from tuple list to dict
- fix data migration

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/projects/serializers.py
+++ b/meinberlin/apps/projects/serializers.py
@@ -8,7 +8,6 @@ from rest_framework import serializers
 
 from adhocracy4.phases.models import Phase
 from adhocracy4.projects.models import Project
-from adhocracy4.projects.models import Topic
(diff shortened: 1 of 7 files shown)


meinberlin/apps/projects/serializers.py
```diff
@@ -106,6 +99,9 @@ class Meta:
             "url",
         ]
 
+    def get_topics(self, instance):
+        return [topic.code for topic in instance.topics.all()]
```
m4ra, line 103: this serializer is meant only to read existing topics of a project? not to add topics to project (write mode)? @goapunk

goapunk: Currently we only read them, or do we have code anywhere which adds them through the serializer?

PR #5361, opened 2023-12-07
Move topics to enum class

m4ra:
**Tasks**
- [x] Changelog

### DIFF
--- a/meinberlin/apps/contrib/enums.py
+++ b/meinberlin/apps/contrib/enums.py
@@ -1,0 +1,22 @@
+# fmt: off
+
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class TopicEnum(models.TextChoices):
+    """Choices for project topics."""
+
+    ANT = "ANT", _("Anti-discrimination"),
+    WOR = "WOR", _("Work & economy"),
+    BUI = "BUI", _("Building & living"),
+    EDU = "EDU", _("Education & research"),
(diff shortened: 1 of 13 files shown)


meinberlin/apps/contrib/enums.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    URB = (
+        "URB",
+        _("Urban development"),
+    )
+    ENV = (
+        "ENV",
+        _("Environment & public green space"),
+    )
+    TRA = "TRA", _("Traffic")
```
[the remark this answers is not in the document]
m4ra: found it. it was black that was changing it. Is there a reason why we have both black and flake8? @goapunk

meinberlin/apps/plans/models.py
```diff
@@ -196,9 +196,7 @@ def administrative_district(self):
 
     @property
     def topic_names(self):
-        if hasattr(settings, "A4_PROJECT_TOPICS"):
-            return [topic.name for topic in self.topics.all()]
-        return []
+        return [topic.__str__() for topic in self.topics.all()]
```
m4ra, line 199: @goapunk I will change these to str(topic) so we don't call private functions. and the __str__ one is specifically for the output of the instance when called with print() and str().

goapunk: @m4ra yes, good idea

goapunk (approved):
LGTM!

PR #5363, opened 2023-12-07
assets/ck_content: add style file for styling ck editor content as no…

philli-m:
…longer inline

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/assets/scss/components/_document.scss
+++ b/meinberlin/assets/scss/components/_document.scss
@@ -11,6 +11,7 @@
 }
 
 .paragraph__content {
+    @include clearfix;
     margin-bottom: 1px;
 }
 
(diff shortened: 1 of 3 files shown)


meinberlin/assets/scss/components/_document.scss
```diff
@@ -11,6 +11,7 @@
 }
 
 .paragraph__content {
+    @include clearfix;
```
philli-m, line 14: @goapunk it would be better if we could give all ckeditor rendered content a class that we could assign this to?

goapunk (changes requested):
I think this can be solved by configuring ckeditor better. The default image mode is block, not inline as we want it. lets have a look at it together

goapunk:
@philli-m will close this as we solved it with the config

PR #5368, opened 2023-12-08
settings/base: update ckeditor5 config

goapunk:
- remove blockquote as we don't have styling for it
- make image insertion smarter, which allows inserting images inline instead of always as block
- fix broken list properties
- remove unused customColorPalette

**Tasks**
- [ ] Changelog


m4ra (approved):
looks good

PR #5371, opened 2023-12-14
[7884] add pyenv support for make clean 

m4ra:
**Tasks**
- [ ] Changelog


m4ra:
@hom3mad3 can you test if that works locally for you? When you run `make clean` it should delete all python deps.
Check you have things installed with `pip freeze` then run the clean command, then another `pip freeze` that should return empty.

PR #5372, opened 2023-12-14
[7806] plans and projects api caching

m4ra:
**Tasks**
- [x] Changelog

### DIFF
--- a/docs/api_caching.md
+++ b/docs/api_caching.md
@@ -1,0 +1,37 @@
+## Background
+
+We have noticed that the page load of `mein.berlin.de/projekte/` is pretty slow with about 6s for 550 projects. Three API calls are particularly slow:
+
+- https://mein.berlin.de/api/projects/?status=pastParticipation		2.811s
+- https://mein.berlin.de/api/plans/					                      3.613s
+- https://mein.berlin.de/api/extprojects/				                  5.041s
+
+These paths correspond to the following api views:
+
+- `projects/api.py::ProjectListViewSet`
+- `plans/api.py::PlansListViewSet`
+- `extprojects/api.py::ExternalProjectListViewSet`
+
+we decided to start caching the endpoints with a redis backend.
+
+## Developer Notes
+
+The cache target is the `list` method of the following views:
+
+- `ExternalProjectListViewSet`
+- `PlansListViewSet`
+- `ProjectListViewSet`
+- `PrivateProjectListViewSet`
+
+Cache keys expire after a timeout (default value 1h) or if a context specific signal is received (e.g. cache keys for projects are deleted if the signal for a saved project is detected).
+
+The cache keys for projects are constructed by the view namespace and their status if exists:
+- `projects`
+- `projects_activeParticipation`
+- `projects_pastParticipation`
+- `projects_futureParticipation`
+- `privateprojects`
+- `extprojects`
+- `plans`
+
+In production, we use django's built-in [Redis](https://docs.djangoproject.com/en/4.2/topics/cache/#redis) as cache backend (see `settings/production.py::CACHES`). For development and testing the cache backend is the default, that is [local memory](https://docs.djangoproject.com/en/4.2/topics/cache/#local-memory-caching). If you want to enable redis cache for local development, then copy the production settings to your `settings/local.py`.
--- a/docs/api_caching.md
+++ b/docs/api_caching.md
@@ -12,7 +12,7 @@ These paths correspond to the following api views:
 - `plans/api.py::PlansListViewSet`
 - `extprojects/api.py::ExternalProjectListViewSet`
 
-we decided to start caching the endpoints with a redis backend.
+we decided to start caching the endpoints with redis.
 
 ## Developer Notes
 
@@ -26,7 +26,6 @@ The cache target is the `list` method of the following views:
 Cache keys expire after a timeout (default value 1h) or if a context specific signal is received (e.g. cache keys for projects are deleted if the signal for a saved project is detected).
 
 The cache keys for projects are constructed by the view namespace and their status if exists:
-- `projects`
 - `projects_activeParticipation`
 - `projects_pastParticipation`
 - `projects_futureParticipation`
@@ -34,4 +33,28 @@ The cache keys for projects are constructed by the view namespace and their stat
 - `extprojects`
 - `plans`
 
+## Celery tasks
+
+A periodic task checks for projects that will become either active or past in the next 10 minutes.
+- schedule_reset_cache_for_projects()
+
+In case of projects becoming active the cache is cleared for:
+- `projects_activeParticipation`
+- `projects_futureParticipation`
+- `privateprojects`
+- `extprojects`
+
+in case of projects becoming past the cache is cleared for:
+- `projects_activeParticipation`
+- `projects_pastParticipation`
+- `privateprojects`
+- `extprojects`
+
 In production, we use django's built-in [Redis](https://docs.djangoproject.com/en/4.2/topics/cache/#redis) as cache backend (see `settings/production.py::CACHES`). For development and testing the cache backend is the default, that is [local memory](https://docs.djangoproject.com/en/4.2/topics/cache/#local-memory-caching). If you want to enable redis cache for local development, then copy the production settings to your `settings/local.py`.
+
+files: 
+- `./meinberlin/apps/plans/api.py`
+- `./meinberlin/apps/extprojects/api.py`
+- `./meinberlin/apps/projects/api.py`
+- `./meinberlin/apps/projects/tasks.py`
+- `./meinberlin/config/settings/production.py`
--- a/meinberlin/apps/extprojects/api.py
+++ b/meinberlin/apps/extprojects/api.py
@@ -1,5 +1,7 @@
+from django.core.cache import cache
 from django.utils import timezone
 from rest_framework import viewsets
+from rest_framework.response import Response
 
 from adhocracy4.projects.enums import Access
 from meinberlin.apps.extprojects.models import ExternalProject
@@ -18,3 +20,13 @@ class ExternalProjectListViewSet(viewsets.ReadOnlyModelViewSet):
     def get_serializer(self, *args, **kwargs):
         now = timezone.now()
         return ExternalProjectSerializer(now=now, *args, **kwargs)
+
+    def list(self, request, *args, **kwargs):
+        data = cache.get("extprojects")
+        if data is None:
+            queryset = self.filter_queryset(self.get_queryset())
+            serializer = self.get_serializer(queryset, many=True)
+            data = serializer.data
+            cache.set("extprojects", data)
+
+        return Response(data)
--- a/meinberlin/apps/extprojects/signals.py
+++ b/meinberlin/apps/extprojects/signals.py
@@ -1,4 +1,5 @@
 from django.core.cache import cache
+from django.db.models.signals import post_delete
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
@@ -6,5 +7,6 @@ from .models import ExternalProject
 
 
 @receiver(post_save, sender=ExternalProject)
-def reset_cache(sender, instance, update_fields, **kwargs):
+@receiver(post_delete, sender=ExternalProject)
+def reset_cache(sender, instance, *args, **kwargs):
     cache.delete("extprojects")
--- a/meinberlin/apps/extprojects/signals.py
+++ b/meinberlin/apps/extprojects/signals.py
@@ -1,0 +1,10 @@
+from django.core.cache import cache
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import ExternalProject
+
+
+@receiver(post_save, sender=ExternalProject)
+def reset_cache(sender, instance, update_fields, **kwargs):
+    cache.delete("extprojects")
--- a/meinberlin/apps/plans/signals.py
+++ b/meinberlin/apps/plans/signals.py
@@ -1,12 +1,12 @@
 from django.core.cache import cache
+from django.db.models.signals import post_delete
 from django.db.models.signals import post_save
-from django.db.models.signals import pre_delete
 from django.dispatch import receiver
 
 from .models import Plan
 
 
 @receiver(post_save, sender=Plan)
-@receiver(pre_delete, sender=Plan)
-def reset_cache(sender, instance, update_fields=None, **kwargs):
+@receiver(post_delete, sender=Plan)
+def reset_cache(sender, instance, *args, **kwargs):
     cache.delete("plans")
--- a/meinberlin/apps/plans/signals.py
+++ b/meinberlin/apps/plans/signals.py
@@ -1,0 +1,12 @@
+from django.core.cache import cache
+from django.db.models.signals import post_save
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
+
+from .models import Plan
+
+
+@receiver(post_save, sender=Plan)
+@receiver(pre_delete, sender=Plan)
+def reset_cache(sender, instance, update_fields=None, **kwargs):
+    cache.delete("plans")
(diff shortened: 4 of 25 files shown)


meinberlin/config/settings/production.py
```diff
@@ -1,6 +1,14 @@
 from .base import *
 
 DEBUG = False
+
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.redis.RedisCache",
```
goapunk, line 7: We'll also need to add this to `travis.py` otherwise it will use the `LocMemCache` from below

meinberlin/apps/extprojects/api.py
```diff
@@ -18,3 +20,13 @@ def get_queryset(self):
     def get_serializer(self, *args, **kwargs):
         now = timezone.now()
         return ExternalProjectSerializer(now=now, *args, **kwargs)
+
+    def list(self, request, *args, **kwargs):
+        data = cache.get("extprojects")
+        if data is None:
+            queryset = self.filter_queryset(self.get_queryset())
+            serializer = self.get_serializer(queryset, many=True)
+            cache.set("extprojects", serializer.data, 360)
```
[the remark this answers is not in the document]
goapunk: One thing we discussed on the old PR: if a project starts at `12:00` and the cache is set at `11:59` the project won't appear in the list for `cache_time` -1 minutes which is not ideal. I think we considered using celery-beat to periodically invalidate the cache or setting the cache_time to the time of the earliest project which is about to start

goapunk (changes requested):
Some initial comments and the question about how we handle cache expiry times

meinberlin/apps/extprojects/api.py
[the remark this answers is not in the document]
m4ra: Let's talk about it tmr after standup. For instance, a periodic task that checks whether a project starts in the next 5 mins? running every 5 mins? if yes. the task clears the cache for these type of projects?
@goapunk

meinberlin/apps/projects/signals.py
```diff
@@ -20,3 +20,5 @@ def post_save_delete(sender, instance, update_fields, **kwargs):
     cache.delete_many(
         ["projects_active", "projects_future", "projects_past", "private_projects"]
     )
+    # TODO: add redis key for earliest next project starts
+    # TODO: add redis key for earliest next project ends
```
m4ra, line 24: @goapunk here I will fetch the earliest start and latest end date from phase objects and set them in redis db

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,72 @@
...
+@shared_task
+def reset_cache_for_projects(ends=False, starts=False):
+    # cache set projects
+    if starts:
+        # remove redis key next_project_start
+        redis_db.delete("next_project_starts")
+    if ends:
+        # remove redis key next_project_ends
+        redis_db.delete("next_project_ends")
+
```
m4ra, line 72: @goapunk have a look and we can discuss it again TMR. The logic is to check every 10 min two things:
one task for future and one for past projects:  if the earliest future/latest past project is happening in the next 10 mins, and if yes then we call the actual reset cache task with a countdown. If not, then one of the next callings of the periodic tasks will do it.

We only query the DB when there is no redis key set, and when a project is saved.

goapunk: structure looks good to me, left a small question but I think this will be a nice solution

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,72 @@
...
+from meinberlin.apps import logger
+
+redis_db = redis.Redis()
+
+
+@shared_task  # periodic task every 10min
+def schedule_reset_cache_for_future_projects():
+    # fetch earliest future project
+    # check if redis has a next_project_starts key
+    next_project_starts = redis_db.get("next_project_starts")
```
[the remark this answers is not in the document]
m4ra: I thought to not mess with the cache, in case of cache.clear(). But prob doesn't matter, I will use cache.

m4ra:
@goapunk have another look when you can. I need to add a test, e.g adding future projects and check if the are added in the cache.

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,107 @@
...
+            will be removed from cache in {remain_time} min"""
+        )
+        return msg
+
+
+@shared_task
+def reset_cache_for_projects(starts, ends):
+    if starts:
+        # remove redis key next_project_start
+        cache.delete("next_project_starts")
```
goapunk, line 104: we still need to actually delete the cache here, right?

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,107 @@
...
+        A string message if the next project starts
+        date will be removed from the cache.
+    """
+
+    # check if redis has a next_project_starts key
+    next_project_starts = cache.get("next_project_starts")
+
+    if not next_project_starts:
+        # set the redis key: value
+        cache.set("next_project_starts", get_next_project_starts())
```
goapunk, line 57: do we need to set `next_project_starts` here? otherwise it will be `None` below

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,107 @@
...
+        msg = logger.info(
+            f"""next project ends date {next_project_ends}
+            will be removed from cache in {remain_time} min"""
+        )
+        return msg
+
+
+@shared_task
+def reset_cache_for_projects(starts, ends):
+    if starts:
```
goapunk, line 102: Do we have to call schedule_reset_cache_for_projects_* here again or run this for all projects starting within 10 minutes ? Considering the following case:
1. Project A starts at 12:01, Project B starts at 12:03
2. The schedule task runs at 12:00 and now schedules a reset_cache_for_projects for 12:01
3. At 12:01 the cache is reset and now includes Project A
4. At 12:03 Project B won't be visible yet because schedule_reset only runs at 12:10
5. At 12:10 schedule_reset runs again, if there's no future project starting the cache won't be updated and Project B still won't be visible, if there is a future Project C, Project B will only become visible once Project C starts

This is more complicated then I hoped :/ Not sure how to make this bulletproof. Maybe we can have a look together?

meinberlin/apps/projects/tasks.py
m4ra: line 57 does that, setting the 'next_project_starts'.

meinberlin/apps/projects/tasks.py
m4ra: Quite a niche case, but valid point nonetheless. I will check for a list of future projects that are to be published in the next 10 mins with the function `get_next_project_starts()` and apply the cache reset for that list of projects.

meinberlin/apps/extprojects/signals.py
```diff
@@ -0,0 +1,10 @@
+from django.core.cache import cache
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import ExternalProject
+
+
+@receiver(post_save, sender=ExternalProject)
```
goapunk, line 8: should we add `post_delete` here as well?

meinberlin/apps/plans/signals.py
```diff
@@ -0,0 +1,12 @@
+from django.core.cache import cache
+from django.db.models.signals import post_save
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
+
+from .models import Plan
+
+
+@receiver(post_save, sender=Plan)
+@receiver(pre_delete, sender=Plan)
```
goapunk, line 10: should we use `post_delete` instead of `pre_delete? seems to be more logical

meinberlin/apps/projects/signals.py
```diff
@@ -0,0 +1,49 @@
...
+            "projects_future*",
+            "projects_past*",
+            "private_projects",
+            "extprojects",
+        ]
+    )
+
+
+@receiver(post_save, sender=Project)
+@receiver(pre_delete, sender=Project)
```
goapunk, line 28: `should we use post_delete instead of `pre_delete? seems to be more logical

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,169 @@
...
+                "%Y-%m-%d, %H:%M:%S %Z"
+            )
+            list_format_phases.append(str_phase)
+    return list_format_phases
+
+
+def get_next_project_ends() -> str:
+    """
+    Helper function to query the db and
+    retreive the erliest phase that will end.
```
goapunk, line 41: typo: erliest -> earliest

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,169 @@
...
+        cache.delete("next_project_starts")
+        cache.delete_many(
+            [
+                "projects_activeParticipation",
+                "projects_futureParticipation",
+                "private_projects",
+                "extprojects",
+            ]
+        )
+        if cache.get("projects_*") or cache.get("extprojects"):
```
goapunk, line 150: `cache.get("projects_*")` will only work with redis, not with any of the other caching backends, right?

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,169 @@
...
+
+@shared_task
+def reset_cache_for_projects(starts, ends):
+    msg = "Clear cache "
+    if starts:
+        # remove redis key next_project_start
+        cache.delete("next_project_starts")
+        cache.delete_many(
+            [
+                "projects_activeParticipation",
```
goapunk, line 144: above we use other keys:
```
  cache.delete_many(
        [
            "projects_active*",
            "projects_future*",
            "projects_past*",
            "private_projects",
            "extprojects",
        ]
    )
```
i assume we need the same here ? :)

meinberlin/apps/projects/signals.py
```diff
@@ -0,0 +1,49 @@
...
+def post_save_delete(sender, instance, update_fields=None, **kwargs):
+    """
+    Delete cache for project list views.
+    Capture any new phases that may got created/updated while saving a project.
+    """
+
+    cache.delete_many(
```
goapunk, line 35: below and in the docs it says the keys are:
```
       cache.delete_many(
            [
                "projects_activeParticipation",
                "projects_futureParticipation",
                "private_projects",
                "extprojects",
            ]
        )
```
I assume you forgot to change it here ? :)

meinberlin/apps/projects/signals.py
```diff
@@ -0,0 +1,49 @@
...
+from meinberlin.apps.projects.tasks import get_next_projects_start
+
+
+@receiver(a4dashboard_signals.project_created)
+@receiver(a4dashboard_signals.project_published)
+@receiver(a4dashboard_signals.project_unpublished)
+def post_dashboard_signal_delete(sender, project, user, **kwargs):
+    cache.delete_many(
+        [
+            "projects_active*",
```
goapunk, line 18: do we need the globbing here? the wildcard will only match `projects_activeParticipation`, right ? And the wildcard will make it incompatible with non-redis cache backends

meinberlin/apps/projects/tasks.py
goapunk: hmm, I can't find any information on this though. The django docs don't mention wildcards (or I'm too stupid to find them), only the django-redis package mentions it. However if we want to be on the safe side I think we can use `cache.get_many(['projects_activeParticipation', 'projects_pastParticipation', ... , 'extprojects',])`

meinberlin/config/settings/production.py
```diff
@@ -30,3 +39,16 @@
     CKEDITOR_CONFIGS["video-editor"]["embed_provider"] = CKEDITOR_URL
 except NameError:
     pass
+
+CELERY_BEAT_SCHEDULE = {
+    "update-future-projects-every-10-mim": {
+        "task": "schedule_reset_cache_for_projects_becoming_active",
+        "schedule": crontab(minute="*/10"),
```
goapunk, line 46: `crontab` is missing an import here : `from celery.schedules import crontab`

goapunk (commented):
Nice, looks good to me so far, I left some comments with questions. I'm not sure about the failing tests, seems to be a bit tricky with the combination of redis + `CELERY_TASK_ALWAYS_EAGER=True`. Without it we need a running celery worker though

m4ra:
@goapunk there was a typo in the reset cache task of the key to be deleted. As I was testing locally and because of the asterisks I was using, something was not captured. Anyhow I simplified the tests and removed the API calls as these are part of the `test_api_caching`. Also I ended up using a 10min check and a list of projects that will become past same as for future projects, otherwise we may miss any other expiring between the 10mins if one is already set for expiration. Looking again at the tasks and functions I wonder if they should be consolidate instead of having separated for future and past projects.

goapunk:
> @goapunk there was a typo in the reset cache task of the key to be deleted. As I was testing locally and because of the asterisks I was using, something was not captured. Anyhow I simplified the tests and removed the API calls as these are part of the `test_api_caching`. Also I ended up using a 10min check and a list of projects that will become past same as for future projects, otherwise we may miss any other expiring between the 10mins if one is already set for expiration. Looking again at the tasks and functions I wonder if they should be consolidate instead of having separated for future and past projects.

I guess the only difference is the deletion of `projects_futureParticipation` vs `projects_pastParticipation`, so yeah, maybe we just delete them both and consolidate the code? What do you think?

m4ra:
@goapunk I merged the two scheduled tasks, the actual resets for future/past projects i left as it is, since it will save us from queering either future or past projects again via the url, which are time/power consuming.
I added also django query checks in the tests, very helpful and eye opening, as without cache, the api endpoint queries 12 times (with the prefetch-related).

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,192 @@
...
+                "projects_futureParticipation",
+                "private_projects",
+                "extprojects",
+            ]
+        )
+        if cache.get_many(
+            [
+                "projects_activeParticipation",
+                "projects_futureParticipation" "private_projects",
```
goapunk, line 163: is there a comma  + newline missing between the two values?

meinberlin/apps/projects/tasks.py
```diff
@@ -0,0 +1,192 @@
...
+                "projects_activeParticipation",
+                "projects_pastParticipation",
+                "private_projects",
+                "extprojects",
+            ]
+        )
+        if cache.get_many(
+            [
+                "projects_activeParticipation",
+                "projects_pastParticipation" "private_projects",
```
goapunk, line 184: is there a comma + newline missing between the two values?

docs/api_caching.md
```diff
@@ -0,0 +1,63 @@
...
+- `extprojects`
+- `plans`
+
+## Celery tasks
+
+A periodic task checks for projects that will become either active or past in the next 10 minutes.
+- schedule_reset_cache_for_projects()
+
+In case of projects becoming active the cache is cleared for:
+- `projects`
```
goapunk, line 43: we don't clear this key anywhere, right?

docs/api_caching.md
```diff
@@ -0,0 +1,63 @@
...
+- `projects_activeParticipation`
+- `projects_futureParticipation`
+- `privateprojects`
+- `extprojects`
+
+in case of projects becoming past the cache is cleared for:
+- `projects`
```
goapunk, line 50: we don't clear this key anywhere, right?

goapunk (changes requested):
Really nice! Just some small comments left

goapunk:
> @goapunk I merged the two scheduled tasks, the actual resets for future/past projects i left as it is, since it will save us from queering either future or past projects again via the url, which are time/power consuming. I added also django query checks in the tests, very helpful and eye opening, as without cache, the api endpoint queries 12 times (with the prefetch-related).

@m4ra yeah, the query check is really helpful, we definitely need to work on improving the queries/structure. Ideally we will have checks in all our tests in the future :crossed_fingers:

goapunk (approved):
Really cool :) Before we deploy it we need to make some changes in the admin repo

PR #5373, opened 2023-12-14
fix ckeditor content styling and more config fixes

goapunk:
content

**Tasks**
- [ ] Changelog


m4ra:
@goapunk what should we test here when running the PR locally?

goapunk:
@m4ra good question, should have added some context, sorry. I think what should be tested is that content created via the ckeditor is styled correctly in all places we use it. Not sure if every place has to be tested though, as this is mostly config changes and adding the ckeditor css

philli-m (approved):
Looking good!

PR #5385, opened 2024-01-03
redesign phase info component

hom3mad3:
# Redesign Phase Info Component

## Test Page
`projekte/module/module-title` (e.g., [Participatory Budgeting Phase 3](http://127.0.0.1:8003/projekte/module/participatory-budgeting-3-phase-3/))

## Changes
- Implemented redesign for the Phase Info Panel component. Uses [Servicepanel](https://styleguide.berlin.de/patterns/09-vertical_marketing-page-modul-servicepanel/09-vertical_marketing-page-modul-servicepanel.html) and custom style overwrites for the swiper.
- Enhanced for both desktop and mobile views with Swiper integration.

## Previews
- Desktop: PhaseInfo-Desktop
- Mobile: <div></div>

**Tasks**
- [x] Changelog

### DIFF
--- a/meinberlin/assets/js/swiper_phases.js
+++ b/meinberlin/assets/js/swiper_phases.js
@@ -1,5 +1,12 @@
-import Swiper from 'swiper/bundle'
-import 'swiper/css/bundle'
+import Swiper from 'swiper'
+
+// Needed due to issue with Swiper, issue link: https://github.com/import-js/eslint-plugin-import/issues/2266
+/* eslint-disable import/no-unresolved */
+import { Navigation, Pagination } from 'swiper/modules'
+import 'swiper/css'
+import 'swiper/css/navigation'
+import 'swiper/css/pagination'
+/* eslint-enable import/no-unresolved */
 
 const createSwiper = ({ rootElement, options }) =>
   new Swiper(rootElement, options)
@@ -10,6 +17,7 @@ const initSwiper = () => {
     options: {
       direction: 'horizontal',
       loop: true,
+      modules: [Navigation, Pagination],
       navigation: {
         nextEl: '.swiper-button-next',
         prevEl: '.swiper-button-prev'
@@ -26,7 +34,7 @@ const initSwiper = () => {
   }
 
   createSwiper(config)
-};
+}
 
 (function () {
   // Check if Swiper is already available globally
--- a/meinberlin/assets/js/swiper_phases.js
+++ b/meinberlin/assets/js/swiper_phases.js
@@ -1,8 +1,9 @@
+import Swiper from 'swiper/bundle'
+import 'swiper/css/bundle'
+
 const createSwiper = ({ rootElement, options }) =>
-// eslint-disable-next-line no-undef
   new Swiper(rootElement, options)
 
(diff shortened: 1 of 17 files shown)


philli-m:
> @goapunk @philli-m @hom3mad3 i have absolutely no idea how and where to test this, as EverythingWorksOnMyMachine™ but tests are failing :D

@hom3mad3 I'm not sure what this comment is concerning? should we have a look at something specifically?

philli-m:
@hom3mad3  hmm that's probably because we no longer install swiper as the BO styleguide to avoid a doubling up, but it is weird it's not failing locally, I'll have a look, maybe we need to import it differently or something

philli-m:
@hom3mad3 checking this i'm still not sure why the additional swiper.js is required? the swiper should be initialized by the BO js, is it just the style overwrites? maybe we can talk tomorrow?

meinberlin/assets/scss/style_user_facing.scss
```diff
@@ -50,7 +50,7 @@
 @import "../berlin_css/berlin_marketing";
 
 // extra scss / overwrites
-@import "../extra_css/swiper.scss";
+@import "../extra_css/swiper";
```
philli-m, line 53: removed as didn't pass linter, not sure why there wasn't an error when commiting, maybe we need to check out commit linting?

meinberlin/assets/js/swiper_phases.js
```diff
@@ -26,7 +34,7 @@ const initSwiper = () => {
   }
 
   createSwiper(config)
-};
+}
```
philli-m, line 37: also lint error

meinberlin/assets/js/swiper_phases.js
```diff
@@ -1,5 +1,12 @@
...
-import 'swiper/css/bundle'
+import Swiper from 'swiper'
+
+// Needed due to issue with Swiper, issue link: https://github.com/import-js/eslint-plugin-import/issues/2266
+/* eslint-disable import/no-unresolved */
+import { Navigation, Pagination } from 'swiper/modules'
+import 'swiper/css'
+import 'swiper/css/navigation'
+import 'swiper/css/pagination'
+/* eslint-enable import/no-unresolved */
```
philli-m, line 9: This is apparently a known issue, have added link, also just did a little update to only import the used modules as the bundle is a bit big

philli-m:
> > @hom3mad3 checking this i'm still not sure why the additional swiper.js is required? the swiper should be initialized by the BO js, is it just the style overwrites? maybe we can talk tomorrow?
> 
> this was the only way i could find to be able to style the swiper pagination and make it accessible, otherwise i cannot access the Swiper object which as been instantiated by the BO component and add customizations. i removed the modul-buhne element which initializes the swiper and was causing the issue.

fair play, might make the liberary integration a bit of a pain but I am hoping that will not be our job :p 
It's looking good! I fixed the linting so the tests should pass now, just to check do you get errors when you commit? the staged linting should have picked up on that but it can break sometimes :(

Happy to merge when tests pass, just left some comments as change the import a bit, let me know if that's ok?!

PR #5399, opened 2024-01-11
[#7768] redesign registration and login

hom3mad3:
## **Describe your changes**
Implement new designs for login and registration pages according to new styleguide

### **Login:** [http://localhost:8003/accounts/login/](http://localhost:8003/accounts/login/)
anmeldung

### **Signup:** [http://localhost:8003/accounts/signup/](http://localhost:8003/accounts/signup/)
register

**Tasks**
- [ ] Update Translations
- [x] Changelog


meinberlin/templates/account/login.html
```diff
@@ -1,44 +1,51 @@
...
-
-        {% if redirect_field_value %}
-            <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}"/>
+    <div class="account__wrapper">
+        <h1>{% translate "Login" %}</h1>
+
+        <form method="post" action="{% url 'account_login' %}" class="form--base">
+            {% if form.non_field_errors %}
+                <ul class="errorlist" aria-live="assertive" aria-atomic="true">
+                    {% for error in form.non_field_errors %}
```
goapunk, line 15: I couldn't find anything in the designs regarding the non-field error messages but should they also use the `message message--error` class? Currently they are shown as a list item: 
 * Der Anmeldename und/oder das Passwort sind leider falsch.

goapunk (changes requested):
Looks mostly good to me, the pages look great. Regarding the styling there are two little issues with the error messages. For the form refactoring I'm not sure as I'm not too much into the redesign process but since they are used everywhere they also break some things in the dashboard (helptexts are now below the fields, some styles are broken). Maybe we need different templates for user and dashboard side?

meinberlin/apps/contrib/templates/meinberlin_contrib/includes/form_checkbox_field.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n widget_tweaks %}
 
 <div class="form-check">
-    <label class="form-check__label">
+    <label class="form-check-label">
```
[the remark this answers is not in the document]
goapunk: seems like a good solution for now!

meinberlin/apps/contrib/templates/meinberlin_contrib/includes/form_field.html
```diff
@@ -1,26 +1,33 @@
...
+
     {% block field %}
         <div class="widget widget--{{ field|widget_type }}">
             {{ field|add_error_attr:"aria-invalid:true" }}
         </div>
-    {% endblock %}
+    {% endblock field %}
+    
+
+    {% if field.help_text %}
```
goapunk, line 20: This will move the helptexts for all fields in the dashboard and break the styling, so either we also copy this file or at least add the `form-hint` class back in so the styling is still there.

goapunk (changes requested):
Looks mostly good, just the remark about the `form_field.html`. On an unrelated note: do we still want the placeholder texts? I feel like we wanted to stop using them but can't really remember. Doesn't need to be part of this story in any case

meinberlin/apps/contrib/templates/meinberlin_contrib/includes/form_field.html
goapunk: sorry, could have added a link: check http://127.0.0.1:8003/dashboard/organisations/liqd/plans/create/

goapunk: yeah, it was nice while we had the same styles everywhere but now with the redesign they are more of a burden. I think having one for user_facing and one for dashboard takes away the complexity for now and is easy to do

goapunk:
> @goapunk i would suggest we merge this if the login and registration pages are ok with you. i would like to tackle unrelated issues on a different task as not to increase the scope of this task too much.

@hom3mad3 let's merge it then. I think we should try to fix the broken form_field as soon as possible, otherwise we'll just forget. I can work on it if you want

goapunk (approved):
LGTM

Issue #5370, opened 2023-12-13
#7702 [mB] change design of React proposal list to BO card - missing pluralization of comment/ like counts 


sblindow:
**URL:** 
[Example-url](https://meinberlin-design-dev.liqd.net/projekte/module/ideensammlung-mit-karte-5/?mode=list)

**user:** 
All

**expected behaviour:** 
The application should dynamically adjust the word to its singular or plural form based on the count

For example, "1 Kommentar", "2 Kommentare", "1 Like", "0 Likes", etc.
 
**behaviour:** 
The terms indicating the count (comments, likes, dislikes) are static and do not change according to the number, leading to grammatical inaccuracies.

**important screensize:**
Not relevant

**device & browser:** 
All devices/ browsers

**Screenshots:** 

<img width="1001" alt="Bildschirmfoto 2023-12-13 um 10 57 56" src="https://github.com/liqd/a4-meinberlin/assets/144320958/be210d59-398c-444b-930f-2f8877be691a">

<img width="1003" alt="Bildschirmfoto 2023-12-06 um 13 57 37" src="https://github.com/liqd/a4-meinberlin/assets/144320958/b8dad537-1982-4f43-bfdc-410765528d8a">

philli-m:
NOTE: comment and vote strings should change according to the code however possible this is a translation issue, plural function should be added to the like and dislike

PR #5407, opened 2024-01-22
react/CardStats: update stat strings to include plural versions fixes…

philli-m:
… #5370 and fix tests

**Describe your changes**

**Tasks**
- [ ] Changelog


philli-m:
eeek, just checked and looks like the django pluralisation is broken in comments contribution string too, will have a look, might have been django update :(

philli-m:
@hom3mad3 I checked and it's weird translation issue, because we have a non pluralised version of 'Comment' in a4, the translations get confused, you can only see the error when you deleted the locale files and re-make po. Luckily we don't use that string it seems so I have removed it in this [PR](https://github.com/liqd/adhocracy4/pull/1518/files), once thats merged and we use the new a4 the pluralisation should work.
Contributions plurals in comments is a different issue but it does work after page reload, not ideal but not terrible

philli-m:
NOTE: didn't update the a4 hash as this will be done with rebase but after the comment string will be fixed

philli-m:
@hom3mad3 this should now be fully working with updated a4

Issue #5366, opened 2023-12-07
UI/UX Test - #7701 [mB] replace project hero with image with aside


mcastro-lqd:
**URL:** https://meinberlin-design-dev.liqd.net/
**important screensize:** *Both*
**device & browser:** *Chrome and Safari*
**Comment/Question:** *
Spacing here between Title and Subhead is a bit too much compared to the design. Can we lower a bit? And it’s also missing the Headerbild Urheber*in on the side that was on Zeplin when an Image is Added. 
<img width="1784" alt="Screenshot 2023-12-07 at 16 42 52" src="https://github.com/liqd/a4-meinberlin/assets/59610786/d20901eb-4c61-4a43-940d-3e6acc24f34a">

Is it possible to have the image cropped full width here instead of the black spaces on the sides? 
![IMG_1397](https://github.com/liqd/a4-meinberlin/assets/59610786/a6717214-059d-4ee3-9274-ac1fb4184e6a)

Cheers,

*

philli-m:
@mcastro-lqd for the cropping issue I'm not sure as link isn't for a project but I think the screenshot you sent, the image does have black edges is it this ones? https://meinberlin-design-dev.liqd.net/projekte/test-modules/

CarolingerSeilchenspringer:
@mcastro-lqd can you check above question?

mcastro-lqd:
@philli-m @CarolingerSeilchenspringer Omg yes, forget about that then! Sorryyy

PR #5404, opened 2024-01-22
assets/herounit-image-with-aside: add title styling and fix copyright

philli-m:
…d more specific target for copyright fixes #5366

**Describe your changes**
Briefly explain what you did and provide context for a clearer understanding.

**Tasks**
- [x] PR name contains story or task reference
- [x] Steps to recreate and test the changes
- [ ] Documentation (docs and inline)
- [ ] Tests (including n+1 and django_assert_num_queries where applicable)
- [ ] Changelog

fix can be seen on project detail http://localhost:8003/projekte/liqd-project/ if project has a header image with copyright

fixed image copyright so can be seen issue - needs image and copyright #5366 

expected outcome:
screenshot(29)

PR #5409, opened 2024-01-22
hero-unit: fix redesign issues

hom3mad3:
## hero-unit: redesign fixes

- fixes issue: https://github.com/liqd/a4-meinberlin/issues/5366 
- adds max-width to container to fit design
- fixes follow button icon color

hero-unit-fix

possible duplicate of: https://github.com/liqd/a4-meinberlin/pull/5406 and https://github.com/liqd/a4-meinberlin/pull/5404

**Tasks**
- [ ] Changelog


meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -48,22 +48,20 @@
             <div class="mainbar__left">
                 <div class="image">
                     <div class="image__image">
-                        <img src="{{ project.image |thumbnail_url:'heroimage' }}" alt="{% if project.image_alt_text %}{{ project.image_alt_text }}{% else %}{% translate 'Here you can find a decorative picture.' %}{% endif %}" />
+                        <img src="{{ project.image | thumbnail_url:"heroimage" }}" alt='{% if project.image_alt_text %}{{ project.image_alt_text }}{% else %}{% translate "Here you can find a decorative picture." %}{% endif %}' width="100%" height="auto" loading="lazy"/>
```
philli-m, line 51: the linting here is a little off, I assume you'll get errors too if you make changes and try to commit but it should list all the changes you need to make, annoyingly the fix command does not deal with most of them :(

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -48,22 +48,20 @@
             <div class="mainbar__left">
                 <div class="image">
                     <div class="image__image">
-                        <img src="{{ project.image |thumbnail_url:'heroimage' }}" alt="{% if project.image_alt_text %}{{ project.image_alt_text }}{% else %}{% translate 'Here you can find a decorative picture.' %}{% endif %}" />
```
philli-m, line 51: I can't add a comment above but on line 43 there's a `FIXME`statement concerning where `layout-grid__area--maincontent` should be and it seems like they have fixed the issue in their last update so that can now be moved under the  `herounit-image-with-aside`element

philli-m (commented):
coolio! nice catch on the width issue, that passed me by, just a note about class names and the linting, but soon it will all be done! (hopefully)

philli-m:
Oh and no double up with the tabs [PR](https://github.com/liqd/a4-meinberlin/pull/5406) by the way, thought I wrote that somewhere but it seems lost now?!

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
philli-m: I actually updated the linting rules to reduce them a bit so if you rebase i think it won't complain, at least not as much

meinberlin/assets/scss/components_user_facing/_follow.scss
```diff
@@ -1,10 +1,16 @@
+.follow__container > span {
+    width: 100%;
+}
```
philli-m, line 3: @hom3mad3 this span is the data widget from the template ta hence no class and as it is an inline element it had a limited width (i think anyway) so it was also limiting the nested elements

philli-m (approved):
perfect! :) will approve and should auto merge after CI

Issue #5367, opened 2023-12-07
UI/UX Test - #7703 [mB] change design of ampel on proposal list


mcastro-lqd:
**URL:** https://meinberlin-design-dev.liqd.net/
**important screensize:** *Both*
**device & browser:** *Chrome and Safari*
**Comment/Question:** *
Here the colours seems the old ones, the ones from [berlin.de](http://berlin.de/) are these as you can check on Zeplin or on BerlinDE CI
<img width="620" alt="Screenshot 2023-12-07 at 16 51 38" src="https://github.com/liqd/a4-meinberlin/assets/59610786/2bf9d23a-a53a-4d4f-95d3-6fee47084950">

Red: #E40622
Yellow: #FFE60C
Green: #00A982
*

PR #5410, opened 2024-01-22
ampel redesign fixes

hom3mad3:
## Ampel (Moderator Status): update color variables
Fixes https://github.com/liqd/a4-meinberlin/issues/5367

moderator-status-colors

Can be tested on http://localhost:8003/projekte/module/idea-collection/ 

### Steps
once moderation is added to idea, check if moderator status colors on list item card match the redesign

### DIFF
--- a/meinberlin/assets/scss/styles_user_facing/_variables.scss
+++ b/meinberlin/assets/scss/styles_user_facing/_variables.scss
@@ -35,9 +35,9 @@ $link-hover-color-btn: darken($link-color-btn, 15%) !default;
 $border-color: #ccc !default;
 $input-border-color: $border-color !default;
 
-$feedback-color-accepted: #0a820a !default;
-$feedback-color-rejected: #dc3818 !default;
-$feedback-color-consideration: #ffae00 !default;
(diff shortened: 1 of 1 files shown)


meinberlin/assets/scss/components_user_facing/_moderator_status.scss
```diff
@@ -12,4 +12,20 @@
...
+        background-color: $feedback-color-consideration;
+    }
+
+    &.label--rejected {
+        background-color: $feedback-color-rejected;
+    }
+
+    &.label--accepted {
+        background-color: $feedback-color-accepted;
+    }
```
philli-m, line 30: these styles are currently defined in label.scss https://github.com/liqd/a4-meinberlin/blob/b1033aba0f8442c10fab3fa31c0e0deed6d380d0/meinberlin/assets/scss/components/_label.scss#L39
I think they should either be updated only there or deleted

philli-m (changes requested):
looks good just need to clean up styles

philli-m (approved):
Nice!

PR #5416, opened 2024-01-25
fix email signature

xi:
see https://datatracker.ietf.org/doc/html/rfc3676#section-4.3

### DIFF
--- a/meinberlin/templates/email_base.txt
+++ b/meinberlin/templates/email_base.txt
@@ -10,7 +10,7 @@
 
 {% block reason %}{% endblock %}
 
----
+-- 
 {% translate 'meinBerlin is a participation platform operated by' %}
 Liquid Democracy e.V., Am Sudhaus 2, D-12053 Berlin
 


goapunk:
hi @xi,
thanks for the fix and the link to the docs! We are currently still in the process of clarifying some issues around accepting PRs from external contributors and I would therefore refrain from merging your PR until we resolved them :/ We discussed the change and generally agree that we would like to fix the non-standard signature dashes. If you don't mind I would open a new PR and close this one so it is included in the upcoming release

xi:
It feels a bit strange to be an "external contributor" when most of the commits in this repo were created by me :smile:. But true, it's been a few years. For a small change like this I see no harm in creating a new PR if that makes things easier for you.

goapunk:
> It feels a bit strange to be an "external contributor" when most of the commits in this repo were created by me 😄. But true, it's been a few years. For a small change like this I see no harm in creating a new PR if that makes things easier for you.

fair point ;-) thanks, will make a new PR tomorrow

PR #5433, opened 2024-02-01
apps/bplan: remove topics from updated_fields

goapunk:
**Describe your changes**
fixes https://sentry.liqd.net/organizations/liqd/issues/3580/

Testing: create a development bplan in dasbboard (with a valid identifier ? ->  `1 - 1`)

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/bplan/tasks.py
+++ b/meinberlin/apps/bplan/tasks.py
@@ -55,7 +55,17 @@ def get_bplan_api_pk_to_a4_admin_district_dict():
 
 
 @shared_task
-def get_location_information(bplan_id):
+def get_location_information(bplan_id: int):
+    """Fetch and update the location information for a bplan.
+    Called via the post_save signal in signals.py. To prevent the signal
+    from retriggering after updating the location information save() needs
+    to be called with the update_fields parameter: save(update_fields=["point", "admistrative_district"])
+
+    Parameters
+        ----------
+        bplan_id
+            The primary key of the bplan
+    """
     bplan = Bplan.objects.get(pk=bplan_id)
     point, district_pk = get_bplan_point_and_district_pk(bplan.identifier)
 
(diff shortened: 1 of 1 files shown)


meinberlin/apps/bplan/tasks.py
```diff
@@ -73,4 +73,4 @@ def get_location_information(bplan_id):
 
     topic = Topic.objects.get(code="URB")
     bplan.topics.set([topic])
-    bplan.save(update_fields=["point", "administrative_district", "topics"])
+    bplan.save(update_fields=["point", "administrative_district"])
```
m4ra, line 76: here just a `save()` is enough since the new fields are assigned in lines 64 and 65. And in line 75, should be `bplan.topics.add(topic)`

m4ra (changes requested):

meinberlin/apps/bplan/tasks.py
m4ra: just to clarify, that the `update_fields` get populated when we call `update()` rather than `save()`. Here for instance we could do `bplan.update(administrative_district = dis_dict[district_pk], point=point)` on line 64, instead of calling `save()` with the `updated_fields` arg. It skipped me last time I reviewed it.

goapunk: @m4ra thanks for the explanation, what do you suggest? stick with save or go for update()?

goapunk: @m4ra ok, this should have been documented but we do need the `update_fields`, otherwise it's an endless loop of celery tasks, see here: https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/apps/bplan/signals.py#L11

m4ra: oh yes, `update()` bypasses `save()`, and updates directly the DB values.

m4ra (approved):

Issue #5427, opened 2024-01-30
#7762 [mB] Fullwidth-Servicepanel with phase title etc. breaks at 1181px width


pascalkienast:
**URL:** https://meinberlin-design-dev.liqd.net/
**user:** any
**expected behaviour:** i want to have the servicepanel with the phase title and description like in the designs and also like the berlin-styleguide
**behaviour:** on devices with an screensize smaller than 1182px width (so all mobile devices), the fullwidth service panel breaks and is only half displayed
**important screensize:** 1181px width
**device & browser:** any
**Comment/Question:** 

Screenshot?
![Bildschirmfoto 2024-01-30 um 10 23 21](https://github.com/liqd/a4-meinberlin/assets/6682526/80f23c54-60f9-426d-870c-bcb24122c3ff)

![Bildschirmfoto 2024-01-30 um 10 23 09](https://github.com/liqd/a4-meinberlin/assets/6682526/f65ad121-d179-4c76-89a4-a15ff1e54899)

On mobile devices:
![Screenshot 2024-01-30 at 10-01-45 Samsung Galaxy S22 v12 0](https://github.com/liqd/a4-meinberlin/assets/6682526/7bb18a90-4b29-4614-860c-29b8c226a9c4)
![Screenshot 2024-01-30 at 10-03-33 iPhone 12 v14 7](https://github.com/liqd/a4-meinberlin/assets/6682526/624b5d39-e34c-4e9c-8198-8c2c0e64f13e)

goapunk:
fixed by #5442

PR #5442, opened 2024-02-05
fix for issue #5427: addressing bug introduced in prior fix

hom3mad3:
## Description:
This PR addresses issue #5427 on GitHub, where a bug was introduced in a prior fix that did not actually resolve the underlying issue. The fix has been removed, and the issue has been properly resolved.

## Changes:

Removed previously added fix that introduced a bug and did not solve the issue effectively.

## Testing:
Create a single module project:
    1. go to http://127.0.0.1:8003/dashboard/organisations/liqd/projects/
    2. click on "Add new project", add the project and then fill all the required fields
    3. add a module (e.g. participatory budgeting 2) and fill in all fields
    4. click add module to project
    5. publish project

The changes can be tested by navigating to `/projekte/module/name-of-module` and verifying that the issue is resolved.

**Tasks**

### DIFF
--- a/meinberlin/assets/scss/components_user_facing/_phase_info.scss
+++ b/meinberlin/assets/scss/components_user_facing/_phase_info.scss
@@ -1,16 +1,6 @@
 .phase-info {
     background-color: $blue-dark;
     color: $white;
-
-    // copied from BO as fullwidth styling not applying on single module project page
-    position: relative;
-    z-index: 0;
-    left: 50%;
-    right: 50%;
-    width: 100vw;
-    margin-left: -50vw;
-    margin-right: auto;
-    overflow: hidden;
 }
 
 .phase-info__title {


meinberlin/assets/scss/components_user_facing/_phase_info.scss
```diff
@@ -1,16 +1,6 @@
 .phase-info {
     background-color: $blue-dark;
     color: $white;
-
```
goapunk, line 4: this was introduced to fix the phase-info for single module projects, so removing it will break them again. I think we need to change this so it doesn't apply to the phase-info in multi-module projects. Not sure if that is the best way but having a separate `.tabpanel .phase-info {` seems to work for me.

goapunk (changes requested):
currently breaks the phase-info on single module projects (see comment)

meinberlin/assets/scss/components_user_facing/_phase_info.scss
goapunk: sure, locally there is no single module project by default, so you have to create one, rough steps are:

1.  go to http://127.0.0.1:8003/dashboard/organisations/liqd/projects/
2. click on "Add new project", add the project and then fill all the required fields
3. add a module (e.g. participatory budgeting 2) and fill in all fields
4. click add module to project
5. publish project

If you can't reproduce it we can also do a quick call :)

goapunk (approved):
LGTM

PR #5506, opened 2024-02-13
WIP projects: cache project queryset

m4ra:
@goapunk have a look, it's a first attempt to set the queryset in the cache when we call `reset_cache_for_projects`. Another option would be to set the queryset in the cache every 12hrs as a scheduled task. 
Also I added cache timeout as none, which means keep forever. Because we clear the cache everytime there are project updates, i thought might be ok that cache never expires. Otherwise i can set it as 24hrs.

Test are pending fixing. 

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/projects/api.py
+++ b/meinberlin/apps/projects/api.py
@@ -1,5 +1,4 @@
 from django.core.cache import cache
-from django.db.models import Q
 from django.utils import timezone
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import viewsets
@@ -9,6 +8,7 @@ from adhocracy4.projects.enums import Access
 from adhocracy4.projects.models import Project
 from meinberlin.apps.projects import serializers as project_serializers
 from meinberlin.apps.projects.filters import StatusFilter
+from meinberlin.apps.projects.tasks import get_project_queryset
 
 
 class ProjectListViewSet(viewsets.ReadOnlyModelViewSet):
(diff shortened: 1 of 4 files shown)


meinberlin/apps/projects/api.py
```diff
@@ -20,34 +20,17 @@ def __init__(self, *args, **kwargs):
...
+        return get_project_queryset()
 
     def list(self, request, *args, **kwargs):
         statustype = ""
         if "status" in self.request.GET:
             statustype = self.request.GET["status"]
         data = cache.get("projects_" + statustype)
         if data is None:
-            queryset = self.filter_queryset(self.get_queryset())
+            queryset = cache.get("project_queryset")
```
goapunk, line 31: the old code uses `self.filter_queryset()` but we don't seem to use it here anymore?

meinberlin/apps/projects/tasks.py
```diff
@@ -190,6 +219,12 @@ def reset_cache_for_projects(starts: bool, ends: bool) -> str:
...
             msg += "failed for active projects becoming past"
         else:
-            msg += "succeeded for active projects becoming past"
-    logger.info(msg)
-    return msg
+            msg += "succeeded for future projects becoming active"
+
+    # set the redis key: value for project's queryset
+    try:
+        cache.set("project_queryset", get_project_queryset())
```
goapunk, line 226: I think this might not work because the queryset at this point is not evaluated yet, I remember having that issue when I looked into it, that's why I had to cache it after serializing it.

m4ra:
To be fixed in a later stage, when we have more time to address the issue of the queryset.

PR #5518, opened 2024-02-15
fix template reset done

m4ra:
**Describe your changes**

**Tasks**
- [x] Changelog

### DIFF
--- a/locale/de_DE/LC_MESSAGES/djangojs.po
+++ b/locale/de_DE/LC_MESSAGES/djangojs.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-02-01 14:51+0100\n"
+"POT-Creation-Date: 2024-02-16 17:35+0100\n"
 "PO-Revision-Date: 2024-02-08 08:57+0000\n"
 "Last-Translator: Steffen Blindow <s.blindow@liqd.net>\n"
 "Language-Team: German <https://weblate.liqd.net/projects/meinberlin/js/de/>\n"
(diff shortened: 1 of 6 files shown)


meinberlin/templates/account/password_reset_done.html
```diff
@@ -0,0 +1,19 @@
...
+    {% translate "Password Reset" %}
+{% endblock head_title %}
+
+{% block content %}
+    <h1>{% translate "Password Reset" %}</h1>
+    {% if user.is_authenticated %}
+        {% include "account/snippets/already_logged_in.html" %}
+    {% endif %}
+    <p>
+        {% blocktrans %}We have sent you an email. If you have not received it please check your spam folder. Otherwise contact us if you do not receive it in a few minutes.{% endblocktrans %}
```
goapunk, line 17: we have a custom german translation for this but it isn't picked up anymore because there was a wording change I think. So either we change `email` to `e-mail` here or change it the other way round in`meinberlin/apps/contrib/django_standard_messages.py`.

goapunk (changes requested):
Really nice, just a small comment about the translation / wording

meinberlin/templates/account/password_reset_done.html
m4ra: I do see the German translation though locally
image

goapunk: yes, that's the one provided by allauth / django, if you check on stage: https://meinberlin-stage.liqd.net/accounts/password/reset/done/ the wording is different (more formal german)

m4ra: oh ok, I got what you mean now.
I will change the english

m4ra: If I add what we have in django messages in the template doesn't work, then it doesn't render into German at all. I saw that the django translation in locales, it is using the original msg of the template.

m4ra: @goapunk any idea about that above? Can't make it to work.

locale/de_DE/LC_MESSAGES/djangojs.po
```diff
@@ -190,6 +190,51 @@ msgstr "Bearbeiten"
 msgid "Report"
 msgstr "Melden"
 
+#: adhocracy4/comments_async/static/comments_async/ai_report.jsx:6
+#: adhocracy4/comments_async/static/comments_async/ai_report.jsx:6
+msgctxt "defakts"
+msgid ""
```
goapunk, line 196: hmmm should we delete the translations for defakts? They will just confuse translators for mb

goapunk (approved):
LGTM

Issue #5401, opened 2024-01-17
#7804 [mB] re-design login button


NaneKL:
**URL:** 
https://meinberlin-design-dev.liqd.net/projekte/new/

**user:** 
All

**expected behaviour:** 
The display of login is changed according to design

**behaviour:** 
There is a white gap between the login and the picture below, that isn´t there in the guidelines

**important screensize:**
Not relevant

**device & browser:** 
All

**Comment/Question:** 

Screenshot?
<img width="1245" alt="Bildschirmfoto 2024-01-17 um 09 34 58" src="https://github.com/liqd/a4-meinberlin/assets/147801301/4b2a34f9-bc99-4175-9b40-a6f3b4032bb5">

PR #5544, opened 2024-02-21
fix issue #5401

hom3mad3:
**Describe your changes**
fix https://github.com/liqd/a4-meinberlin/issues/5401 

added a negative margin on child element in order to fix inheritance issue.


goapunk (approved):
LGTM

PR #5550, opened 2024-02-27
[7863] redesign comment form

goapunk:
**Describe your changes**

**Tasks**
- [ ] Changelog


meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
```diff
@@ -253,7 +253,6 @@ <h3 class="detail-info__section-title">{% translate 'Remark' %}</h3>
         {% endif %}
     {% endblock %}
         <section>
-            <h2 class="visually-hidden">{% translate 'Comments' %}</h2>
```
goapunk, line 256: already comes from the react comments

meinberlin/assets/scss/components_user_facing/_control-bar.scss
```diff
@@ -23,12 +23,4 @@
     .button {
         margin-left: 0;
     }
-
-    @media screen and (max-width: var(--breakpoint-tablet)) {
```
goapunk, line 27: `var()` doesn't work in `@media` queries as far as I know? Adding the real breakpoint seems to distort the buttons so I think we should remove this for now

goapunk: I actually added the BO breakpoints (e.g. `breakpoint-tablet`) but then ended up removing them again as I didn't really need them and was wondering if there was a good reason why they haven't been added yet. But I agree we should add them then rather using actual numbers as it's much harder to read

goapunk:
@hom3mad3 as note/question: on mobile there is a bigger margin between the two action buttons (cancel and submit, e.g. when editing a comment) then in the designs. However the margins come from the styleguide so I wasn't sure what do here (follow styleguide or follow design).

meinberlin/assets/scss/components_user_facing/_a4-comments.scss
```diff
@@ -1,53 +1,65 @@
...
+
+    .a4-comments__char-count-word {
+      display: none;
     }
+  }
 }
 
+/* TODO: check if everything below here is still needed
+
 .a4-comments__child--list {
```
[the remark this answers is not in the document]
goapunk: I kinda thought to do that gradually with each comment story (we probably still need parts of this later) but I can also remove it now and add it back in when needed.

meinberlin/apps/contrib/templates/meinberlin_contrib/item_detail.html
goapunk: I think we do need it in a+, so it's a bit of a clash. I guess we could remove it in a4 and then a+ has to set it in the templates where the comments are used. Would that be better?

goapunk: I updated the a4 PR to remove it

goapunk: @hom3mad3 just realized you mentioned "Feedback", that's not part of the comments, that's the moderator feedback. So we probably still want the `Comments` landmark? However, the "Join the Discussion" then should be a h3?

goapunk: I updated it in a4, should be better now

PR #5565, opened 2024-03-13
settings/base: add ckeditor link decorator to allow opening links in …

goapunk:
…new tab

**Describe your changes**
We want this to achieve the same behavior as with the old ckeditor. When inserting a link you can now choose to open it in a new tab

**Tasks**
- [ ] Changelog


goapunk:
@hom3mad3 good point, so I guess we'd have to decide if we still want to keep it for Android Webviews or ignore them. What do you think?

PR #5577, opened 2024-03-19
[8037] WIP: update a4forms/form_field.html and use it in all dashboard forms

goapunk:
…ib/form_field.html before the redesign, use for all dashboard forms

**Describe your changes**
I've updated the a4forms/form_field.html to be equal to the old meinberlin_contrib/form_field.html and swapped it out everywhere in the dashboard, so most or even all dashboard forms should be styled properly again.

**Tasks**
- [ ] Changelog


goapunk:
> @goapunk just wondering if those changes would that make it possible to delete `meinberlin/templates/a4dashboard/includes/form_field.html` (also replaced in the files where its being instantiated)?

We'll have to check if it's used somewhere in a4, otherwise we have to keep the overwrite. I'll have a look

goapunk:
@hom3mad3 it's used a lot in a4, so I think we need to keep it

PR #5578, opened 2024-03-20
[7862] redesign comment filters and move some ControlBar components to a4 to be

goapunk:
used for the comment filtering and sorting

**Describe your changes**
redesign comment filters and sorting according to design and mb styleguide
most changes are done in the a4 pr, this just adding the styles and adjusting the ControlBar component to the components which moved to a4.

depends on https://github.com/liqd/adhocracy4/pull/1585

Can be tested on any module which has comments, e.g. idea collection. To test you need to `npm link` the above branch locally.

**Tasks**
- [x] Changelog


goapunk:
@hom3mad3 the overlay should be fixed now!

PR #5660, opened 2024-07-09
[8163] WIP api reset cache with a helper query

m4ra:
**Describe your changes**

**Tasks**
- [ ] Changelog


meinberlin/apps/projects/api.py
```diff
@@ -11,6 +11,28 @@
...
+        .select_related("administrative_district", "organisation")
+        .prefetch_related(
+            "moderators",
+            "plans",
+            "organisation__initiators",
+            "module_set__phase_set",
+        )
+    )
+    return projects
+
```
m4ra, line 35: I left this as the reference of the ORM filtering

meinberlin/apps/projects/api.py
```diff
@@ -20,25 +42,7 @@ def __init__(self, *args, **kwargs):
...
-            .order_by("created")
-            .select_related("administrative_district", "organisation")
-            .prefetch_related(
-                "moderators",
-                "plans",
-                "organisation__initiators",
-                "module_set__phase_set",
-            )
-        )
+        projects = PublicProjectView.objects.all()
```
m4ra, line 45: this is now querying the virtual db table

meinberlin/apps/projects/helpers.py
```diff
@@ -0,0 +1,40 @@
...
+            a4modules_module module ON project.id = module.project_id
+        LEFT JOIN 
+            a4phases_phase phases ON module.id = phases.module_id
+        WHERE 
+            (project.project_type = 'a4projects.Project' OR project.project_type = 'meinberlin_bplan.Bplan')
+            AND (project.access = 1 OR project.access = 2)
+            AND project.is_draft = FALSE
+            AND project.is_archived = FALSE
+        ORDER BY 
+            project.created;"""
```
m4ra, line 37: I think I need a GROUP BY to add the start/end dates here. Have to look more into that.

meinberlin/apps/projects/helpers.py
```diff
@@ -0,0 +1,40 @@
...
+        LEFT JOIN 
+            a4projects_project_moderators moderator ON project.id = moderator.project_id
+        LEFT JOIN 
+            meinberlin_plans_plan plan ON project.id = moderator.project_id
+        LEFT JOIN 
+            meinberlin_organisations_organisation_initiators initiators ON organisation.id = initiators.organisation_id
+        LEFT JOIN 
+            a4modules_module module ON project.id = module.project_id
+        LEFT JOIN 
+            a4phases_phase phases ON module.id = phases.module_id
```
m4ra, line 30: Not sure if those are the correct query to get the phases, any ideas welcome. need to research more.

meinberlin/apps/projects/tasks.py
```diff
@@ -191,5 +192,7 @@ def reset_cache_for_projects(starts: bool, ends: bool) -> str:
             msg += "failed for active projects becoming past"
         else:
             msg += "succeeded for active projects becoming past"
+    # update virtual table view
+    public_projects_exec_raw_query()
```
m4ra, line 196: when clearing the cash, the raw query it reconstructs the virtual table. But we could also make a separate task to do that few times in a day.

meinberlin/apps/projects/helpers.py
goapunk: hmm, as this is already a rather complex query I'm not super convinced this is the easier way to go (i.e. having to maintain a raw sql query if we change anything on the models vs the ORM just giving us an error) compared to moving the query and the filtering into functions we can call from the cache task. What are your thoughts, I haven't worked a lot with raw sql in django before, so this seems way more complicated to me :/

m4ra:
Closing this.

PR #5664, opened 2024-07-15
[8163] add public projects queryset

m4ra:
**Describe your changes**

**Tasks**
- [x] Changelog

### DIFF
--- a/meinberlin/apps/projects/api.py
+++ b/meinberlin/apps/projects/api.py
@@ -1,5 +1,6 @@
 from django.core.cache import cache
 from django.db.models import Q
+from django.db.models import QuerySet
 from django.utils import timezone
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import viewsets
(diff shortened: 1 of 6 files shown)


meinberlin/apps/projects/api.py
```diff
@@ -20,25 +43,10 @@ def __init__(self, *args, **kwargs):
...
-                "moderators",
-                "plans",
-                "organisation__initiators",
-                "module_set__phase_set",
-            )
-        )
+        projects = cache.get("public_projects")
+        if projects is None:
+            projects = get_public_projects()
+            cache.set("public_projects", list(projects))
```
m4ra, line 49: @goapunk returning it as list, it loads the query eagerly I found out.

goapunk (commented):
I think this is much easier to read! However, this introduces a new cache instead of refreshing the existing one. Wouldn't it make more sense to also a function which takes the queryset from `get_public_projects()` and then 
1. filters it same as in https://github.com/liqd/a4-meinberlin/blob/main/meinberlin/apps/projects/filters.py
2. serializes it with the 4 serializers (active, future, past, all)
3. overwrites the old caches (active, future, past, all)

meinberlin/apps/projects/tasks.py
```diff
@@ -134,62 +190,16 @@ def schedule_reset_cache_for_projects() -> bool:
...
-                "extprojects",
-            ]
-        ):
-            msg += "failed for active projects becoming past"
-        else:
-            msg += "succeeded for active projects becoming past"
-    logger.info(msg)
-    return msg
+
+    set_cache_for_projects()
```
m4ra, line 205: @goapunk this is called whenever a project change status of participation

meinberlin/config/settings/production.py
```diff
@@ -48,4 +48,9 @@
         "schedule": crontab(minute="*/10"),
         "args": (),
     },
+    "set-cache-for-projects-every-1-hr": {
+        "task": "set_cache_for_projects",
+        "schedule": crontab(minute=0, hour="*/1"),
+        "args": (),
+    },
 }
```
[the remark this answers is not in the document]
goapunk: the cache timeout is 24hrs, so personally I would say every 6h or even 12h is enough (if a project stars/ends/changes it's reset anyway)

tests/projects/test_tasks_caching.py
```diff
@@ -114,3 +115,56 @@ def test_task_schedule_reset_cache_for_projects_becoming_past(
...
+    # check function set_cache_for_projects
+    set_cache_for_projects()
+
+    with freeze_time(now):
+        active_projects = cache.get("projects_activeParticipation")
+        future_project = cache.get("projects_futureParticipation")
+        past_project = cache.get("projects_pastParticipation")
+        projects = cache.get("projects_")
+        assert active_projects is not None
+        print(active_projects)
```
goapunk, line 166: will remove this and the line above (or was there a reason to assert `active_projects` but not the others?

goapunk (approved):
Really nice, I added a small change as mentioned in a comment.

Issue #5587, opened 2024-04-08
Unify dropdown functionality across all dropdowns


hom3mad3:
## **Issue:**
`dropdown.js` lacks certain features found in `user_indicator.js`, such as keyboard navigation and close-on-click-outside functionality. However, `user_indicator.js` isn't easily modularized due to its design, which focuses on a single element without considering component reusability.

## **Proposal:**
To address this, I'd suggest refactoring `user_indicator.js` onto `dropdown.js` to integrate the missing features, enhancing the overall reusability of dropdown functionality. This will ensure that all dropdowns use one JS file, even when multiple instances are present on the page.

## 1. **dropdown.js**
element triggered with `dropdown.js` and styled by `scss/components_user_facing/_dropdown.scss`:

<img src="https://github.com/liqd/a4-meinberlin/assets/8156337/c351d568-c9fe-4611-a1e8-283f341dc80d" alt="item-detail-dropdown" width="300px">

## 2. **user_indicator.js**
event is triggered w/ `user_indicator.js` and styled by `/scss/components_user_facing/_user_indicator.scss`:

<img src="https://github.com/liqd/a4-meinberlin/assets/8156337/bf7ce748-717a-4e03-97db-f2aa167243ef" alt="user-dropdown" width="300px">

Issue #5596, opened 2024-04-16
#8039 [mB] re-design component item detail: Dots menu only collapses when clicking the button that originally opened it


sblindow:
**URL:** 
https://meinberlin-design-dev.liqd.net/ideas/2024-01098/

**user:** 
all users

**expected behaviour:**
The Dots menu should close when clicking anywhere outside the menu after it has been opened.

**behaviour:**
The menu only closes by clicking the button that originally opened it.

**important screensize:**
Not relevant.

**device & browser:**
MacOS; Firefox: 124.0.2 (64-Bit)

Issue #5593, opened 2024-04-11
new item detail dropdown is missing a z-index


goapunk:
**URL:** https://meinberlin-design-dev.liqd.net/mapideas/2023-01046/
**user:** admin
**expected behaviour:** dropdown is completely visible
**behaviour:** dropdown is partly covered by map
**important screensize:**
**device & browser:** 
**Comment/Question:** 

Screenshot?

CarolingerSeilchenspringer:
@hom3mad3 this issue can be closed, right?

PR #5675, opened 2024-07-22
fix: refactor dropdown to make it reusable

hom3mad3:
fixes https://github.com/liqd/a4-meinberlin/issues/5587, https://github.com/liqd/a4-meinberlin/issues/5596, https://github.com/liqd/a4-meinberlin/issues/5593


meinberlin/apps/users/templates/meinberlin_users/user_indicator.html
```diff
@@ -1,61 +1,55 @@
...
+                        </form>
+                    </li>
+                </ul>
+            </div>
+        {% else %}
+            <button data-dropdown-trigger type="button" aria-haspopup="menu" aria-controls="dropdown-menu" tabindex="0">
+                <i class="fas fa-user text--color-primary" aria-hidden="true"></i>
+                {% translate "Login" %}
+            </button>
+            <div role="menu" class="dropdown__menu" aria-label="Login of Register" tabindex="-1" data-dropdown-menu>
```
goapunk, line 40: Nothing new but should this be "Login **or** Register" ? And it should also be translated, right?

meinberlin/apps/contrib/assets/dropdown.js
```diff
@@ -1,41 +1,67 @@
...
+    if (this.container.classList.contains('is-open')) {
+      this.closeDropdown()
+    } else {
+      this.openDropdown()
     }
-  },
+  }
+
+  openDropdown () {
+    this.dropdown.style.display = 'block'
```
goapunk, line 27: I'm fine with this but just wondering if having it in the css is a bit of a clearer division between functionality and styling as we have the `is-open` class to apply it ?

goapunk (changes requested):
Works nicely and looks good, just some small remarks/questions

goapunk (approved):
LGTM

PR #5677, opened 2024-07-23
[6881] replace datepicker with flatpicker

goapunk:
**Describe your changes**
Replace the old datepicker with flatpickr and connect the flatpickr instances (if there are multiple ones, e.g. for module phases). Also adds a time-only datepicker to the time field.

**Tasks**
- [ ] Changelog


goapunk:
@hom3mad3 this is more of a wip/concept to discuss, maybe you can have a look already before this is going in the wrong direction

PR #5679, opened 2024-07-24
[#8232] update user facing variables

hom3mad3:
streamline and organize the user-facing variables

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -52,6 +52,11 @@
       "ignoreKeywords": ["inherit", "transparent"]
     }],
 
-    "declaration-empty-line-before": null
+    "declaration-empty-line-before": null,
+
+    "scss/dollar-variable-empty-line-before": ["always", {
+      "except": ["first-nested"],
+      "ignore": ["after-comment", "after-dollar-variable"]
+    }]
   }
 }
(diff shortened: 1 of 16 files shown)


.stylelintrc.json
```diff
@@ -52,6 +52,10 @@
       "ignoreKeywords": ["inherit", "transparent"]
     }],
 
-    "declaration-empty-line-before": null
+    "declaration-empty-line-before": null,
+
+    "scss/dollar-variable-empty-line-before": ["always", {
+      "except": ["first-nested", "after-comment", "after-dollar-variable"]
```
goapunk, line 58: I think this needs to be split up into `"except": ["first-nested"]` because we never want an empty-line on the first nested element)and `"ignore": ["after-comment", "after-dollar-variable"]` so both empty-line and no empty-line are allowed after a comment or a variable

goapunk (commented):
looks good to me, I'm a bit uncertain of whether removing all placeholders variables is a good or bad idea to me. On one hand it makes reading easier as you know the value, on the other changing e.g. the text color means changing it in a lot of places. I'm fine going with this as this is your area of expertise though.

goapunk (approved):
LGTM

PR #5686, opened 2024-08-15
[#7864] a4-comments: redesign comments and replies

hom3mad3:
**Describe your changes**
redesign comments and replies, mostly overwriting styles a4-comments
- depends on https://github.com/liqd/adhocracy4/pull/1669
- can be tested [here](https://meinberlin-design-testing-dev.liqd.net/ideas/2024-01098/)
- after a couple of false starts i decided to reinstate bootstrap dependencies for the comments dropdown and modal components


meinberlin/assets/scss/components_user_facing/_a4-comments.scss
```diff
@@ -1,94 +1,233 @@
...
-  }
+    .form-group {
+        margin-bottom: 0;
+    }
 
-  .a4-comments__char-count {
-    @extend .formgroup__help;
-    color: var(--color-black);
-    margin-top: 0.25 * $spacer;
+    &__char-count {
```
goapunk, line 108: this doesn't seem to work, the char count doesn't get any of the styling

goapunk (changes requested):
looks good, one small thing is the char-count styling is not working any more and the action buttons start disappearing on small screen sizes (<380)

meinberlin/assets/scss/components_user_facing/_a4-comments.scss
goapunk: ah ok, in firefox I don't get a horizontal scrollbar so they just disappear to the right (share button)

goapunk: @hom3mad3 sorry, it actually works, I was just too stupid to trigger it!

PR #5689, opened 2024-08-27
a11y issue fix: pill text color not accessible

hom3mad3:
apply base text-color to all pill elements

 | Before      | After |
| ----------- | ----------- |
| pill-before      | pill-after      |

issue:
pill-color


m4ra (approved):
LGTM

PR #5695, opened 2024-09-03
meinberlin_cms/blocks: wagtail accordion redesign

hom3mad3:
**Redesign wagtail accordion element**

In this PR, I'm setting the `<details>` HTML element as our go-to for accordions in the codebase, which i already used in `moderator_notes.html`. The BO styleguide’s accordion is overloaded with JavaScript and does a lot of complicated things just to create a simple component and improve accessibility (which the `<details>` element already does by default). Their accordion animation is a really nice touch, but it needs JS to work.

The new details element version is not animated, as I made the choice to forego the use of JavaScript here, though I'm open to find a compromise and try some progressive enhancement for the animation.

I've adopted a functional naming for the new Wagtail block to maintain flexibility and clarity (and consistency with the template names)

The ck-editor element is importing the BO styles, and I’m not sure if we should unify them. The issue is that ck-editor isn’t generating accessible children anyways, so I wonder if it’s more trouble than it’s worth to adapt the `<details>` element in there. We’d also need to figure out how to turn it into a list of accordions for better semantics.

I would tackle the ck-editor in a different PR

Screenshot 2024-09-03 at 17-26-30 meinBerlin — meinBerlin

**Tasks**
- [x] Changelog


meinberlin/apps/cms/models/pages.py
```diff
@@ -82,7 +82,7 @@ class HomePage(Page):
 class DocsPage(Page):
     body = fields.StreamField(
         [
-            ("documents_list", cms_blocks.DocsBlock()),
+            ("documents_list", cms_blocks.AccordionBlock()),
```
goapunk, line 85: should we change this to `accordion_list` so the element in wagtail is also called that?

goapunk (changes requested):
Looks good to me, just some remarks about translations and a general question on the process: 
With this pr all existing watgtail accordions will be wiped, we should check with PMs if that's ok (if not done already), otherwise I'd suggest we keep the old document block and just add the accordion block(s) as new ones, that way they can copy the old content into the new blocks more easily and once that's done we remove the old block entirely.

goapunk (approved):
Nice!

PR #5702, opened 2024-09-27
emails: improve text-based emails (remove html)

### LINKED ISSUE #5417: HTML in plain text emails
I just received an email to confirm my mail address. It contained HTML in the plain text part:

```
<p>
bitte bestätigen Sie Ihre E-Mail-Adresse um Ihr Konto auf mein.berlin.de freizuschalten.
</p>
<p>
Nach der Bestätigung können Sie Ideen und Kommentare anlegen und bewerten und an Umfragen teilnehmen.
</p>
<p>
Bei Fragen können Sie sich an support@mein.berlin.de wenden.
</p>
```

I tried to dig into the code to see why this happens. It seems like most email templates overwrite `{% block content %}` which is used in both `email_base.html` and `email_base.txt`. The proper solution is probably to only use plain text inside `{% block content %}`, and additionally also overwrite `{% block content_html %}` is HTML is desired.

vellip:
This fixes #5417, there might still be some occasions with HTML in other blocks (like `reason`), but this removes all the html from emails in case your email client doesn't support html.


xi:
Hi @vellip :wave: !

vellip:
hey @xi 😄 👋

PR #5712, opened 2024-10-01
templates/account: refactor narrow wrapper

hom3mad3:
### Changes
- Renamed `_account.scss` to `_narrow-wrapper.scss` to reflect its use on only a handful of account pages that implement a custom narrow container. This container has a special treatment of the button, which is full width on all viewports. 
- Updated relevant templates (`signup.html`, `login.html`, and `password_change.html`) to use the renamed styles.  
- Regarding the `secondary-cta` class, I applied it as a nested class since it's a one-off style, although it does not strictly follow BEM conventions. Not quite sure about that, maybe it belongs on utility?

- should i rather create a separate cta-wrapper and then take the variation into account there? 

Let me know if this works or if you'd like any adjustments!

Screenshot 2024-10-01 at 13-50-37 Passwort zurücksetzen — meinBerlin

### DIFF
--- a/meinberlin/templates/account/login.html
+++ b/meinberlin/templates/account/login.html
@@ -6,7 +6,7 @@
 {% endblock head_title %}
 
 {% block content %}
-    <div class="account__wrapper">
+    <div class="narrow-wrapper">
         <h1>{% translate "Login" %}</h1>
 
         <form method="post" action="{% url 'account_login' %}" class="form--base">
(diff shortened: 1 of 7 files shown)


meinberlin/templates/account/login.html
```diff
@@ -27,17 +27,19 @@ <h1>{% translate "Login" %}</h1>
...
-                <a href="{% url 'account_reset_password' %}">{% translate "Forgot Password?" %}</a>
-            </div>
+            <div class="narrow-wrapper__cta">
+                <div class="secondary-cta">
+                    <a href="{% url 'account_reset_password' %}">{% translate "Forgot Password?" %}</a>
+                </div>
 
-            <button class="button account__button" type="submit">{% translate "Login" %}</button>
+                <button class="button button--full-width" type="submit">{% translate "Login" %}</button>
+            </div>
```
vellip, line 36: maybe we can drop the `narrow-wrapper__cta` and instead of `secondary-cta` call it `narrow-wrapper__secondary-cta`? That would also make it more consistent with the other html templates in this PR as the buttons there are not wrapped in `narrow-wrapper__cta`.

meinberlin/templates/account/password_reset.html
```diff
@@ -1,29 +1,29 @@
...
 {% block content %}
-    <h1>{% translate "Password Reset" %}</h1>
-    {% if user.is_authenticated %}
-        {% include "account/snippets/already_logged_in.html" %}
-    {% endif %}
+    <div class="narrow-wrapper">
+        <h1>{% translate "Password Reset" %}</h1>
 
-    <p>{% translate "Forgotten your password? Enter your e-mail address below, and we'll send you an e-mail allowing you to reset it." %}</p>
+        <p>{% translate "Forgot password? Enter your email address and you will receive an email with which you can reset your password. Please contact us if you have any problems resetting your password." %}</p>
```
vellip, line 12: When are you updating messages, in general? As in translations? Should this happen in this PR?

PR #5718, opened 2024-10-10
update buttons wording and modify submit templates

hom3mad3:
update a few strings and expand previously created submit template snippets


vellip (approved):
LGTM, but do you think there is a simple and quick way to document these partials/includes of yours so others know how/ where to use them? Or do you think it's already simple enough? 
Feel free to merge otherwise.

goapunk:
@hom3mad3 not that I know of. We could create e.g. a subfolder where we can add docs for components if that suits the purpose? If a visual thing is better we could extend the component library with explanations I guess: https://meinberlin-dev.liqd.net/components/

PR #5721, opened 2024-10-15
[#8447] project overview: convert vanilla leaflet map to react leaflet & update styling

vellip:
uses react leaflet for project overview and adds some rough styling/ layout


vellip:
> @vellip i'm not sure, but should we also remove the packages from package.json?

I'd rather not do it yet. Some of leaflet is still referenced in the webpack config.

PR #5724, opened 2024-10-16
Fix styling for actions of comments on mobile

### LINKED ISSUE #5694: #7864 [mB] redesign comments and replies (Report/Share)
**URL:** https://meinberlin-design-testing-dev.liqd.net/ideas/2024-01098/
**important screensize:** *Mobile*
**device & browser:** *Safari Version 17.5 (19618.2.12.11.6) / Chrome Version 126.0.6478.127 (Official Build) (arm64)*
**Comment/Question:** 

Screenshot 2024-09-03 at 09 00 26

Regarding this area here, I agree that it looks an error. I was thinking, if its not super difficult/time consuming to change that now, would it be possible to just leave the Like/Dislike/Comment on Mobile and add the Report and Share under the ellipsis-V? 

Also for Mobile, we could maybe leave a little bit less spacing between them cause on iPhone it looks like this and it would make more sense to have them more aligned with the grey line:
IMG_2453

vellip:
closes #5694 
the size of the last two icons bother me a bit still, because they seem so large in comparison to the others, but i left it as is for now.

CleanShot 2024-10-16 at 18 38 06@2x

desktop:
CleanShot 2024-10-16 at 18 44 29@2x

cc @CarolingerSeilchenspringer

### DIFF
--- a/meinberlin/assets/scss/components_user_facing/_rating.scss
+++ b/meinberlin/assets/scss/components_user_facing/_rating.scss
@@ -1,6 +1,10 @@
 .rating {
-    display: inline-block;
-    white-space: nowrap;
+    display: flex;
+
+    @media screen and (min-width: $breakpoint-palm) {
+        display: inline-block;
+        white-space: nowrap;
+    }
 }
 
 .rating__label {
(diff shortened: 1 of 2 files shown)


MarleyMi:
@vellip @CarolingerSeilchenspringer  The Icons also dont look aligned (maybe because of the Size) Should I ask our Design student Anastasiia to have look ?

PR #5731, opened 2024-11-05
[#8464] live-events hero

vellip:
introduce styling for live-questions and improve functionalities like full-screen and showing videos even if their phase is over

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -44,7 +44,6 @@ module.exports = {
     },
     dsgvo_video_embed: {
       import: [
-        'dsgvo-video-embed/css/dsgvo-video-embed.css',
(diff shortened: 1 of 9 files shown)


webpack.common.js
```diff
@@ -44,7 +44,6 @@ module.exports = {
     },
     dsgvo_video_embed: {
       import: [
-        'dsgvo-video-embed/css/dsgvo-video-embed.css',
```
goapunk, line 47: @vellip @hom3mad3 the css file produced by webpack (`dsgvo_video_embed.css`) is still used in various places. Removing this here will cause a 500 (https://meinberlin-design-dev.liqd.net/projekte/participatory-budgeting-3-support-test/). We need to either add it back in or remove it from the templates and use the new video-wall there as well

PR #5740, opened 2024-11-12
[8473] add kiezradar models

m4ra:
**Describe your changes**

- A Kiezradar query with a `text` field which can accommodate free text query.  
- A search profile related to user,  corresponds to the project overview filters, and which has a m2m relation with the existing project overview filters. The idea is that these filter can be multiselected and attached to different Search Profiles.
- A Serializer with a custom create() and update() to creating/updating m2m relations of the SearchProfile.

**Tasks**
- [x] Changelog


meinberlin/apps/kiezradar/views.py
```diff
@@ -0,0 +1,24 @@
...
+class SearchProfileViewSet(viewsets.ModelViewSet):
+    queryset = SearchProfile.objects.all()
+    serializer_class = SearchProfileSerializer
+
+    def perform_create(self, serializer):
+        # Create the SearchProfile instance
+        search_profile = serializer.save()
+
+        # Define topics or retrieve from request data
+        topics = self.request.data.get("topics", ["Default Topic 1", "Default Topic 2"])
```
m4ra, line 17: I added the default ones for testing purposes.

meinberlin/apps/kiezradar/signal.py
```diff
@@ -0,0 +1,4 @@
+from django.dispatch import Signal
+
+# Define custom signal with topics as an argument
+add_kiezradars_to_profile = Signal(providing_args=["instance", "topics"])
```
goapunk, line 4: the `providing_args` argument was removed in django 4.0 :)

meinberlin/apps/kiezradar/models.py
```diff
@@ -0,0 +1,64 @@
+from django.conf import settings
+from django.db import models
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from adhocracy4.maps.fields import PointField
+
+
+class KiezRadarFilter(models.Model):
+    topic = models.CharField(max_length=126, unique=True)
```
goapunk, line 10: I'm not entirely sure if I follow the structure here, do we need the `KiezRadarFilter` because it will be extended at some point in the future? Because if it only has the `topic` we could also just have a m2m to the `Topic` model on the `SearchProfile`, right? And in the case here: would it make more sense for the topic to be a foreign key on `Topic` instead of a `CharField` ? This way only topics which are defined can be added

meinberlin/apps/kiezradar/views.py
```diff
@@ -0,0 +1,24 @@
...
+
+        # Send the custom signal with topics
+        add_kiezradars_to_profile.send(
```
goapunk, line 20: Why do we need a signal here? can't we just add the topics to the `search_profile` directly?

goapunk (changes requested):
Added some questions and a note regarding `providing_args`

meinberlin/apps/kiezradar/signal.py
m4ra: oh shoot

meinberlin/apps/kiezradar/views.py
m4ra: Because of the m2m relation, search profile needs to be saved first in order to add topics, that's why the signal is `post_save`.

meinberlin/apps/kiezradar/models.py
m4ra: I understood the multiselect filter later will have options that are added either by admins/initiators, or on the fly like keywords, while creating the search profile, not from the topics which come from the django settings. Also the topic field here could be a json field which combines different topics. Need to ask PM for clarification.

goapunk: my understanding of the story was that currently we only add the existing filters:
- project title (free text)
- organisation (free text)
- type (informational, participation, ...)
- status (all, running, ended) 
- district
- topic (as in the settings)

and you can select multiple options per filter (except for status filter). Maybe we can talk about it in daily again with the responsible PM as I don't remember anything about admins/initiators being able to add options so might be good to clarifiy

m4ra: Clarified now, topics are the one predefined. Bur multiple filters for all categories, will result in search profiles that are almost similar for different free text searches, and too many relations.

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -0,0 +1,104 @@
...
+        # Pop many-to-many fields from validated_data
+        query = validated_data.pop("query", None)
+        organisations_data = validated_data.pop("organisations", None)
+        districts_data = validated_data.pop("districts", None)
+        project_types_data = validated_data.pop("project_types", None)
+        topics_data = validated_data.pop("topics", None)
+
+        search_profile = SearchProfile.objects.create(**validated_data)
+        if query:
+            query, _ = KiezradarQuery.objects.get_or_create(pk=query.id)
```
goapunk, line 63: I'm not sure I follow here: why use `get_or_create` here ? It only passes the pk, so the `text` field of the query will be empty, right?

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -0,0 +1,104 @@
...
+        if topics_data:
+            search_profile.topics.set(topics_data)
+
+        return search_profile
+
+    def update(self, instance, validated_data):
+        """
+        Custom update to handle many-to-many relationships.
+        """
+        # Pop many-to-many fields from validated_data
```
goapunk, line 83: what happens if I change the `query` on an existing search profile? Is this currently missing here? Do we need a `get_or_create` here as well?

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -0,0 +1,104 @@
...
+        organisations_data = validated_data.pop("organisations", None)
+        districts_data = validated_data.pop("districts", None)
+        project_types_data = validated_data.pop("project_types", None)
+        topics_data = validated_data.pop("topics", None)
+
+        # Update non-m2m fields
+        for attr, value in validated_data.items():
```
goapunk, line 90: could we just call `super` here or do we need to do it manually?

goapunk (commented):
Thanks, looks really good! Just a few questions about the `query` / serializer

meinberlin/apps/kiezradar/serializers.py
m4ra: forgot to add the text.

meinberlin/apps/kiezradar/serializers.py
m4ra: no, because the query relation to the search profile is not m2m, but one to many. so a query can linked to many profiles. I ditched the m2m here bcz having many queries for a profile would result in too many results, and also would be slow to load. We can change later if there is a strong request/opinion by the PM

m4ra: In the end, I had to add it in the update() as well, in case the user adds a new query text.

m4ra:
@goapunk ready for another round :)

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -0,0 +1,111 @@
...
+        # need to create a profile before we can add m2m relations
+        search_profile = SearchProfile.objects.create(**validated_data)
+
+        if organisations_data:
+            search_profile.organisations.set(organisations_data)
+        if districts_data:
+            search_profile.districts.set(districts_data)
+        if project_types_data:
+            for project_type in project_types_data:
+                project_type, _ = ProjectType.objects.get_or_create(pk=project_type.id)
```
goapunk, line 78: the project types are predefined and created in the data migration, right ? Do we need `get_or_create` here? When I look at `project_types_data` it already contains `ProjectType` objects, so we don't need to fetch them from the db again, doing `set()` should work I think.

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -0,0 +1,111 @@
...
+        districts_data = validated_data.pop("districts", None)
+        project_types_data = validated_data.pop("project_types", None)
+        topics_data = validated_data.pop("topics", None)
+
+        if query_text:
+            query, _ = KiezradarQuery.objects.get_or_create(text=query_text)
+            validated_data["query"] = query
+
+        # need to create a profile before we can add m2m relations
+        search_profile = SearchProfile.objects.create(**validated_data)
```
goapunk, line 70: Instead of popping and then setting the m2m relations manually below, can we just call `return super().create(validated_data)` here as it will set m2m relations for us ?

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -0,0 +1,111 @@
...
+        project_types_data = validated_data.pop("project_types", None)
+        topics_data = validated_data.pop("topics", None)
+
+        # Update one-to-many relationship
+        if query_text:
+            # remove existing query and assign or create another
+            query, _ = KiezradarQuery.objects.get_or_create(text=query_text)
+            validated_data["query"] = query
+
+        if organisations_data:
```
goapunk, line 102: same as for `create` above, the super call at the end already sets the m2m relations, so we don't need to pop and manually set them

meinberlin/apps/kiezradar/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
goapunk, line 1: this file should be removed :)

tests/kiezradar/test_api_kiezradar.py
```diff
@@ -0,0 +1,142 @@
...
+        "districts": [district.id],
+        "query": query.id,
+        "organisations": [organisation.id],
+        "topics": [topic.id],
+        "project_types": [project_type.id],
+    }
+
+
+@pytest.mark.django_db
+def test_create_search_profile(client, user, setup_data, kiezradar_query_factory):
```
goapunk, line 32: `kiezradar_query_factory` is unused here, right?

goapunk (changes requested):
Looks good, I left a few comments on (hopefully) simplifying the serializer create and update, and the empty test file needs to go to make the linter happy

meinberlin/apps/kiezradar/serializers.py
m4ra: the search profile has to be created first in order to add m2m relations

meinberlin/apps/kiezradar/serializers.py
m4ra: true, i wrote this before the custom migration

meinberlin/apps/kiezradar/serializers.py
goapunk: yes, but the super call does that as far as I can see: https://github.com/encode/django-rest-framework/blob/dbac145638758413b966c3418fa5f3f651e3e02a/rest_framework/serializers.py#L989

m4ra: Will try again.

goapunk: if it doesn't work we can also proceed like it is!

m4ra: I did, last time I tried I forgot to remove the line where I also save it manually

meinberlin/apps/kiezradar/tests.py
m4ra: yes, was generated when i created the app

meinberlin/config/settings/test.py
```diff
@@ -15,3 +15,12 @@
     from .polygons import *
 except ImportError:
     pass
+DATABASES = {
```
goapunk, line 18: we don't need this, right  ? :)

goapunk (approved):
Thanks, looks good! Just small not about the database setting in test.py

meinberlin/config/settings/test.py
m4ra: ah no, got committed by accident

PR #5751, opened 2024-11-19
[#8467] Allow moderators to like

sevfurneaux:
**Describe your changes**
This PR refactors the `QuestionModerator` component to a functional component, and also allows moderators to like alongside other users.

**Tasks**
- [x] Changelog

### DIFF
--- a/meinberlin/react/livequestions/QuestionModerator.jsx
+++ b/meinberlin/react/livequestions/QuestionModerator.jsx
@@ -1,190 +1,208 @@
-import React from 'react'
+import React, { useState, useEffect, useCallback } from 'react'
 import django from 'django'
 import LikeCard from './LikeCard'
 import { classNames } from '../contrib/helpers'
 
-export default class QuestionModerator extends React.Component {
-  constructor (props) {
-    super(props)
+const hiddenText = django.gettext('mark as hidden')
+const undoHiddenText = django.gettext('undo mark as hidden')
+const doneText = django.gettext('mark as done')
+const addLiveText = django.gettext('added to live list')
+const removeLiveText = django.gettext('remove from live list')
+const addBookmarkText = django.gettext('add bookmark')
+const removeBookmarkText = django.gettext('remove bookmark')
 
-    this.state = {
-      is_on_shortlist: this.props.is_on_shortlist,
-      is_live: this.props.is_live,
-      likes: this.props.likes.count,
-      session_like: this.props.likes.session_like,
-      is_hidden: this.props.is_hidden,
-      is_answered: this.props.is_answered
-    }
-  }
+const QuestionModerator = ({
+  category,
+  children,
+  displayIsAnswered,
+  displayIsHidden,
+  displayIsLive,
+  displayIsOnShortlist,
+  handleLike,
+  id,
+  is_answered: _isAnswered,
+  is_hidden: _isHidden,
+  is_live: _isLive,
+  is_on_shortlist: _isOnShortlist,
+  likes,
+  removeFromList,
+  togglePollingPaused,
+  updateQuestion
+}) => {
+  const [isOnShortlist, setisOnShortlist] = useState(_isOnShortlist)
+  const [isLive, setIsLive] = useState(_isLive)
+  const [isHidden, setIsHidden] = useState(_isHidden)
+  const [isAnswered, setIsAnswered] = useState(_isAnswered)
+  const [likeCount, setLikeCount] = useState(likes.count)
+  const [sessionLike, setSessionLike] = useState(likes.session_like)
 
-  toggleIsOnShortList () {
-    const value = !this.state.is_on_shortlist
-    const boolValue = (value) ? 1 : 0
+  // Copying props to local state ensures instant updates when interacting with the moderator buttons (for example: toggleIsOnShortList)
+  // If the props were used directly, there would be a UI delay due to polling in the QuestionBox component.
+  useEffect(() => {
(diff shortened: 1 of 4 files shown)


meinberlin/react/livequestions/QuestionModerator.jsx
```diff
@@ -1,190 +1,189 @@
...
+    setIsLive(props.is_live)
+  }, [props.is_live])
+
+  useEffect(() => {
+    setIsHidden(props.is_hidden)
+  }, [props.is_hidden])
+
+  useEffect(() => {
+    setIsAnswered(props.is_answered)
+  }, [props.is_answered])
```
vellip, line 36: Why are you doing it this way instead of using a callback provided by the parent component? If you need to update your props, you should use a callback and not copy props into state.

meinberlin/react/livequestions/QuestionModerator.jsx
```diff
@@ -1,190 +1,189 @@
...
-    this.state = {
-      is_on_shortlist: this.props.is_on_shortlist,
-      is_live: this.props.is_live,
-      likes: this.props.likes.count,
-      session_like: this.props.likes.session_like,
-      is_hidden: this.props.is_hidden,
-      is_answered: this.props.is_answered
-    }
-  }
+const QuestionModerator = (props) => {
```
vellip, line 14: Maybe destructure the props argument

meinberlin/react/livequestions/QuestionModerator.jsx
vellip: Ah, I see, you just converted the old code. Could you investigate it still? Only if it doesn't take much time. The live-questions aren't really used that much.

meinberlin/react/livequestions/QuestionModerator.jsx
```diff
@@ -1,190 +1,189 @@
...
+            title={doneText}
+          >
+            <i className="fas fa-check" aria-hidden="true" />
+          </button>
+        )}
+        {props.displayIsLive && (
+          <button
+            type="button"
+            className={classNames(
+              'cardbutton card__button card__button--live',
```
vellip, line 174: can you check if `cardbutton` is still needed? I didnt find any styling for it. It also doesn't look very BEM to me.

vellip (commented):
Thanks for converting & updating this part too. Good job! I just got a few things.

meinberlin/react/livequestions/QuestionModerator.jsx
sevfurneaux: > Why are you doing it this way instead of using a callback provided by the parent component?

@vellip I asked myself the same question of the original implementation, so I used all the `props` directly, without copying state, however doing this means on update of the parent component, everything is delayed due to polling, even with the changed `500` ms.

If we can think of an alternative approach, happy to change it!

vellip: Eh, then let's leave it in. Maybe you can add a small comment with your findings or link to this discussion.

meinberlin/react/livequestions/QuestionModerator.jsx
sevfurneaux: `cardbutton` is from the compiled berlin_css.css which is part of the Card HTML in the styleguide: https://styleguide.berlin.de/patterns/09-vertical_marketing-page-modul-card/09-vertical_marketing-page-modul-card.html

meinberlin/react/livequestions/QuestionModerator.jsx
sevfurneaux: @vellip Destructuring to this:

```
const QuestionModerator = ({
  category,
  children,
  displayIsAnswered,
  displayIsHidden,
  displayIsLive,
  displayIsOnShortlist,
  handleLike,
  id,
  is_answered,
  is_hidden,
  is_live,
  is_on_shortlist,
  likes,
  removeFromList,
  togglePollingPaused,
  updateQuestion
}) => {
```

Causes eslint errors:

`error  Identifier 'is_answered' is not in camel case `

But changing to `isAnswered` then conflicts with the `isAnswered` local state naming:

`const [isAnswered, setIsAnswered] = useState(is_answered)`

Let's keep as `props` for now?

vellip:
```jsx
const QuestionModerator = ({
  category,
  children,
  displayIsAnswered,
  displayIsHidden,
  displayIsLive,
  displayIsOnShortlist,
  handleLike,
  id,
  **is_answered: _isAnswered,**
  is_hidden,
  is_live,
  is_on_shortlist,
  likes,
  removeFromList,
  togglePollingPaused,
  updateQuestion
}) => {
```

does that work?

vellip: without the ** of course

sevfurneaux: Oh nice, yeah that works 👍

https://github.com/liqd/a4-meinberlin/pull/5751/files#diff-11e1fc49d7ab8e826f53c0b167259c0e8f36bff2da36beac044a81d4531cd694R23-R26

meinberlin/react/livequestions/QuestionModerator.jsx
sevfurneaux: Added a comment here: https://github.com/liqd/a4-meinberlin/pull/5751/files#diff-11e1fc49d7ab8e826f53c0b167259c0e8f36bff2da36beac044a81d4531cd694R39-R40

PR #5779, opened 2024-12-03
[8450] update django-allauth

goapunk:
**Describe your changes**
Update django-allauth to latest version and adapt to changes. Remove the unused social settings and forms.

**Tasks**
- [x] Changelog


goapunk:
> if socialaccount is not used, then lets remove it also from the settings INSTALLED_APPS.
yes, good point!

> There is also a change on how the user is logged in after clicking the link `Verifying the email address by clicking on the link would no longer log you in, even in case of ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION = True.` from [changelog](https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#6421-2024-09-05). Good to add this in the changelog maybe? We need to inform also PM?

To me this reads like a bugfix, they fixed that `ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION=True` is now working again as intended after it broke in a previous release, so the intended behaviour still the same. I will test if the bug is on stage and if it behaves different to this PR. If it's the same I guess we don't need to tell PMs

goapunk:
@m4ra I tested the behaviour on stage against this PR and for both you get automatically logged in when verifying your email, so I think we missed the version of allauth where it broke :)

m4ra:
@goapunk I tested already on stage, and logs you in after verifying email. So should be fine

goapunk:
> @goapunk I tested already on stage, and logs you in after verifying email. So should be fine

ah, thanks!

m4ra (approved):
thanks!

goapunk:
@m4ra hmm, now the tests are failing :/ maybe I leave the social account stuff in for now ? We usually have the policy to remove unused code but also this might be out of scope for the release story

m4ra:
Leave it out yes. Maybe create an gh issue

goapunk:
created https://github.com/liqd/a4-meinberlin/issues/5828

m4ra (approved):

PR #5884, opened 2024-12-12
cherry-pick first batch of changes from main

goapunk:
**Describe your changes**
I cherry-picked the changes from `main` we want to bring to `dev`, these are mostly manual changes, I will open a second PR for dependency updates.

I'll just tag everyone involved here and maybe you can have a quick look at the list of commits of concern to you and say if any of them shouldn't be included here.

@vellip wasn't sure about [select_dropdown.scss: remove confusing greyed out first select item](https://github.com/liqd/a4-meinberlin/pull/5884/commits/bba78e110c58986698d6bd1e12f42bd1bae5b091) but seemed reasonable to have that on dev as well?

**Tasks**
- [x] Changelog


m4ra (approved):
python picks LGTM!

vellip:
I'd rather let @hom3mad3 have a look at the FE parts, as I haven't worked on these.

Regarding

> @vellip wasn't sure about [select_dropdown.scss: remove confusing greyed out first select item](https://github.com/liqd/a4-meinberlin/pull/5884/commits/bba78e110c58986698d6bd1e12f42bd1bae5b091) but seemed reasonable to have that on dev as well?

I think it doesn't matter too much to have that in because the style changed for `dev` anyway and I am not even sure it's being applied in `dev` because the structure of the SCSS changed. But it also shouldn't cause any harm to pull it in.

goapunk:
> I'd rather let @hom3mad3 have a look at the FE parts, as I haven't worked on these.
> 
> Regarding
> 
> > @vellip wasn't sure about [select_dropdown.scss: remove confusing greyed out first select item](https://github.com/liqd/a4-meinberlin/pull/5884/commits/bba78e110c58986698d6bd1e12f42bd1bae5b091) but seemed reasonable to have that on dev as well?
> 
> I think it doesn't matter too much to have that in because the style changed for `dev` anyway and I am not even sure it's being applied in `dev` because the structure of the SCSS changed. But it also shouldn't cause any harm to pull it in.

Thanks, I'll take it out then!

goapunk:
I'll go ahead and merge this so we can update a4 again. @hom3mad3 if you have any remarks I can fix them later!

goapunk:
@m4ra hmm, now the tests are failing but locally it works both on sqlite3 and postgresl, no idea what's wrong :/

m4ra:
Here same, tests pass both in sqlite3 and postgresql. I re-run the build.

goapunk:
> Same error. I don't remember why we have set 3 queries when requesting the plans. Do you? @goapunk

not really, I think that was just the number when we first ran the test. The confusing thing is it works on main, so I don't really know what's different on dev which breaks it? I will try to have a closer look at the queries

Issue #5849, opened 2024-12-04
#8471 (mB) Add Alerts according to the new Design - Info and Success Alerts overlap


sblindow:
**user:** 
Initiators, Admins

**expected behaviour:** 
Multiple Alerts that appear at the same time should not overlap.

**behaviour:**
On the page of an unpublished project, where there is an Info Alert pointing out that the Project is not currently published, other alerts, like "successfully logged in" directly overlap the Info Alert. 

**Screenshot**
This is a screenshot of the page. In the html inspector you can see that there are one info and one success alert on the page. However only the success alert is visible because it directly overlaps with the info alert.

---
<img width="1709" alt="Bildschirmfoto 2024-12-04 um 11 34 30" src="https://github.com/user-attachments/assets/9ea3d1c9-1708-4859-88eb-050cbf96badf">

CarolingerSeilchenspringer:
they used to be below and I guess this is what we need @mcastro-lqd ?

PR #5893, opened 2024-12-17
meinberlin_contrib/components: create info-box

hom3mad3:
# Refactor Alert to Status Info Box for Project Publication State
While investigating overlapping alerts for unpublished projects, identified that the current alert component was semantically incorrect for its use case.

fixes https://github.com/liqd/a4-meinberlin/issues/5849

info-box

## Changes
- Converted alert component to a static info box
- Removed dismissible functionality (close button)
- Updated ARIA semantics from `role="alert"` to `role="status"`
- Info box visibility now toggles based on project publication state

## Rationale
The element's primary purpose is to display the current publication state of a project, making it a status indicator rather than an alert. Alerts are typically used for important, time-sensitive messages that require user attention, while this component serves as a persistent status display.

### DIFF
--- a/changelog/_8473.md
+++ b/changelog/_8473.md
@@ -1,0 +1,5 @@
+### Changed
+- Converted alert component to a static info box `contrib/templates/meinberlin_contrib/components/`
+- Removed dismissible functionality (close button)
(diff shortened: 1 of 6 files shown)


sevfurneaux (approved):
Nice, lgtm!

PR #5895, opened 2024-12-17
map: remove color from icons in ideas map display

vellip:


vellip:
@m4ra let me know if we should change the a4 commit hash in this PR directly or not. Regardless it should also be in the release.

m4ra:
Yes, please change the commit hash of a4 so PMs can test it in dev.

vellip:
@m4ra Do you already have a release branch in a4?

m4ra:
No, I can do that quickly, one minute

m4ra:
@vellip you can update the a4 dependency here with the following `liqd/adhocracy4#mB-v2412.2`

vellip:
@m4ra  I just did. One second before merging though, because for some reason I still have the `aural` css class on comments.
CleanShot 2024-12-18 at 12 55 12@2x

vellip:
@m4ra sorry, I am a bit slow today... somehow the change that I did here https://github.com/liqd/adhocracy4/pull/1736/commits/ae6b657cd3a8a01f27637dcca4d89c91a4a44927 doesn't seem to be in my version of a4 after running `make install` even after setting the release branch as target for install. But maybe that's just a local issue with my installation?

vellip:
should be good now, thanks @m4ra

m4ra (approved):
Thanks, I will merge once workflow tests are done.

PR #5910, opened 2025-01-02
apps/bplan: remove char limit on name and description

goapunk:
**Describe your changes**
apps/bplan: remove char limit on name and description and instead truncate the input if too long

This means we won't save  the full name or description if it exceeds the limits. Otherwise we'd have to make a new model or change the char limit for all projects, not just bplans we don't want to do

**Tasks**
- [x] Changelog


m4ra (approved):
LGTM

PR #5928, opened 2025-01-09
[8664] auto searchprofile names

goapunk:
**Describe your changes**
Add code to automatically name searchprofiles as `Searchprofile 1`, `Searchprofile 2`, etc.

**Tasks**
- [x] Changelog


meinberlin/apps/kiezradar/serializers.py
```diff
@@ -103,3 +105,13 @@ def to_representation(self, instance):
...
+
+    def _set_number_and_name(self, validated_data):
+        """Calculate the next free number and set the name"""
+        user = validated_data["user"]
+        next_number = 1
+        if SearchProfile.objects.filter(user=user).exists():
+            latest = SearchProfile.objects.filter(user=user).latest("number")
+            next_number = latest.number + 1
+        validated_data["number"] = next_number
+        validated_data["name"] = "Searchprofile " + str(next_number)
```
goapunk, line 117: note to self: how do we translate this in the frontend?

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -102,4 +105,15 @@ def to_representation(self, instance):
             {"id": topic.id, "name": topics_enum(topic.code).label}
             for topic in instance.topics.all()
         ]
+        if not instance.name:
+            representation["name"] = _("Searchprofile %d") % instance.number
         return representation
+
+    def _set_number(self, validated_data):
```
goapunk, line 112: I'm adding the number here, but now unsure if in the models `save()` is better, what do you think?

meinberlin/apps/kiezradar/migrations/0005_alter_searchprofile_number_and_more.py
```diff
@@ -0,0 +1,36 @@
...
+
+def add_number_to_search_profiles(apps, schema_editor):
+    SearchProfile = apps.get_model("meinberlin_kiezradar", "SearchProfile")
+    for sp in SearchProfile.objects.all():
+        if sp.number is None:
+            next_number = 1
+            latest = SearchProfile.objects.filter(user=sp.user).latest("number")
+            if latest and latest.number:
+                next_number = latest.number + 1
+            sp.number = next_number
```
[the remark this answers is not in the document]
goapunk: ah, we should filter out objects where number is None anyway!

meinberlin/apps/kiezradar/models.py
```diff
@@ -100,8 +101,24 @@ class SearchProfile(models.Model):
...
+            if update_fields:
+                update_fields.append("number")
+        super().save(update_fields=update_fields, *args, **kwargs)
+
     class Meta:
-        ordering = ["name"]
+        ordering = ["number"]
+        constraints = [
+            models.UniqueConstraint("user", "number", name="unique-search-profile")
+        ]
```
m4ra, line 121: we also need an index on the number field.

meinberlin/apps/kiezradar/serializers.py
```diff
@@ -112,6 +114,8 @@ def to_representation(self, instance):
             {"id": topic.id, "code": topic.code, "name": topics_enum(topic.code).label}
             for topic in instance.topics.all()
         ]
-
         representation["query_text"] = instance.query.text if instance.query else ""
+
+        if not instance.name:
+            representation["name"] = _("Searchprofile %d") % instance.number
```
m4ra, line 120: Let's save the object with a name, from the customized save method, and remove this repr here.

meinberlin/apps/kiezradar/models.py
```diff
@@ -100,8 +101,24 @@ class SearchProfile(models.Model):
...
 
+    def save(self, update_fields=None, *args, **kwargs):
+        """Custom save() to add the next unused number per user"""
+        if self.number is None:
+            latest = None
+            try:
+                latest = SearchProfile.objects.filter(user=self.user).latest("number")
+            except SearchProfile.DoesNotExist:
+                pass
+            self.number = latest.number + 1 if latest else 1
```
m4ra, line 112: here we check if self has no id, which means is new, and give it the generated name. Which also means we don't need to have the name field optional

m4ra (changes requested):
Left few comments, and I think we shouldn't let the name optional, but save it with the auto-generated function.

meinberlin/apps/kiezradar/migrations/0007_alter_searchprofile_number_and_more.py
```diff
@@ -0,0 +1,41 @@
...
+        if sp.number is None:
+            latest = None
+            try:
+                latest = (
+                    SearchProfile.objects.filter(number__isnull=False)
+                    .filter(user=sp.user)
+                    .latest("number")
+                )
+            except SearchProfile.DoesNotExist:
+                pass
```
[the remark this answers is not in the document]
goapunk: `latest()` throws a `DoesNotExist` as  well according to docs, also happened to me in that migration: 
>Like [get()](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#django.db.models.query.QuerySet.get), earliest() and latest() raise [DoesNotExist](https://docs.djangoproject.com/en/5.1/ref/models/class/#django.db.models.Model.DoesNotExist) if there is no object with the given parameters.

meinberlin/apps/kiezradar/serializers.py
goapunk: how do we handle translations then?

meinberlin/apps/kiezradar/models.py
```diff
@@ -100,8 +101,24 @@ class SearchProfile(models.Model):
...
+            if update_fields:
+                update_fields.append("number")
```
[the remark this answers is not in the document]
goapunk: good catch, I'll remove it

meinberlin/apps/kiezradar/migrations/0007_alter_searchprofile_number_and_more.py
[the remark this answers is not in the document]
m4ra: ah ok, sorry for this, haven't worked with latest much

meinberlin/apps/kiezradar/serializers.py
m4ra: I didn't think of translations, you are right. Leave it in the serializer then, and we see later.

m4ra:
@goapunk looking at the story again, it asks that we always increase the index, even it existing profiles are deleted. But I cannot think of a simple way to do so, hence you may want to communicate this decision in the story for when the testing phase starts, they are aware.

goapunk:
> @goapunk looking at the story again, it asks that we always increase the index, even it existing profiles are deleted. But I cannot think of a simple way to do so, hence you may want to communicate this decision in the story for when the testing phase starts, they are aware.

will do, its already in the comments of the story as well

sevfurneaux:
As Janek has confirmed we'll use this implementation for the naming, I'll wait for this PR to be merged and check all is working in my [[#8661] Add 'Save Profile' button PR](https://github.com/liqd/a4-meinberlin/pull/5940).

m4ra (approved):
thank you!

PR #5933, opened 2025-01-14
[#8681] notifications-settings: add new notifications page

vellip:
**Describe your changes**
Implements a page for setting user notifications 
CleanShot 2025-01-14 at 13 19 22

**Tasks**
- [x] Changelog

cc @goapunk @m4ra no need to create a view for it, I already did that. I also added an example of what the data structure looks like the frontend code currently works with in `notifications.html`

### DIFF
--- a/meinberlin/react/account/NotificationsList.jsx
+++ b/meinberlin/react/account/NotificationsList.jsx
@@ -1,0 +1,126 @@
+import React, { useState } from 'react'
+import django from 'django'
+import { ToggleSwitch } from '../contrib/ToggleSwitch'
+import NotificationToggle from './NotificationToggle'
+
+const notifications = [
+  {
+    header: django.gettext('Project-Related Notifications'),
+    notifications: {
+      email_newsletter: {
+        title: django.gettext('E-Mail Newsletter'),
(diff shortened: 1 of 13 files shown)


meinberlin/react/account/NotificationsList.jsx
```diff
@@ -0,0 +1,127 @@
...
+import NotificationToggle from './NotificationToggle'
+
+const notifications = [
+  {
+    header: django.gettext('Project-Related Notifications'),
+    notifications: {
+      email_newsletter: {
+        title: django.gettext('E-Mail Newsletter'),
+        description: django.gettext('Receive newsletters with updates and news about the projects you follow via e-mail.'),
+        showInActivityFeed: true
```
goapunk, line 13: In the design the email newsletter has no `Show in activity feed` button which kind of makes sense for an email I think ? :)

goapunk (changes requested):
Looks good to me, just a note regarding the email newsletter.

meinberlin/react/account/NotificationsList.jsx
vellip: ah good catch, thanks D:

goapunk (approved):
thanks!

PR #5951, opened 2025-01-16
apps/bplan: Replace imghdr with python-magic

partizipation:
`imghdr` support was deprecated in python, so this uses `python-magic` and built-in `mime` to check the filetype of images files for bplans created using the API.

Also reorders some conditionals to fix a bug where bug tile_image was being used instead of image_url.

Test with curl like so:
```
curl  -X POST http://127.0.0.1:8003/api/organisations/1/bplan/ \
 --user 'admin@liqd.net':'password' \
 -H "Content-Type: application/json" \
 -d \
 '
 {
   "name":"Luisenblock Ost - Bebauungsplan 1-70",
   "bplan_id": "VI - 96a",
   "description": "Test",
   "url": "https://mein.berlin.de",
   "office_worker_email": "test@example.com",
   "start_date": "2019-01-01T00:00",
   "end_date": "2022-01-01T00:00",
   "image_url": "https://cdn.britannica.com/39/226539-050-D21D7721/Portrait-of-a-cat-with-whiskers-visible.jpg"
 }
'
```

The new bplan should be at [http://127.0.0.1:8003/dashboard/organisations/liqd/bplans/](http://127.0.0.1:8003/dashboard/organisations/liqd/bplans/)

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/bplan/serializers.py
+++ b/meinberlin/apps/bplan/serializers.py
@@ -1,10 +1,11 @@
 import base64
 import datetime
-import imghdr
+import mimetypes
 import posixpath
 import tempfile
 from urllib.parse import urlparse
 
+import magic
 import requests
 from django.apps import apps
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


meinberlin/apps/bplan/serializers.py
```diff
@@ -327,14 +329,13 @@ def _generate_image_filename(self, url_path, file):
 
         root_path, extension = posixpath.splitext(url_path)
         if file:
-            # Workaround: imghdr expects the files position on 0
-            file.seek(0)
-            extension = imghdr.what(file) or "jpeg"
+            file_mime = magic.from_buffer(file.read(), mime=True)
```
goapunk, line 332: did you check that we don't need the `file.seek(0)` before the `file.read()` because the reason for that shouldn't have changed I think.

goapunk (changes requested):
Nice! Just some small comments

partizipation:
Tested that it actually identifies filetype with jpeg, png, and gif. It does indeed need the file seek 0.

goapunk (approved):
Thanks!

PR #5993, opened 2025-01-28
[#8769] Kiezradar: Allow creation and editing of kiezradars

sevfurneaux:
**Describe your changes**
This PR adds the kiezradar filter list, new and editing pages.

**Tasks**
- [X] Changelog


meinberlin/apps/kiezradar/templatetags/react_kiezradar_filters.py
```diff
@@ -8,10 +8,12 @@
 
 
 @register.simple_tag()
-def react_kiezradar_search_profiles():
+def react_kiezradar_filters():
     attributes = {
         "apiUrl": reverse("kiezradar-list"),
         "planListUrl": reverse("meinberlin_plans:plan-list"),
+        "kiezradarFiltersUrl": "/account/kiezradar/",
```
sevfurneaux, line 15: @vellip @hom3mad3 these URLs are currently hardcoded, as I had no idea how to get the Django `reverse()` URL for this 🤨.

The `reverse()` URLs should be based on `urls.py` (`kiezradar_filters` and `kiezradar_new`):

```
urlpatterns = [
    path(
        "kiezradar/",
        views.KiezRadarListView.as_view(),
        name="kiezradar_filters",
    ),
    path(
        "kiezradar/<int:id>/",
        views.KiezRadarListView.as_view(),
        name="kiezradar_edit",
    ),
    path(
        "kiezradar/new/",
        views.KiezRadarListView.as_view(),
        name="kiezradar_new",
    ),
    ...
```

meinberlin/react/kiezradar/Kiezradars.jsx
```diff
@@ -0,0 +1,62 @@
...
+          />
+          <Route
+            path={props.kiezradarNewUrl}
+            element={<NewKiezradar {...props} />}
+          />
+          <Route
+            path={props.kiezradarFiltersUrl + ':id/'}
+            element={<EditKiezRadar {...props} />}
+          />
+        </Routes>
```
vellip, line 58: Not sure if we need a router here. I think its better to go route by route and not let django serve the same view for all cases but have one for each. Not having to load map JS on a list view where no map will ever be displayed is one reason why.

meinberlin/react/kiezradar/Kiezradars.jsx
```diff
@@ -0,0 +1,62 @@
...
+import NewKiezradar from './NewKiezradar'
+
+const titleText = django.gettext('Kiez selection')
+const descriptionText = django.gettext(
+  'Set up your Kiez to get notified when new projects in your area get published.'
+)
+const manageKiezesText = django.gettext('Manage Kiezes')
+const createKiezText = django.gettext('Create a Kiez')
+
+export default function Kiezradars (props) {
```
vellip, line 16: Can you streamline whether it's `Kiezradar`, `KiezRadar`or `kiezradar`? And if you use `KiezRadar` also reflect it in css like so `kiez-radars__links`?

meinberlin/apps/kiezradar/templatetags/react_kiezradar_filters.py
hom3mad3: @goapunk @partizipation

meinberlin/react/kiezradar/Loading.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import django from 'django'
+
+const loading = django.gettext('loading')
+
+export default function Loading () {
+  return (
+    <div className="kiezradars__loading" aria-label={loading}>
+      {loading}...
```
vellip, line 9: We have a spinner, or why did you choose not to use that?

meinberlin/react/kiezradar/KiezradarList.jsx
```diff
@@ -0,0 +1,134 @@
...
+          </>
+          )}
+    </>
+  )
+}
+
+function EditButtonLink ({ to }) {
+  return (
+    <Link to={to} className="kiezradar-list__button">
+      <i className="fa-solid fa-pencil mr-1" />
```
vellip, line 130:
```suggestion
      <i className="fa-solid fa-pencil mr-1" aria-hidden="true" />
```

meinberlin/react/kiezradar/KiezradarList.jsx
```diff
@@ -0,0 +1,134 @@
...
+            <Link to={kiezradarNewUrl} className="button">
+              {addKiezText}
+            </Link>
+          </div>
+          )
+        : (
+          <>
+            <div className="kiezradar-list">
+              {kiezradars.map((kiezradar) => (
+                <div key={kiezradar.id} className="kiezradar-list__kiezradar">
```
vellip, line 80: See if you can re-use the scss component `actionable-list`, i created it for cases like these

meinberlin/assets/scss/components_user_facing/_kiezradar.scss
```diff
@@ -0,0 +1,141 @@
...
+}
+
+.kiezradar-list__buttons {
+  display: flex;
+  gap: 1.5rem;
+}
+
+.kiezradar-list__button:link,
+.kiezradar-list__button:visited {
+  color: $black;
```
hom3mad3, line 99: minor detail, but we could use `$text-base` here

meinberlin/assets/scss/components_user_facing/_kiezradar.scss
```diff
@@ -0,0 +1,141 @@
...
+    display: none;
+  }
+}
+
+.kiezradar-list__footer-buttons div:only-of-type {
+  margin-left: auto;
+}
+
+.kiezradar__error {
+  color: $danger;
```
hom3mad3, line 132: we could use `$text-error` here

meinberlin/react/kiezradar/KiezradarList.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+  if (loading) {
+    return <Loading />
+  }
+
+  if (error) {
+    return (
+      <div className="kiezradar__error">
+        {errorText}: {error}
+      </div>
```
vellip, line 55: The `alert` component from a4 should have correct styling now, maybe you can use it instead.

meinberlin/react/kiezradar/Kiezradar.jsx
```diff
@@ -0,0 +1,85 @@
...
+      setError(err.message)
+    } finally {
+      setLoading(false)
+    }
+  }
+
+  return (
+    <>
+      {(error) && (
+        <div className="kiezradar__error">{errorText + ': ' + error}</div>
```
vellip, line 57: Can you use `Alert` from a4?

vellip (changes requested):
Thanks for the first part of that story! Great work, I just left a few remarks where you might want to have a second look.

meinberlin/react/kiezradar/KiezradarList.jsx
```diff
@@ -0,0 +1,134 @@
...
+            <p>{noSavedKiezradarsText}</p>
+            <Link to={kiezradarNewUrl} className="button">
+              {addKiezText}
+            </Link>
+          </div>
+          )
+        : (
+          <>
+            <div className="kiezradar-list">
+              {kiezradars.map((kiezradar) => (
```
hom3mad3, line 79: i'n not sure this is on purpose, but should we wrap the elements in a list and overall improve semantics here?

meinberlin/react/kiezradar/Kiezradars.jsx
```diff
@@ -0,0 +1,62 @@
...
+  const location = useLocation()
+  const isNewKiez = location.pathname.includes(props.kiezradarNewUrl)
+
+  return (
+    <>
+      <h1>{titleText}</h1>
+      <p>{descriptionText}</p>
+      <div className="kiezradars">
+        <div className="kiezradars__links">
```
hom3mad3, line 25: is this supposed to be some sort of nav element?

meinberlin/react/kiezradar/Loading.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import django from 'django'
+
+const loading = django.gettext('loading')
+
+export default function Loading () {
+  return (
+    <div className="kiezradars__loading" aria-label={loading}>
```
hom3mad3, line 8: do you think we need aria-live or some way to hand live status here? @vellip your thoughts?

meinberlin/react/kiezradar/KiezradarList.jsx
hom3mad3: @vellip thanks for the reminder

meinberlin/react/kiezradar/Loading.jsx
sevfurneaux: When things load very quickly, the `<Spinner />` flashes as a black circle on the screen, and then the loaded UI, which feels kinda broken. Can add it back in, it just feels like a disjointed experience.

meinberlin/assets/scss/components_user_facing/_kiezradar.scss
sevfurneaux: > i'm confused now, how come you are able to access the dashboard variables if you're loading the styles in style_user_facing without getting any errors 🤔

No idea 😅, maybe that CSS is being incorrectly compiled as part of this? Thanks though, I'll add `$text-error`.

meinberlin/react/kiezradar/Kiezradars.jsx
sevfurneaux: Thanks. By doing route by route, presumably we'd need an `...init.jsx` files for each route, and they're server rendered via Django? So:

`react_kiezradar_filters_init.jsx`
`react_kiezradar_new_init.jsx`
`react_kiezradar_edit_init.jsx`

One benefit of using the router, is it is _super_ snappy between page clicks, as it is SPA-like.

> Not having to load map JS on a list view where no map will ever be displayed is one reason why.

That could help, but isn't it very likely a user would navigate from the list view to create a kiezradar where the map is? I was just thinking if all in same bundle, then yes initially slower first load but then the js is browser cached.

vellip: That's correct.  I don't think however, that every or nearly every page visit will lead the user further to the edit/create page as we also have the "view projects" button. You can leave it but then I would suggest to at least:

- revert the django views so that each url has their own view again and they don't all point to the same view
- make the different django views serve the same template

or

- rename the listview as its not just a list view anymore
- delete the other views if they are not needed

and in either case try to use dynamic imports for `EditKiezRadar` and `NewKiezradar`.

sevfurneaux: Thanks, sounds good! Good idea with dynamic imports. I just had an idea: why don't we make the map a dynamic import? So rather than dynamically importing `EditKiezRadar` and `EditKiezRadar`, it would be the most expensive thing, map itself:

```js
const Map = React.lazy(() => import("./Map"));

const EditKiezRadar = () => {
  return (
      {...}
      <Suspense fallback={<div>Loading component...</div>}>
        <Map />
      </Suspense>
  );
};
```

meinberlin/apps/kiezradar/templatetags/react_kiezradar_filters.py
goapunk: Without having tested it, afair it should just be `reverse("<name of the view>")`, did  you test e.g. `reverse("kiezradar_new")` ? :)

meinberlin/assets/scss/components_user_facing/_kiezradar.scss
hom3mad3: yes, i will have a look at it. it should not allow you to add dashboard variables.

meinberlin/apps/kiezradar/templatetags/react_kiezradar_filters.py
sevfurneaux: Aha, that works! I was following convention and trying the prefix `reverse("meinberlin_kiezradar:kiezradar_new")` 🤦‍♂️

meinberlin/assets/scss/components_user_facing/_kiezradar.scss
sevfurneaux: Thanks, that is now added 👍: https://github.com/liqd/a4-meinberlin/pull/5993/files#diff-15e8f498e6bcb34a40d5ae33de2ddb295751ba1bca2ef508530af3c8ea44cb4dR110

meinberlin/react/kiezradar/Kiezradars.jsx
sevfurneaux: @vellip I went ahead based on what I mentioned above and added a placeholder `KiezradarMap` component via dynamic import: https://github.com/liqd/a4-meinberlin/pull/5993/files#diff-20f519237b6ded6ff08a3069341129231b59545ac49280e62df5f26d12f1b2afR7

meinberlin/react/kiezradar/Kiezradars.jsx
sevfurneaux: I've streamlined as `Kiezradar` in the components and the CSS 👍

meinberlin/react/kiezradar/KiezradarList.jsx
sevfurneaux: I tried the `actionable-list` class, but it is _slightly_ different than the kiezradar list items, which have different borders (for example: no border at the top).

meinberlin/react/kiezradar/KiezradarList.jsx
sevfurneaux: Added ✅

meinberlin/react/kiezradar/Kiezradar.jsx
sevfurneaux: Added ✅

meinberlin/react/kiezradar/KiezradarList.jsx
sevfurneaux: @hom3mad3 Changed thanks to your suggestion in chat!

meinberlin/react/kiezradar/Loading.jsx
sevfurneaux: @hom3mad3 I added `aria-polite`, what do you think?

meinberlin/react/kiezradar/Kiezradars.jsx
vellip: Looks good to me 👍🏽

meinberlin/react/kiezradar/KiezradarList.jsx
vellip: Feel free to add a modifier to it or let the designer know that there is an inconsistency and they should update it to look the same like i.e. the email user setting page

meinberlin/react/kiezradar/Loading.jsx
vellip: `aria-live` should exist at the top level, not the loading indicator. It only announces when descendants change so putting it on a loading indicator will not announce any change as it will just always be the same text. You will want to...

- set the innermost dom node around dynamic content in a component that does fetching to `aria-live="polite"`
```jsx
return (
  <>
    <h1>Title</h1>
    <p>Text</p>
    <div aria-live="polite">
      {data.length && (
        <ul>
          {data.map(i => <li>{i}</li>)}
        </ul>
      )}
    </div>
  </>
)
```
- have a loading indicator **within** the live-region
- have that loading indicator not be inserted on first render (because a live-region only announces changes to the DOM)
- instead have it be rendered in once you start the fetching in the useEffect (basically, initialise `isLoading` with `false` and call `setIsLoading` in the `useEffect` when fetching)

meinberlin/react/kiezradar/KiezradarList.jsx
sevfurneaux: Thanks, I'll let Janek/Carolin know before changing things as the kiezradar layouts were approved by them.

meinberlin/react/kiezradar/Loading.jsx
hom3mad3: @vellip thank you for the explanation, i learned a lot ❤️

sevfurneaux: Thanks @vellip, this is now updated 👍

sevfurneaux:
@vellip @hom3mad3 @goapunk got the frontend build passing 🟢, but now there's a coverage/coveralls error 🔴 as the coverage decreased by `-0.005%` 😄. The .py changes were just renaming things:

- [react_kiezradar.py](https://github.com/liqd/a4-meinberlin/pull/5993/files#diff-098ef49b05373fa1d246a5f94ec626c8bc84e776c552320710e55e36e586db01)
- [urls.py](https://github.com/liqd/a4-meinberlin/pull/5993/files#diff-27cf7ae5163b766bfce8723d02f0d3bc5340a53ba1c0e2378e22f7d898ded4a7)
- [views.py](https://github.com/liqd/a4-meinberlin/pull/5993/files#diff-dfeb079c059030fdd5257654a553a4fbc6b479d7e6b1a19e015a06b695d96542)

Can we bypass this?

hom3mad3 (approved):
@sevfurneaux thank you for the great work and patience with all those comments, it looks great! ✨ 🤗

sevfurneaux:
> @sevfurneaux thank you for the great work and patience with all those comments, it looks great! ✨ 🤗

A pleasure! Thanks for both your comments, @hom3mad3 and @vellip: learnt a lot 🫶

hom3mad3:
> Thanks for the first part of that story! Great work, I just left a few remarks where you might want to have a second look.

@vellip, when you have some time, could you give it a thumbs-up? I don’t want to dismiss your feedback. Even though we don’t have a code of conduct (which isn't ideal), I want to make sure we're being considerate and giving everyone a final word.

vellip (approved):
Looks good, thank you so much!

hom3mad3:
> > @sevfurneaux thank you for the great work and patience with all those comments, it looks great! ✨ 🤗
> 
> A pleasure! Thanks for both your comments, @hom3mad3 and @vellip: learnt a lot 🫶

same here! 📙🤓

PR #6030, opened 2025-02-03
Add hero and project information page  

hom3mad3:
### **Description:**
This PR combines two originally separate stories due to their high interdependence:
- [#8765] Redesign frontend: Opener: Image with Aside and detail Project 
- [#8766] Redesign frontend & backend - More information page (project detail page)

### **Hero Project Page**
| Desktop View                                         | Mobile View                                           |
|------------------------------------------------------|-------------------------------------------------------|
| Desktop View | Mobile View |

### **Hero without buttons | Plan Page**
Screenshot 2025-02-03 at 10-11-48 Radfahrstreifen Sonnenallee S-Bahnhof – meinBerlin

### **Project information page**
Screenshot 2025-02-03 at 14-34-22 Stadtpark Neugestaltung 2025 -Information

### **Contact info section (updated in plan detail and project information)**
Screenshot 2025-02-03 at 14-43-31 Stadtpark Neugestaltung 2025 -Information

### **Changes**
- Added new reusable hero snippet template and styles
- Added new project information page
- Cleaned up project_detail.html and plan_detail.html - Hero was adapted to the respective use cases
- Streamlining contact section for maintainability, added a few flags in https://github.com/liqd/adhocracy4/pull/1768
- Restyled follow button


meinberlin/apps/plans/models.py
```diff
@@ -43,6 +43,11 @@ class Plan(ProjectContactDetailMixin, UserGeneratedContentModel):
         max_length=120,
         verbose_name=_("Title of your plan"),
     )
+
+    @property
```
hom3mad3, line 47: @m4ra this is to deal with naming inconsistency in our models so i can create a reusable snippet. 
1. no idea whether this makes sense. 2. not sure it should go at the end of the file, as i kept it closest to the object it refers to.

meinberlin/apps/contrib/templates/meinberlin_contrib/components/hero.html
```diff
@@ -0,0 +1,48 @@
+{% load i18n thumbnail %}
+
+<article class="hero {{ extra_class|default:'' }}">
+  <div class="hero__container">
+    <div class="hero__content">
+        <h1 class="hero__title">
+          {% if show_buttons %}
+              <a href="{{ hero_link }}" class="hero__title-link">{{ content.name }}</a>
+          {% else %}
+              {{ content.name }}
```
hom3mad3, line 10: @m4ra here is where i needed to use `name`

meinberlin/apps/plans/templatetags/plan_tags.py
```diff
@@ -0,0 +1,13 @@
+from django import template
+
+register = template.Library()
+
+@register.simple_tag
+def modify_hero_content(content):
```
hom3mad3, line 6: for the plan detail page, we need to replace the content of the with another object (point_label) - again brings the lack of consistency in our naming throughout the apps. probably something to add to a documentation/ set of rules whenever we create a new app/model.

meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
```diff
@@ -0,0 +1,59 @@
...
+        <h2 id="description-title">
+          <span class="aural">{% trans 'Project Description' %}</span>
+          {{ project.name }}
+        </h2>
+        <div>{{ project.information|safe }}</div>
+      </section>
+
+      {% if project.has_any_contact_info %}
+        <hr aria-hidden="true" />
+        <section class="flexgrid grid--2 mb-3" aria-labelledby="contact-title">
```
hom3mad3, line 51: im tempted to create another snippet for the contact section, unsure since it adds another level of data handling

meinberlin/apps/projects/views.py
```diff
@@ -384,3 +384,29 @@ def get_context_data(self, **kwargs):
...
+    @cached_property
+    def project(self):
+        return self.get_object()
+        
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['project'] = self.project
+        # TODO: Refactor breadcrumbs to use a global dynamic system to avoid redundancy 
+        # and ensure better maintainability. Consider moving to a context processor or view mixin.
+        context['breadcrumbs'] = [
```
hom3mad3, line 406: In this update, I’m introducing dynamic breadcrumbs specifically for this template. right now, all breadcrumbs are hardcoded throughout the project, which isn't scalable or maintainable.

To improve this, we should create a global solution that handles breadcrumbs dynamically.

In the meantime, is that a good option?

meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
```diff
@@ -0,0 +1,59 @@
...
+              <a href="{{ crumb.url }}">{{ crumb.title }}</a>
+            </li>
+          {% else %}
+            <li aria-current="page">{{ crumb.title }}</li>
+          {% endif %}
+        {% endfor %}
+      </ol>
+    </nav>
+  </div>
+{% endblock breadcrumbs %}
```
vellip, line 32: This is a different breadcrumbs html than what we normally have. For consistency I'd suggest we stick with something like (example from a different page):
```
{% block breadcrumbs %}
    <div id="content-header">
        <nav class="breadcrumb" aria-label="{% translate 'You are here:' %}">
            <ol>
                <li>
                    <a href="{% url 'wagtail_serve' '' %}">meinBerlin</a>
                </li>
                <li>
                    <a href="{% url 'meinberlin_plans:plan-list' %}">{% translate 'Project Overview' %}</a>
                </li>
                <li>
                    <a href="{% url 'project-detail' project.slug %}">{{ project.name|truncatechars:50 }}</a>
                </li>
                {% if module.is_in_module_cluster %}
                    <li>
                        <a href="{{ module.get_detail_url }}">{{ module.name|truncatechars:50 }}</a>
                    </li>
                {% endif %}
                <li class="active" aria-current="page">{{ object.name }}</li>
            </ol>
        </nav>
    </div>
{% endblock breadcrumbs %}
```

meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
vellip: I think it's fine to leave as is. Can always refactor if we notice it being used a ton.

meinberlin/apps/projects/views.py
```diff
@@ -384,3 +384,29 @@ def get_context_data(self, **kwargs):
...
+        context = super().get_context_data(**kwargs)
+        context["project"] = self.project
+        # TODO: Refactor breadcrumbs to use a global dynamic system to avoid redundancy
+        # and ensure better maintainability. Consider moving to a context processor or view mixin.
+        context["breadcrumbs"] = [
+            {"title": _("meinBerlin"), "url": "/"},
+            {"title": _("Kiezradar"), "url": "/kiezradar/"},
+            {"title": self.project.name, "url": self.project.get_absolute_url()},
+            {"title": _("Project Information"), "url": "#"},
+        ]
```
vellip, line 411: As mentioned before, I think sticking with the hard-coded breadcrumbs in HTML is better than opening up another variant right now. That's also nicer with URL generation, since you don't want to use hardcoded paths in django.

meinberlin/apps/projects/views.py
```diff
@@ -384,3 +384,29 @@ def get_context_data(self, **kwargs):
...
+    def get_permission_object(self):
+        return self.get_object()
+
+    @cached_property
+    def project(self):
+        return self.get_object()
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["project"] = self.project
```
vellip, line 403: the project will already be available in the template via `object`. You can rename that variable with `context_object_name = "project"` where defining the class attributes.

meinberlin/assets/scss/components_user_facing/_hero.scss
```diff
@@ -0,0 +1,104 @@
+.hero {
+  width: 100%;
+  background-color: $gray-lightest;
+}
+
+.hero__container {
+  max-width: 980px;
+  min-height: 352px;
```
vellip, line 8: This seems a bit like a magical number. Does it exactly need to be `352`? Maybe refactor it out into a local scss variable? Also makes it less "magical".

meinberlin/assets/scss/components_user_facing/_hero.scss
```diff
@@ -0,0 +1,104 @@
...
+  padding: 1rem;
+}
+
+.hero__title {
+  margin-top: 0;
+  margin-bottom: 10px;
+}
+
+.hero__title-link {
+  color: $text-base!important;
```
vellip, line 26: Usually I increase specificity of the selector by nesting more, when trying to override BO styles. Would that be an option? Or do you need to override an id selector?

meinberlin/assets/scss/components_user_facing/_hero.scss
```diff
@@ -0,0 +1,104 @@
...
+@media (min-width: $breakpoint-tablet) and (max-width: $breakpoint-desktop) {
+  .hero__content {
+    margin-left: 2em;
+  }
+}
+
+@media (min-width: $breakpoint-tablet) {
+  .hero {
+    width: 100vw;
+    height: 352px;
```
vellip, line 69: Same as above, maybe re-use a local scss variable

meinberlin/assets/scss/components_user_facing/adhocracy4/_a4-follow.scss
```diff
@@ -1,29 +1,27 @@
...
 .a4-follow__notification {
     display: none;
 }
 
-.follow__container {
-    display: flex;
-    align-items: end;
-    margin-bottom: 2 * $spacer;
+.a4-follow__btn--content:after {
+    font-family: "Font Awesome 5 Free", sans-serif;
```
vellip, line 16: Didn't we upgrade to v6 some time ago?

meinberlin/apps/plans/models.py
vellip: I think, maybe it makes more sense to use `firstof` in the template. https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#firstof

You can even store that in a `variable` via  `firstof ... as ...`.

vellip: Otherwise maybe it makes sense for liqd, to consider a story to streamline naming issues like these in the future.

meinberlin/apps/contrib/templates/meinberlin_contrib/components/hero.html
```diff
@@ -0,0 +1,48 @@
+{% load i18n thumbnail %}
+
+<article class="hero {{ extra_class|default:'' }}">
```
vellip, line 3: maybe you can add a `block` class for some vertical space below? Or how did you plan on achieving that?

meinberlin/apps/projects/templates/meinberlin_projects/includes/contact_person.html
```diff
@@ -0,0 +1,22 @@
...
+      {% if contact.contact_email %}
+        <p>{% trans "Email" %}: 
+           <a href="mailto:{{ contact.contact_email }}" aria-label="{% trans 'Email' %} {{ contact.contact_name }}">{{ contact.contact_email }}</a>
+        </p>
+      {% endif %}
+      
+      {% if contact.contact_phone %}
+        <p>{% trans "Phone" %}: 
+           <a href="tel:{{ contact.contact_phone }}" aria-label="{% trans 'Call' %} {{ contact.contact_name }}">{{ contact.contact_phone }}</a>
+        </p>
```
vellip, line 18: does it make sense to use a `dl` for these lists?

vellip (approved):
LGTM, thanks for the large changes! I'd suggest waiting for @m4ra to also have a look regarding the django code.

meinberlin/assets/scss/components_user_facing/_hero.scss
hom3mad3: it was from the design, and for heights/padding/margins i usually don't like steering too much away from the original. i can make it 350px if that feels more rounded.

meinberlin/apps/contrib/templates/meinberlin_contrib/components/hero.html
hom3mad3: I think we should handle spacing in the parent container rather than within the blocks themselves. Managing it at the block level isn’t really best practice, especially with everything being block-based these days. It often causes issues when dealing with footers or other unexpected elements.

One possible approach could be using the lobotomized owl selector (* + *) for consistent vertical spacing, though there might be better solutions worth exploring. Since the hero is always meant to sit at the top, your suggestion works for now. Figuring out a more global approach to vertical space after the release would definitely be a worthwhile step.

meinberlin/apps/projects/views.py
hom3mad3: Fair, I'll create an issue for dynamic global handling of breadcrumbs after the release.

meinberlin/apps/projects/views.py
hom3mad3: That's a relief! Not a big fan of this solution...

meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
hom3mad3: This proves the point for the need of a breadcrumb component 🙈

meinberlin/assets/scss/components_user_facing/adhocracy4/_a4-follow.scss
hom3mad3: ok, then i'll have a look, something might not be right with my setup..thanks for pointing it out

m4ra (dismissed):
That's a biggie indeed! Thanks for all the refactoring @hom3mad3 
Few things need still some work. See my inline comments.

meinberlin/apps/plans/models.py
hom3mad3: > all methods go under the class Meta attribute, so below line 131

thanks for confirming :)

meinberlin/apps/plans/templates/meinberlin_plans/plan_detail.html
```diff
@@ -157,58 +124,12 @@ <h1 class="title herounit-image-with-aside__title">{{ object.title }}</h1>
...
-                                <a target="_blank" href="{{ object.organisation.url }}">
-                                    {% translate 'Website' %}
-                                </a>
-                            </p>
-                        {% endif %}
-                    </address>
-                </div>
-            {% endif %}
-            </div>
+        {% if object.has_any_contact_info %}
```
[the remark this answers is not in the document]
hom3mad3: we should only show the title if one of the options is being displayed.

meinberlin/apps/plans/templatetags/plan_tags.py
hom3mad3: @m4ra Thanks for the feedback — I get that plans and projects have different needs. What I meant by 'lack of consistency' is more about standardizing shared building blocks across models. Right now, they both have a name/title, a description, and sometimes a short description, but we're naming these differently (title vs. name, description vs. details, etc).

It’d be helpful if we kept these core properties consistent across models. That way, the code becomes more predictable and easier to work with while still giving space for each model’s unique specifics. Probably something worth documenting as a guideline for future apps.

hom3mad3: In this particular case, we also need to do some frontend magic because the design isn’t aware of the different building blocks.

hom3mad3:
> That's a biggie indeed! Thanks for all the refactoring @hom3mad3 Few things need still some work. See my inline comments.

@m4ra thank you for going over this long one, and for the super insightful feedback 🤗

meinberlin/apps/contrib/templates/meinberlin_contrib/components/hero.html
```diff
@@ -0,0 +1,48 @@
...
+
+    <div class="hero__media">
+      <figure class="hero__figure">
+        {% if show_buttons %}
+            <a href="{{ hero_link }}">
+        {% endif %}
+        <img
+          class="hero__image"
+          src="{{ content.image|thumbnail_url:'heroimage' }}"
+          alt="{{ content.image_alt_text|default:_('Here you can find a decorative picture.') }}"
```
vellip, line 39: I think it's better to leave this out, right? IIRC we also said that elsewhere.

meinberlin/apps/contrib/templates/meinberlin_contrib/components/hero.html
```diff
@@ -0,0 +1,48 @@
...
+
+    <div class="hero__media">
+      <figure class="hero__figure">
+        {% if show_buttons %}
+            <a href="{{ hero_link }}">
+        {% endif %}
+        <img
+          class="hero__image"
+          src="{{ content.image|thumbnail_url:'heroimage' }}"
+          alt="{{ content.image_alt_text|default:'' }}"
```
vellip, line 39: 👌🏽

meinberlin/apps/projects/views.py
```diff
@@ -384,3 +384,20 @@ def get_context_data(self, **kwargs):
...
+    def get_permission_object(self):
+        return self.get_object()
+
+    @cached_property
+    def project(self):
+        return self.get_object()
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        return context
```
vellip, line 403: you can remove this if it's not needed.

I am not sure about the other two functions, they might work as is as well.

```python
    def get_permission_object(self):
        return self.get_object()
```
is the default afaik

and

```python
    @cached_property
    def project(self):
        return self.get_object()
```
is not used anywhere, is it?

meinberlin/assets/scss/components_user_facing/_hero.scss
```diff
@@ -0,0 +1,106 @@
+$hero-height: 350px;
```
vellip, line 1: nice

vellip (approved):
Looks good to me. I just left one comment regarding the view. After that feel free to merge, unlees you want to wait for @m4ra ofc.

meinberlin/apps/projects/views.py
hom3mad3: Cool, it makes the code cleaner! I actually just learned that `PermissionRequiredMixin` defaults to get_object() for permissions, so I simplified it and kept only the basic settings.

PR #6031, opened 2025-02-03
[#8750] Add Kiezradar map

sevfurneaux:
**Describe your changes**
This PR adds the Kiezradar map.

**Tasks**
- [X] Changelog


meinberlin/apps/kiezradar/serializers.py
```diff
@@ -14,7 +16,7 @@
 
 
 class KiezRadarSerializer(serializers.ModelSerializer):
-    point = serializers.JSONField(required=False, write_only=True, binary=True)
+    point = serializers.JSONField(required=False)
```
sevfurneaux, line 19: @goapunk the `point` property wasn't returning in the API response, so had to remove these for it to display in the response. I don't know much Python, so let me know if this makes sense 😅

meinberlin/react/kiezradar/Kiezradar.jsx
```diff
@@ -15,21 +20,38 @@ const errorLimitedExceededText = django.gettext(
...
+  const handleLocationChange = (location) => {
+    locationRef.current = location
+  }
 
   const handleSubmit = async (e) => {
     e.preventDefault()
     setLoading(true)
     setError(null)
 
+    const geojsonPoint = {
```
sevfurneaux, line 42: @goapunk when saving the `point` property, I'm doing so in this format:

```
    const geojsonPoint = {
      type: 'Feature',
      geometry: {
        type: 'Point',
        coordinates: locationRef.current.point
      }
    }
```

I referenced this format from the tests: is this correct?

meinberlin/react/kiezradar/Kiezradars.jsx
```diff
@@ -46,26 +87,66 @@ export default function Kiezradars (props) {
...
           />
         </Routes>
       </div>
     </>
   )
 }
+
+function Alert ({ title, message, onClose }) {
+  const htmlMessage =
+    '<p><strong>' + title + '</strong><br />' + message + '</p>'
```
sevfurneaux, line 143: @vellip @hom3mad3 Unless I'm mistaken, the a4 `Alert` doesn't have a `title` prop:

`Alert = ({ type = 'info', message, onClick, timeInMs })`

...so I'm passing HTML into the `message` prop like this. A little hacky, but maybe we can add a `title` prop to `Alert` in the future?

hom3mad3: Good catch! Since the new alert designs all have title options, this should be reflected in the A4 React component as well. How much effort would it take to adjust the actual component? If it's too much for now, could you add a comment explaining your intent so it’s easier to extract and refactor later? Thanks!

meinberlin/react/kiezradar/KiezradarMap.jsx
```diff
@@ -1,5 +1,145 @@
...
+        style={{ minHeight: '100%', height: '100%' }}
+      >
+        <Radius
+          point={point}
+          radius={radius}
+          minRadius={minRadius}
+          maxRadius={maxRadius}
+          onChange={onChange}
+        />
+        <SearchAndShowAddress apiUrl="https://bplan-prod.liqd.net/api/addresses/" />
```
sevfurneaux, line 48: @vellip I've added the `SearchAndShowAddress` here, but currently it is creating a new marker alongside the loaded radius based marker:

Screenshot 2025-02-04 at 08 53 43

In a future PR, maybe we introduce an optional `marker` prop, so we can customise the icon and listen for eventHandlers in `SearchAndShowAddress` like this:

```js
<SearchAndShowAddress
  markerOptions={{
    icon: ...,
    draggable: true,
    eventHandlers={{
      dragend: (e) => {
        onChange(...)
      }
    }}
  }}
  apiUrl="https://bplan-prod.liqd.net/api/addresses/"
/>
```

Based on the interaction with the `SearchAndShowAddress`, we can remove the other marker. 

Let me know what you think!

meinberlin/react/kiezradar/Kiezradars.jsx
sevfurneaux: Ah cool: yes I'll add a comment in this one, as we're under time pressure to get this working. But shall I create a new ticket about adding `title` to the a4 `Alert`?

meinberlin/apps/kiezradar/serializers.py
sevfurneaux: Ah, @hom3mad3 just mentioned that @goapunk is away this week. @vellip would you mind checking this? Thanks!

meinberlin/react/kiezradar/Kiezradar.jsx
sevfurneaux: Same for this one, @vellip as @goapunk is away. I _think_ this is the correct saved format?

meinberlin/react/kiezradar/Kiezradars.jsx
hom3mad3: yes, please

meinberlin/react/kiezradar/Kiezradar.jsx
vellip: It should be right, but the coordinates are reversed normally, iirc.

However, please use https://leafletjs.com/reference.html#marker-togeojson to generate the geojson from a marker.

meinberlin/react/kiezradar/Kiezradars.jsx
```diff
@@ -46,26 +87,66 @@ export default function Kiezradars (props) {
...
+
+  return (
+    <A4Alert
+      type="success"
+      message={<div dangerouslySetInnerHTML={{ __html: htmlMessage }} />}
```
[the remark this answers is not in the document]
vellip: CleanShot 2025-02-04 at 16 01 37

meinberlin/react/kiezradar/KiezradarMap.jsx
```diff
@@ -1,5 +1,145 @@
...
+      <Map
+        id="project-map"
+        zoomControl={false}
+        maxZoom={18}
+        attribution={attribution}
+        center={center}
+        bounds={bounds}
+        baseUrl={baseUrl}
+        mapboxToken={mapboxToken}
+        omtToken={omtToken}
```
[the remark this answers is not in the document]
vellip: If you need an example I think the normal Map component does that too

meinberlin/react/kiezradar/KiezradarMap.jsx
vellip: What I was envisioning for this story when i made the changes was more like:

```jsx
<div>
  <AddressSearch
    apiUrl={apiUrl}
    onSelectAddress={(feature) => {
      // update point for radius
    }}
  />
  <Map>
    <Radius point={updatedPoint}>
  </Map>
</div>
```

Have a look at the code of `SearchAndShowAddress`. Your case should be a reduced version of that. That then also allows the control to be **above** the map as it was planned originally.

meinberlin/react/kiezradar/KiezradarMap.jsx
```diff
@@ -1,5 +1,145 @@
...
+
+function Radius ({
+  point: _point,
+  radius: _radius,
+  minRadius,
+  maxRadius,
+  onChange
+}) {
+  const [point, setPoint] = useState(_point)
+  const [radius, setRadius] = useState(_radius)
```
vellip, line 63: You don't need the state here, do you? Let `KiezRadarMap` keep track of the point and radius (as you already do I think?) and then just pass that as props, so that it will update the radius component. Then you can get rid of the `setRadius` below and let the parent component update its state when calling `onChange` of the `Radius` Component.

```jsx
function Parent() {
  const [radius, setRadius] = useState(0)
  const [point, setPoint] = useState({})

  return (
    <div>
      <KiezRadarMap
        point={point}
        radius={radius}
        onChange={(newRadius) => {
          setRadius(newRadius.radius)
          setPoint(newRadius.point)
        }}
      />
    </div>
  )
}
```

vellip: Or add a simple context to prevent deeply nested props

meinberlin/react/kiezradar/KiezradarMap.jsx
```diff
@@ -1,5 +1,145 @@
...
+  return (
+    <>
+      <ControlWrapper position="topright">
+        <div className="kiezradar__controls--tablet">{radiusControls}</div>
+        {createPortal(
+          <div className="kiezradar__controls--mobile">{radiusControls}</div>,
+          document.getElementById('mobile-controls')
+        )}
+      </ControlWrapper>
+      <GeoJsonMarker
```
vellip, line 86: You can also use a normal marker, no need to generate the whole geojson structure then. https://react-leaflet.js.org/docs/example-draggable-marker/

meinberlin/react/kiezradar/KiezradarMap.jsx
```diff
@@ -1,5 +1,145 @@
...
+  }
+
+  return (
+    <div className="kiezradar__radius-controls">
+      <div className="kiezradar__radius-info">
+        <label className="kiezradar__radius-label" htmlFor="radius">
+          {chooseYourRadiusText}
+        </label>
+        <div className="kiezradar__radius-measurement">
+          {(radius / 1000).toFixed(1)} km
```
vellip, line 131: This splits with a decimal point, in Germany we use a comma ',' to indicate decimals.

meinberlin/react/kiezradar/Kiezradars.jsx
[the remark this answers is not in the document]
sevfurneaux: Alright, for now shall we not render the title, then implement the title once it is added on `A4`?

meinberlin/apps/kiezradar/serializers.py
vellip: That should be fine. Did it not work with the `binary` included?

meinberlin/react/kiezradar/KiezradarMap.jsx
sevfurneaux: Aha, got it. Yeah, that could work. Thanks!

meinberlin/react/kiezradar/Kiezradars.jsx
sevfurneaux: Added a subticket for this: [#8877 A4: Add title prop to Alert (and then add title in a4-meinberlin)](https://taiga.liqd.net/project/liqd-liquid-software/task/8877) 👍.

I'll hold out adding the titles as [mentioned here](https://github.com/liqd/a4-meinberlin/pull/6031#discussion_r1941478371), and will only add them once they are in A4.

meinberlin/react/kiezradar/Kiezradars.jsx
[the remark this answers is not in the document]
sevfurneaux: Added a ticket, [#8877 A4: Add title prop to Alert (and then add titles to kiezradar alerys in a4-meinberlin)](https://taiga.liqd.net/project/liqd-liquid-software/task/8877), so will apply after this PR is merged.

meinberlin/react/kiezradar/KiezradarMap.jsx
sevfurneaux: Changed to `toLocaleString`:

```js
<div className="kiezradar__radius-measurement">
  {(radius / 1000).toLocaleString('de-DE', {
    minimumFractionDigits: 1,
    maximumFractionDigits: 1
  })}
  km
</div>
```

meinberlin/react/kiezradar/KiezradarMap.jsx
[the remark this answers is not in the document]
sevfurneaux: Changed to:

```js
<Map
  id="project-map"
  zoomControl={false}
  maxZoom={18}
  {...mapProps}
  style={{ minHeight: '100%', height: '100%' }}
>
```

meinberlin/react/kiezradar/Kiezradar.jsx
sevfurneaux: Nice, I've now added that:

```js
onChange({ point: marker.toGeoJSON(), radius })
```

meinberlin/react/kiezradar/KiezradarMap.jsx
sevfurneaux: Same as mentioned [here](https://github.com/liqd/a4-meinberlin/pull/6031#discussion_r1941474358), adding state in `KiezRadarMap` causes map to constantly flicker.

Unsure how to resolve that, without using a ref?

meinberlin/react/kiezradar/KiezradarMap.jsx
sevfurneaux: This is now implemented:

```jsx
function Search ({ onChange }) {
  const map = useMap()

  const handleAddressChange = (point) => {
    const [lat, lng] = [...point.geometry.coordinates].reverse()
    map.flyTo({ lat, lng }, 13)
    onChange(point)
  }

  return (
    <ControlWrapper position="topleft" className="kiezradar__map-search">
      <AddressSearch
        apiUrl="https://bplan-prod.liqd.net/api/addresses/"
        onSelectAddress={handleAddressChange}
      />
    </ControlWrapper>
  )
}
```
Thanks for the advice!

meinberlin/react/kiezradar/KiezradarMap.jsx
sevfurneaux: Thanks, the state is now moved into `KiezradarMap`: much cleaner 🧼

meinberlin/react/kiezradar/KiezradarMap.jsx
vellip: I thought the AddressSearch was supposed to be above the map?

meinberlin/react/kiezradar/KiezradarMap.jsx
vellip: What I have done for i.e. the project overview map was to separate the everything that needs state into its own component that then handles all that outside of the component where the map is rendered. Maybe you can have a look at that map.

meinberlin/react/kiezradar/Kiezradar.jsx
```diff
@@ -15,21 +20,38 @@ const errorLimitedExceededText = django.gettext(
...
+export default function Kiezradar ({ kiezradar, onKiezradarSave, ...props }) {
   const navigate = useNavigate()
   const [loading, setLoading] = useState(false)
   const [error, setError] = useState(null)
   const [limitExceeded, setLimitExceeded] = useState(false)
+  const locationRef = useRef({
+    point: kiezradar?.point?.geometry.coordinates ?? CENTRAL_BERLIN,
+    radius: kiezradar?.radius ?? MIN_RADIUS
+  })
+
```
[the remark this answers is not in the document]
sevfurneaux: After more investigation, it looks like the flickering issue is in the `adhocracy4` `BaseMap` component, as using the "vanilla" Leaflet `MapContainer` component does not cause this issue when the state is in the parent:

### BaseMap component



### Leaflet MapContainer component



For now, I've changed the map to use the Leaflet `MapContainer`:

```jsx
<MapContainer ...>
  <MaplibreGlLayer ...  />
  <Radius ... />
  ...
</MapContainer>
```

...and have removed `useRef` and am using normal React `useState` 👍

With the flickering issuer with `BaseMap`: we should add a ticket for this and try and resolve it when we have time.

vellip (approved):
Thanks for your changes, feel free to merge when you are done.

meinberlin/react/kiezradar/Kiezradar.jsx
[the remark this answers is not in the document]
vellip: Hmm, interesting. Please create an issue then, yes

meinberlin/react/kiezradar/KiezradarMap.jsx
sevfurneaux: In the design, the address bar is in the same place as on the other maps.

meinberlin/apps/kiezradar/serializers.py
sevfurneaux: Reapplying `binary=True` causes the backend error when creating a kiezradar:

`{"point":["Wert muss gültiges JSON sein."]}`

Unsure why, as `updateItem` is JSON stringifying the payload:

```json
{"name":"sadasd","point":{"type":"Feature","properties":{},"geometry":{"type":"Point","coordinates":[13.405,52.52]}},"radius":1275}
```

sevfurneaux: Ah, solved it. If we keep `binary=True`, then we need to also `JSON.stringify` the `point` payload:

```js
const payload = {
  name: e.target.elements.name.value,
  point: JSON.stringify(location.point),
  radius: location.radius
}
```

This results in:

```json
{"name":"sadasd","point":"{\"type\":\"Feature\",\"properties\":{},\"geometry\":{\"type\":\"Point\",\"coordinates\":[13.405,52.52]}}","radius":1300}
```

vellip: In that case leave it out. We can still add it, if @goapunk later says it makes sense.

sevfurneaux: Ok, I've reverted it. But with `binary=True`, we have two breaking Django tests so the build is failing 🙈:

I've fixed them with the changes:

`assert kiezradar.point == geojson_point` 👉 `assert json.loads(kiezradar.point) == geojson_point`

`assert kiezradar.point == {}` 👉  `assert json.loads(kiezradar.point) == {}`

@goapunk here's the changed test file for reference: https://github.com/liqd/a4-meinberlin/pull/6031/files#diff-00309a7d3faedaaba62c06ba74cb545c8bce60f942ca30d881700ece8f5e3cfc

meinberlin/react/kiezradar/KiezradarMap.jsx
sevfurneaux: Nice, I've moved the state handling into the parent component, `Kiezradar`:

```js
const [point, setPoint] = useState(kiezradar?.point ?? defaultPoint)
const [radius, setRadius] = useState(kiezradar?.radius ?? MIN_RADIUS)
```

sevfurneaux:
Merged! ✅

Thanks for all the feedback @vellip.

PR #6036, opened 2025-02-04
[#8767] notifications: add fine-grained notification settings per user

vellip:
**Describe your changes**
This change adds the backend for more fine-grained notification settings. They are separated into:

- `notify` fields, which are meant for email notifications, and
- `track` fields, which are meant for activity-feed type (in-app) notifications

To not overflow the `User` model with a multitude of boolean fields a `OneToOneField` was used, pointing to `User`. It will be generated via signal, and if it does not exist still, on signup.

This PR also updates the frontend to allow for saving these values into the database and a cleaner frontend code.

**Tasks**
- [x] Changelog


meinberlin/react/account/useNotifications.js
```diff
@@ -0,0 +1,40 @@
...
+    initializeNotificationState(initialNotifications, showRestricted)
+  )
+
+  const getMasterToggles = useCallback(() => {
+    return notificationData
+      .filter((n) => !n.restricted || (n.restricted && showRestricted))
+      .map((n) => Object.keys(n.notifications).some((key) => notificationsState[key]))
+  }, [notificationsState, showRestricted])
+
+  return [
```
hom3mad3, line 33: wonder if returning an object here would make life easier

hom3mad3 (approved):

hom3mad3:
@vellip since this is quite django heavy, i'll leave it for someone else to review and merge @goapunk @m4ra @partizipation

meinberlin/apps/notifications/migrations/0002_add_notification_settings_per_user.py
```diff
@@ -0,0 +1,35 @@
...
+            user=user,
+            email_newsletter=user.get_newsletters,
+            notify_followers_phase_started=user.get_notifications,
+            notify_followers_phase_over_soon=user.get_notifications,
+            notify_followers_event_upcoming=user.get_notifications,
+            notify_creator=user.get_notifications,
+            notify_creator_on_moderator_feedback=user.get_notifications,
+            notify_initiators_project_created=user.get_notifications,
+            notify_moderator=user.get_notifications,
+        )
```
m4ra, line 23: might worth to put this for loop in a try / except block with a warning for when it runs in prod environment

meinberlin/apps/notifications/models.py
```diff
@@ -0,0 +1,57 @@
...
+    notify_initiators_project_created = models.BooleanField(default=True)
+    notify_moderator = models.BooleanField(default=True)
+
+    """
+    Tracked fields are being used to check if a notification should show in
+    the activity feed.
+    """
+    track_followers_phase_started = models.BooleanField(default=True)
+    track_followers_phase_over_soon = models.BooleanField(default=True)
+    track_followers_event_upcoming = models.BooleanField(default=True)
```
m4ra, line 40: by default they are all enabled?

meinberlin/apps/users/signals.py
```diff
@@ -1,10 +1,20 @@
 from allauth.account.signals import email_confirmed
+from django.db.models.signals import post_save
 from django.dispatch import receiver
 
+from ..notifications.models import NotificationSettings
```
m4ra, line 5: same, use absolute path

tests/users/test_signup.py
```diff
@@ -22,7 +43,8 @@ def test_signup_user_newsletter_checked(client):
     )
     assert resp.status_code == 302
     user = User.objects.get()
-    assert user.get_newsletters
+    assert user.notification_settings.email_newsletter
+    assert not user.notification_settings.notify_creator
```
m4ra, line 47: why here the notification is false for notify_creator and in the previous test is true? I don't see a change in the payload

meinberlin/apps/notifications/models.py
vellip: Yes, these are no email notifications, they will just determine whether these will be shown in the activity feed or not so I thought it would be fine to start with them enabled. There's also a comment like that in taiga.

tests/users/test_signup.py
vellip: `"get_notifications": "on",` is not set in the first test.

m4ra (approved):
thanks,
@vellip needs a rebase, as I merged your other PRs

PR #6052, opened 2025-02-06
[#8874][#8873] style into grid and sort module tiles

vellip:
**Describe your changes**
Currently based on @hom3mad3 s PR, probably better to review once that is merged.

put module tiles in grid and sort them according to specification, with help of a new template tag that checks 
CleanShot 2025-02-06 at 12 12 13

**Tasks**
- [x] Changelog


vellip:
rebased now, so ready for review

meinberlin/templates/a4modules/module_detail.html
```diff
@@ -77,19 +77,16 @@ <h1>
...
+        <section>
+          <h2>{% translate 'More from this online participation:' %}</h2>
+          <ul class="list--clean flexgrid grid--3">
+            {% get_sorted_modules as modules %}
+            {% for module in modules %}
+              <li>
+                {% include "meinberlin_projects/includes/project_module_tile.html" with project=project module=module %}
+              </li>
+            {% endfor %}
+          </ul>
```
m4ra, line 89: I think this introduced some errors in the other modules templates (brainstorming, idea). Check the test failures.

m4ra (changes requested):
Need to address the tests failing, and a rebase to dev.

meinberlin/templates/a4modules/module_detail.html
vellip: @m4ra i addressed this here https://mattermost.liqd.net/liqd/pl/zbz5ehuk7f84mrbwxcddyyjhyh

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
```diff
@@ -33,6 +35,24 @@ def is_external(project):
...
 
 
+@register.simple_tag(takes_context=True)
+def get_sorted_modules(context):
+    project = context["project"]
+    module = context.get("module", None)
+    module_qs = project.modules
+
+    if module:
+        module_qs = module.other_modules
```
m4ra, line 45: @vellip I don't understand what's happening here, when there is module in the context, we don't care about the project's modules?

vellip: @m4ra When there is a  module in context we only care about the other modules and
not the current module, because the user is already on that module detail
page in that case.

m4ra (approved):
thanks!

PR #6059, opened 2025-02-10
[#8895] Refactor search profile, update alerts and add delete modal

sevfurneaux:
**Describe your changes**
This PR refactors the search profile, to have a similar pattern of updating the alerts as the `Kiezradars` component where most of the state is in the parent.

It also updates the search profile alert text and adds the delete modal when deleting search profiles.

**Tasks**
- [X] Changelog


meinberlin/react/kiezradar/DeleteModal.jsx
```diff
@@ -0,0 +1,46 @@
...
+    const handleKeyDown = (event) => {
+      if (event.key === 'Escape') closeModal()
+    }
+
+    document.addEventListener('keydown', handleKeyDown)
+
+    return () => document.removeEventListener('keydown', handleKeyDown)
+  }, [])
+
+  useEffect(() => {
```
hom3mad3, line 25: i'd add a null check before calling showModal()
```
useEffect(() => {
  if (dialogRef.current) {
    dialogRef.current.showModal()
  }
}, [])
```

meinberlin/react/kiezradar/DeleteModal.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+export default function DeleteModal ({ title, message, onDelete, onClose }) {
+  const dialogRef = useRef(null)
+
+  const closeModal = () => {
+    dialogRef.current.close()
+    onClose()
+  }
+
+  useEffect(() => {
```
hom3mad3, line 15: we could probably save some performance and drop this `useEffect`, and add onKeyDown as dialog attribute/prop. see comment below

meinberlin/react/kiezradar/DeleteModal.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+    return () => document.removeEventListener('keydown', handleKeyDown)
+  }, [])
+
+  useEffect(() => {
+    dialogRef.current.showModal()
+  }, [])
+
+  return (
+    <dialog className="kiezradar__modal" ref={dialogRef} aria-modal="true">
```
hom3mad3, line 30:
```
    <dialog 
      className="kiezradar__modal" 
      ref={dialogRef} 
      aria-modal="true" //can't remember whether we still need this
      aria-labelledby="modal-title"
      onKeyDown={(event) => {
        if (event.key === 'Escape') closeModal()
      }}
    >
```

hom3mad3: also, since we have a title we might add a `aria-labelledby="modal-title"` and the `id="modal-title"` to the h3

meinberlin/react/kiezradar/SearchProfile.jsx
```diff
@@ -22,25 +24,32 @@ const statusNames = {
...
 
-export default function SearchProfile ({ apiUrl, planListUrl, profile: profile_, onDelete }) {
+export default function SearchProfile ({ apiUrl, planListUrl, searchProfile, onSave, onDelete }) {
   const [isEditing, setIsEditing] = useState(false)
   const [loading, setLoading] = useState(false)
   const [error, setError] = useState(null)
-  const [profile, setProfile] = useState(profile_)
+  const [deleteModal, setDeleteModal] = useState(null)
 
   const handleDelete = async () => {
```
hom3mad3, line 41: what about `const handleDelete = useCallback(async () => {` ... and handleSubmit? to prevent unnecessary re-renders

meinberlin/react/kiezradar/SearchProfile.jsx
```diff
@@ -71,100 +80,110 @@ export default function SearchProfile ({ apiUrl, planListUrl, profile: profile_,
   }
 
   const filters = [
-    profile.districts,
-    profile.topics,
-    profile.project_types,
-    profile.status.map((status) => ({ name: statusNames[status.name] })),
-    profile.organisations
+    searchProfile.districts,
```
hom3mad3, line 83: what about using destruct ? `const { id, name, query_text, districts, topics, project_types, status, organisations, plans_only, notification } = searchProfile;`

meinberlin/react/kiezradar/SearchProfile.jsx
```diff
@@ -71,100 +80,110 @@ export default function SearchProfile ({ apiUrl, planListUrl, profile: profile_,
...
-                type="submit"
-                disabled={loading}
-              >
-                {loading ? savingText + '...' : saveText}
-              </button>
+          )}
+        </div>
+        {error && <Alert type="danger" title={errorText} message={error} />}
+        {isEditing && (
+          <form className="form--base panel--heavy search-profile__form" onSubmit={handleSubmit}>
```
hom3mad3, line 131: what do you think of adding `<fieldset disabled={loading}>` below `<form>` to prevent submitting multiple times?

meinberlin/react/kiezradar/SearchProfile.jsx
```diff
@@ -71,100 +80,110 @@ export default function SearchProfile ({ apiUrl, planListUrl, profile: profile_,
...
+              <label htmlFor="name">{renameSearchProfileText}</label>
+              <input
+                id="name"
+                name="name"
+                type="text"
+                defaultValue={searchProfile.name}
+                onKeyDown={(e) => {
+                  if (e.key === 'Enter') {
+                    e.preventDefault()
+                    e.target.form.requestSubmit()
```
hom3mad3, line 142: we probably should avoid directly handling the DOM here 🤔 💭

hom3mad3: sorry a string of meetings interrupted my review 🫠 i think i was going to suggest useRef here..

sevfurneaux: So it turns out by adding `type="button` to `<button type="button" className="link" onClick={() => setIsEditing(false)}>` means we no longer need `onKeyDown` as the browser now natively handles it 🎉

Unsure why doing this fixed it, as it is not the submit button, but prior to this, if we didn't have `onKeyDown`, Chrome was complaining:

`"Form submission canceled because the form is not connected"`

meinberlin/react/kiezradar/DeleteModal.jsx
sevfurneaux: Thanks! Have added 👍

meinberlin/react/kiezradar/DeleteModal.jsx
sevfurneaux: Nice, that's now removed.

meinberlin/react/kiezradar/DeleteModal.jsx
sevfurneaux: Added that 👍

meinberlin/react/kiezradar/SearchProfile.jsx
sevfurneaux: Tried this, but the linter complains:

```js
  const {
    query_text,
    districts,
    topics,
    project_types,
    status,
    organisations,
    plans_only
  } = searchProfile
```

```
  87:5  error  Identifier 'query_text' is not in camel case     camelcase
  90:5  error  Identifier 'project_types' is not in camel case  camelcase
  93:5  error  Identifier 'plans_only' is not in camel case     camelcase
  99:5  error  Identifier 'project_types' is not in camel case  camelcase
 106:6  error  Identifier 'query_text' is not in camel case     camelcase
 108:6  error  Identifier 'plans_only' is not in camel case     camelcase
```

Let's keep as is for now (and not alias to camelCase, as that is misleading what named properties are actually returned from a search profile).

Side note: I am always battling with this linter before pushing 🙈, mainly around formatting. Maybe in the future, you can change to Prettier formatting, which formats on save in most editors.

hom3mad3 (approved):
Thanks ✨

PR #6096, opened 2025-02-14
[#8881] project-info-page: add timeline html/ data

vellip:
**Describe your changes**
Just backend implementation, no styling
CleanShot 2025-02-14 at 16 34 22

**Tasks**
- [x] Changelog


meinberlin/apps/projects/templates/meinberlin_projects/includes/timeline.html
hom3mad3, line 1: what about completed/ended projects? do we display some information then or the timeline disappears? might be good to have the timeline still in.

vellip: It is already still visible if the project has more than one module/ an event, just that everything will be marked as active/past, which would be the correct behavior.

hom3mad3 (approved):

Issue #6071, opened 2025-02-12
#8766 [mB] Redesign frontend & backend - More information page (project detail page) – [shown content in Infobox doesn't match the edit]


Badri39:
**URL:** 
https://meinberlin-design-dev.liqd.net/projekte/hallo-ich-bin-ein-projekt/information/

**user:** 
User with account 

**expected behaviour:** 
- The text written in a project's information box can be resized. 
- Images can also be adjusted in size. 
- Videos (e.g., from YouTube) can be displayed.

**behaviour:** 
- Changed text size is not displayed currently under project information (the entire text appears in the standard size).
- Images are distorted and size adjustments aren’t shown. 
- Embedded video frame is cut off, probably because the “play” video” screen is not displayed but the video tile still shrinks to that size. 

**Browsers and Operating Systems**:
All devices tested

**screenshots:**

<img width="934" alt="Image" src="https://github.com/user-attachments/assets/6cb0fc3f-c28d-4db9-abf0-73a0c124ce6c" /> 

<img width="754" alt="Image" src="https://github.com/user-attachments/assets/a88465f1-b924-4847-a329-1fb6eae53eea" />

<img width="431" alt="Image" src="https://github.com/user-attachments/assets/6c68b14c-f9d8-4f73-8dec-95a8ce42305e" />

<img width="470" alt="Image" src="https://github.com/user-attachments/assets/d63700a1-2856-4715-87e8-195b99817859" />

sblindow:
Just to clarify: We could only reproduce these issues in the ck-editor that's on the info page. That's why we thought that this is a story specific issue.

Issue #6072, opened 2025-02-12
#8766 [mB] Redesign frontend & backend - More information page (project detail page) – address and website of contact person info are not displayed


sblindow:
### URL
```https://meinberlin-design-dev.liqd.net/projekte/[project-name]/information/```

### Users
Any 

### Expected Behavior
Contact person (Kontakt für Rückfragen) may contain following details: name, address, telephone number, email (opens mail programm), website (opens new tab) (generated via dashboard through initiators)

### Behavior
The address and the website of the contact information do not appear at the bottom of the page with the rest of the info.

### Browsers and Operating Systems
All Systems and Browsers tested

PR #6100, opened 2025-02-18
fix info-page issues

hom3mad3:
fixes https://github.com/liqd/a4-meinberlin/issues/6071 and https://github.com/liqd/a4-meinberlin/issues/6072

### DIFF
--- a/meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
+++ b/meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
@@ -1,5 +1,5 @@
 {% extends 'base.html' %}
-{% load i18n %}
+{% load i18n ckeditor_tags %}
 
 {% block title %}
   {{ project.name }} -{% trans 'Information' %}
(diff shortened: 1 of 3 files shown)


meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'base.html' %}
-{% load i18n %}
+{% load i18n ckeditor_tags %}
```
vellip, line 2: Is this necessary? As far as I know, `safe` is built into django?

vellip (approved):
LGTM just one question

meinberlin/apps/projects/templates/meinberlin_projects/project_information.html
hom3mad3: it's needed so the collapsible element (accordion) renders correctly.
as a matter of fact i forgot one extra tag (disable_iframes)

@goapunk's explanation:

> yes it makes the accordions (called collapsibles in our code for some reason) render correctly

> so it's only needed in places the ckeditor allows adding accordions

> and there's a tag for places where the ckeditor allows embeding videos called disable_iframes which prevents the videos from auto-loading and leaking data to youtube before agreeing to the video opt in

hom3mad3: i forgot to explicitly call the collapsible on the template too 🙈

vellip: that explains my confusion :D

PR #6114, opened 2025-02-19
fix geo migrations

goapunk:
**Describe your changes**
data and schema migrations can't be mixed, so separate them out in individual migrations

**Tasks**
- [ ] Changelog


goapunk:
@m4ra same here as in a4

PR #6123, opened 2025-02-20
[#8875] insights: add a+ insights logic and display on project detail

vellip:
also adds project result page, insight logic mostly taken from a+, with some changes. Not moved into a4 for a quicker implementation

**Tasks**
- [x] Changelog


meinberlin/apps/projects/migrations/0007_create_insights.py
```diff
@@ -0,0 +1,111 @@
...
+        model_name = rating.content_type.model
+        content_model = apps.get_model(
+            app_label=rating.content_type.app_label,
+            model_name=model_name,
+        )
+        content_object = content_model.objects.get(id=rating.object_pk)
+
+        if model_name == "comment":
+            project = content_object.project
+        elif model_name in ["idea", "mapidea", "topic", "proposal"]:
```
goapunk, line 72: we will also need `maptopic` here

goapunk (changes requested):
Thanks! I left a comment regarding the ratings

goapunk:
@m4ra one thing we need to take into account that the insights data migration will take quite a long time, locally I got about 7min, from experience that might take up to 1h on prod

goapunk:
@vellip needs cherry-picking because of the force-pushed migrations to dev

meinberlin/apps/projects/insights.py
```diff
@@ -0,0 +1,111 @@
...
+    comments = Comment.objects.filter(
+        Q(project=project) | Q(parent_comment__project=project)
+    )
+    proposals = Proposal.objects.filter(module__in=modules)
+    polls = Poll.objects.filter(module__in=modules)
+    votes = Vote.objects.filter(choice__question__poll__in=polls)
+    answers = Answer.objects.filter(question__poll__in=polls)
+    live_questions = LiveQuestion.objects.filter(module__in=modules)
+    likes = Like.objects.filter(question__in=live_questions)
+    topics = Topic.objects.filter(module__in=modules)
```
goapunk, line 40: We'll also need MapTopics here I think?

goapunk:
@m4ra @vellip I wonder if we should just leave the 007 migration out. This way deploying will be much faster and we can just run the management command to populate the insights without requiring a downtime. What do you think?

vellip:
@goapunk hmm, might make your lives easier when it comes to deploying, yeaah.

meinberlin/apps/projects/migrations/0007_create_insights.py
```diff
@@ -0,0 +1,115 @@
...
+    Proposal = apps.get_model("meinberlin_budgeting", "Proposal")
+    Idea = apps.get_model("meinberlin_ideas", "Idea")
+    Like = apps.get_model("meinberlin_likes", "Like")
+    LiveQuestion = apps.get_model("meinberlin_livequestions", "LiveQuestion")
+    MapIdea = apps.get_model("meinberlin_mapideas", "MapIdea")
+
+    insights = defaultdict(ProjectInsight)
+    user_ids = defaultdict(set)
+
+    logger.info("creating insights")
```
goapunk, line 24: I'd actually prefer not to have that logging, it doesn't really help, django already tells us what migration is currently running :)

goapunk:
> @goapunk hmm, might make your lives easier when it comes to deploying, yeaah. Should I take it out?

Let's take it out, yeah. If we end up needing it we can still add it later

meinberlin/apps/projects/insights.py
```diff
@@ -0,0 +1,118 @@
...
+    map_topics = MapTopic.objects.filter(module__in=modules)
+
+    values = [Rating.POSITIVE, Rating.NEGATIVE]
+    ratings_ideas = Rating.objects.filter(idea__in=ideas, value__in=values)
+    ratings_map_ideas = Rating.objects.filter(mapidea__in=map_ideas, value__in=values)
+    ratings_comments = Rating.objects.filter(comment__in=comments, value__in=values)
+    ratings_topics = Rating.objects.filter(topic__in=topics, value__in=values)
+    ratings_map_topics = Rating.objects.filter(
+        maptopic__in=map_topics, value__in=values
+    )
```
m4ra, line 51: what about the plans? @goapunk @vellip

meinberlin/apps/projects/models.py
```diff
@@ -71,3 +72,70 @@ def get_absolute_url(self):
...
+    live_questions = models.PositiveIntegerField(default=0)
+
+    @staticmethod
+    def update_context(project, context):
+        insight, created = ProjectInsight.objects.get_or_create(project=project)
+        context.update(create_insight_context(insight=insight))
+        return context
+
+    def __str__(self):
+        return "Insights for project %s" % self.project.name
```
m4ra, line 96: few things here ( I know you imported them from aplus, but always room for improvement):
Change the staticmethod to classmethod, then we can pass `cls` in the arguments of the functions, instead of calling `ProjectInsight` which is odd since we are inside the class already.
`

Use f-strings in `__str__()`, more readable

We may want to add the `Meta` attribute to define the ordering of the objects `(-created)`

meinberlin/apps/projects/insights.py
vellip: I think, since plans are hierarchy-wise above projects, we don't need to track them, right?

meinberlin/apps/projects/models.py
```diff
@@ -71,3 +72,70 @@ def get_absolute_url(self):
...
+    ("BS", _("brainstorming")),
+    ("MBS", _("spatial brainstorming")),
+    ("IC", _("idea challenge")),
+    ("MIC", _("spatial idea challenge")),
+    ("TR", _("text review")),
+    ("PO", _("poll")),
+    ("PB", _("participatory budgeting")),
+    ("IE", _("interactive event")),
+    ("TP", _("prioritization")),
+    ("DB", _("debate")),
```
m4ra, line 110: same here, may need to add the missing modules (MTP, PB2 and PB3)

meinberlin/apps/projects/signals.py
```diff
@@ -40,3 +52,83 @@ def post_save_delete(sender, instance, *args, **kwargs):
...
+
+
+@receiver(poll_voted)
+def increase_poll_participant_count(sender, poll, creator, content_id, **kwargs):
+    insight, _ = ProjectInsight.objects.get_or_create(project=poll.module.project)
+    if creator:
+        insight.active_participants.add(creator.id)
+    else:
+        insight.unregistered_participants += 1
+    insight.save()
```
m4ra, line 134: same for Map Topic

meinberlin/apps/projects/views.py
```diff
@@ -360,6 +360,7 @@ def project(self):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["events"] = self.project.offlineevent_set.all().order_by("date")
+        return models.ProjectInsight.update_context(self.project, context)
```
m4ra, line 363: you mean `self.model here`?
when using a `classmethod` in the models.py, this will become `self.model.update_context(context)`

m4ra (commented):
Great! just a few things

meinberlin/apps/projects/models.py
```diff
@@ -71,3 +72,79 @@ def get_absolute_url(self):
...
+
+    active_modules = [
+        module for module in insight.project.modules if not module.is_draft
+    ]
+    blueprint_types = {module.blueprint_type for module in active_modules}
+    show_polls = "PO" in blueprint_types
+    show_live_questions = "IE" in blueprint_types
+    show_ideas = bool(
+        blueprint_types.intersection(
+            {"BS", "IC", "MBS", "TP", "MTP", "MIC", "PB", "PB2", "PB3"}
```
vellip, line 126: I also added Topic Prio and Map Topic Prio here. I hope that is right?

m4ra (approved):
LGTM, thanks!

Issue #5862, opened 2024-12-06
re-design: category select for proposals broken


CarolingerSeilchenspringer:
**URL:** https://meinberlin-design-dev.liqd.net/budgeting/2024-01148/update/
**user:** registered
**expected behaviour:** I can select a category for my proposal
**behaviour:** select is broken and/or selected category displayed below dropdown
**important screensize:**
**device & browser:** Sonoma Firefox (on Chrome the select works but the selected category is still listet below dropdown, clicking on it makes unstyled element (?) appea
**Comment/Question:** 

Screenshot?
chrome
<img width="846" alt="Bildschirmfoto 2024-12-06 um 10 39 42" src="https://github.com/user-attachments/assets/aee23072-3738-4a60-9961-9da769c43987">

PR #6126, opened 2025-02-22
fix category select issue

hom3mad3:
fixes https://github.com/liqd/a4-meinberlin/issues/5862


hom3mad3:
@goapunk @vellip could i delete select2widget altogether? is it being used in any maps or has it been completely replaced? i could not find the select field from 'meinberlin/apps/maps/templates/meinberlin_maps/map_choose_polygon_with_preset_widget.html' in use but wanted to make sure.

goapunk:
It's still used in the dashboard afaik: https://github.com/liqd/a4-meinberlin/blob/dev/meinberlin/apps/maps/dashboard.py ? @vellip ?

vellip:
I don't know if it's still working, like frontend-wise, but its definitely used in the dashboard. So better not completely delete, also shouldn't be replaced with a default select in those cases because it's a UX enhancement

hom3mad3:
@goapunk @vellip thanks, ill leave it as it is then. maybe create and issue to figure out what to do about it later.

vellip (approved):
LGTM!

Issue #6136, opened 2025-02-25
(mB) modul view is cut off// doesn't show correctly


Badri39:
**URL:** 
https://meinberlin-design-dev.liqd.net/dashboard/projects/hallo-ich-bin-ein-projekt/basic/
**user:** 
Admin 
**expected behaviour:** 
The modules on meinBerlin are fully displayed, with no text being cut off.
**behaviour:** 
The modules are cropped at the edges, and some text elements are truncated.
**important screensize:**
Laptop 
**device & browser:** 
Mac: Safari/Firefox 

<img width="826" alt="Image" src="https://github.com/user-attachments/assets/ac9afde7-1088-45ab-89dd-2bbfde346a21" />

Issue #6061, opened 2025-02-11
[redesign] Initiator Dashboard – Dashboard switches to mobile design when closing module selection modal without choosing a module


sblindow:
### URL
Project Dashboard: ```https://meinberlin-design-dev.liqd.net/dashboard/projects/[project-name]```

### User
Initiators, Group Members & Admins

### Expected Behavior
The user can close the module selection modal without selecting a module by clicking on the 'x' in the top right corner.

### Behavior
When the user closes the modal by clicking on the 'x' in the top right corner, the modal closes but the project dashboard breaks into the mobile design. 

### Screenshot
Dashboard after closing the modal on desktop: 

<img width="1048" alt="Image" src="https://github.com/user-attachments/assets/37025f48-c681-4437-9c4f-9b67e1c8052b" />

hom3mad3:
@sblindow i believe this is not a new issue, since it has nothing to do with the changes in the new modals (i'm not touching the dashboard view). now i'm unsure there might be any overlaps? @goapunk maybe you know?

goapunk:
> [@sblindow](https://github.com/sblindow) i believe this is not a new issue, since it has nothing to do with the changes in the new modals (i'm not touching the dashboard view). now i'm unsure there might be any overlaps? [@goapunk](https://github.com/goapunk) maybe you know?

not from the top of my head, only thing I can think of is that the modals have been reworked, would that also affect modals in the dashboard?

hom3mad3:
@sblindow thanks i'll look into it!!

PR #6143, opened 2025-02-25
[#6061] a4dashboard: fix blueprints modal

hom3mad3:
Fixes issue with body element being overwritten by Bootstrap modal classes https://github.com/liqd/a4-meinberlin/issues/6061 and https://github.com/liqd/a4-meinberlin/issues/6136
Also modernizing the modal by removing jQuery, Bootstrap and going for native HTML dialog and vanilla JS.

### DIFF
--- a/meinberlin/assets/js/blueprint-dialog.js
+++ b/meinberlin/assets/js/blueprint-dialog.js
@@ -1,0 +1,67 @@
+document.addEventListener('DOMContentLoaded', function () {
+  initModuleDialog()
+})
+
+function initModuleDialog () {
+  const button = document.getElementById('module-blueprint-btn')
+  const dialog = document.getElementById('module-blueprint-list')
+
+  if (!button || !dialog) return
+
+  // Move dialog to <body> for proper behavior
+  document.body.appendChild(dialog)
+
+  button.addEventListener('click', function (e) {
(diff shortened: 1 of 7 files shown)


meinberlin/assets/js/blueprint-dialog.js
```diff
@@ -0,0 +1,67 @@
...
+          } else {
+            console.warn('Could not extract content from full page response')
+          }
+        }
+
+        dialog.querySelector('.blueprint-dialog__body').innerHTML = html
+        dialog.showModal()
+      })
+      .catch(error => console.error('Error loading dialog content:', error))
+  })
```
vellip, line 52: might look a little cleaner using `await` maybe?

vellip (approved):
LGTM, thanks! 💜

meinberlin/assets/js/blueprint-dialog.js
hom3mad3: @vellip if you agree, i'll merge this now. I would add this as an improvement for later, just so I can move on with the other urgent fixes, does that sound like a plan?

vellip: Yes sure :)

PR #6161, opened 2025-02-26
config: dockerize repository

hom3mad3:
**Description**
A first experiment in using docker to run the project

Mostly based on the existing makefile

TODO
- [ ] how to ignore makefile if building with docker?
- [ ] issues with venv

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,83 @@
+version: '3.8'
+
+services:
+  db:
+    image: postgis/postgis:16-3.5
+    volumes:
+      - postgres_data:/var/lib/postgresql/data/
+    environment:
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_USER=postgres
+      - POSTGRES_DB=django
+    ports:
+      - "5432:5432"
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+  
+  redis:
+    image: redis:alpine
+    ports:
+      - "6379:6379"
+  
+  web:
+    build: .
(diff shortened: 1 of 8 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,83 @@
...
+    ports:
+      - "5432:5432"
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+  
+  redis:
+    image: redis:alpine
```
hom3mad3, line 21: normal install was too big, not sure if we have special needs or if alpine is good enough?

docker-compose.yml
```diff
@@ -0,0 +1,83 @@
...
+  db:
+    image: postgis/postgis:16-3.5
+    volumes:
+      - postgres_data:/var/lib/postgresql/data/
+    environment:
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_USER=postgres
+      - POSTGRES_DB=django
+    ports:
+      - "5432:5432"
```
vellip, line 13: we could use a network and then use the internal address to avoid opening a port on the machine. But maybe that's something you want, for i.e. sql browser software

partizipation:
Impacts PR for specifying python version in adhocracy plus (in a good way): [#2993](https://github.com/liqd/adhocracy-plus/pull/2933)

docker-compose.yml
partizipation: Yeah I'm curious if for our apps we would run into issues if we're dev-alpine prod-ubuntu. I guess if it builds then for now the answer is no? Would we need (or want) to dockerize in production at this point, guessing it's less work to just use this for dev for now?

PR #6178, opened 2025-03-03
search-profiles: fix typo in get_search_profiles_for_project

vellip:
**Describe your changes**

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/kiezradar/models.py
+++ b/meinberlin/apps/kiezradar/models.py
@@ -188,7 +188,7 @@ def get_search_profiles_for_project(project: Project) -> QuerySet[SearchProfile]
         & Q(disabled=False)
     )
     search_term = project.name
-    if connection.vendor == "postresql":
+    if connection.vendor == "postgresql":
         # django has some postgresql-only search tools which are much better
         search_profiles = search_profiles.filter(
             Q(query__text__search=search_term) | Q(query__isnull=True)
--- a/meinberlin/apps/kiezradar/models.py
+++ b/meinberlin/apps/kiezradar/models.py
@@ -3,6 +3,8 @@ from functools import reduce
 from django.conf import settings
 from django.contrib.gis.db import models as gis_models
 from django.contrib.gis.db.models.functions import Distance
+from django.contrib.postgres.search import SearchQuery
+from django.contrib.postgres.search import SearchVector
 from django.core.exceptions import ValidationError
 from django.core.validators import MaxValueValidator
 from django.core.validators import MinValueValidator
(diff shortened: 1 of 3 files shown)


meinberlin/apps/kiezradar/models.py
```diff
@@ -188,7 +188,7 @@ def get_search_profiles_for_project(project: Project) -> QuerySet[SearchProfile]
         & Q(disabled=False)
     )
     search_term = project.name
-    if connection.vendor == "postresql":
+    if connection.vendor == "postgresql":
```
goapunk, line 191: ouch :face_with_diagonal_mouth:

meinberlin/apps/kiezradar/models.py
```diff
@@ -203,5 +208,6 @@ def get_search_profiles_for_project(project: Project) -> QuerySet[SearchProfile]
     if project.point:
         search_profiles = search_profiles.filter(
             Q(kiezradars__radius__gte=Distance("kiezradars__point", project.point))
+            | Q(kiezradars__isnull=True)
```
goapunk, line 211: added a missing OR here

meinberlin/apps/kiezradar/models.py
```diff
@@ -188,10 +190,13 @@ def get_search_profiles_for_project(project: Project) -> QuerySet[SearchProfile]
         & Q(disabled=False)
     )
     search_term = project.name
-    if connection.vendor == "postresql":
+    if connection.vendor == "postgresql":
         # django has some postgresql-only search tools which are much better
+        query = "|".join(search_term.split(" "))
```
vellip, line 195: Isn't this the same as just passing `search_term` to `SearchQuery` and not passing `raw`? 

> If search_type is 'plain', which is the default, the terms are treated as separate keywords.
https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/search/#searchquery

vellip (commented):
LGTM, thanks for taking over! I can't approve because it's my own PR.

meinberlin/apps/kiezradar/models.py
goapunk: yes but then it's an & and not |, so all separate keywords have to match, not just one. This would probably make sense when searching from a searchprofile for a matching project, the problem here is that we are basically reverse searching by using the project.title (etc.) to match a search profile. Or am I confusing something? :sweat_smile:

vellip: Ahh, I think I understand. Yea, all good then!

goapunk:
@m4ra do you want to take a look here?

m4ra (dismissed):
@goapunk one correction for the filtering of organisations and districts. You can go ahead and merge once fixed.

PR #6204, opened 2025-03-04
wip [#8866] templates/projects: create module_tile insights

hom3mad3:
**Description**
module-stats

**Todo**
- [x] Fix test
- [ ] Add missing module types
- [ ] Cleanup template
- [ ] template tag for pluralization?

### DIFF
--- a/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
+++ b/meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
@@ -38,17 +38,9 @@
     {% include 'meinberlin_projects/project_actions.html' %}
 
     {% include 'meinberlin_projects/includes/project_details_summary.html' %}
-    <section aria-labelledby="project-participation" class="block">
-      <h2 id="project-participation">{% translate "Online Participation" %}</h2>
-      <ul class="list--clean flexgrid grid--3 grid--stretch">
-        {% get_sorted_modules as modules %}
-        {% for module in modules %}
-          <li>
-            {% include "meinberlin_projects/includes/project_module_tile.html" with project=project module=module %}
-          </li>
-        {% endfor %}
-      </ul>
-    </section>
+
+    {% include 'meinberlin_projects/includes/project_participation.html' %}
+
     {% if events %}
     <section aria-labelledby="project-events">
       <div class="actionable-list__header">
--- a/meinberlin/apps/projects/templates/meinberlin_projects/includes/module-tile/module_tile.html
+++ b/meinberlin/apps/projects/templates/meinberlin_projects/includes/module-tile/module_tile.html
@@ -2,8 +2,11 @@
 {% project_tile_image project as project_image %}
 {% project_tile_image_copyright project as project_image_copyright %}
 <a href="{% url 'module-detail' module.slug %}" class="module-tile">
-    <h3 class="module-tile__title">{{ module.name }}</h3>
-    <p class="module-tile__description">{{ module.description|truncatechars:240 }}</p>
+    <div class="module-tile__content">
+        <h3 class="module-tile__title">{{ module.name }}</h3>
+        <p class="module-tile__description">{{ module.description|truncatechars:240 }}</p>
+        {% include 'meinberlin_projects/includes/module-tile/module_tile_insights.html' %}
+    </div>
 
     {% if module.module_running_time_left %}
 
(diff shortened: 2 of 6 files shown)


hom3mad3:
@m4ra @goapunk @vellip i still need to fix the templates and the failing test, but maybe you feel like having a look at the django part

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
```diff
@@ -38,17 +38,9 @@
...
-        {% get_sorted_modules as modules %}
-        {% for module in modules %}
-          <li>
-            {% include "meinberlin_projects/includes/project_module_tile.html" with project=project module=module %}
-          </li>
-        {% endfor %}
-      </ul>
-    </section>
+
+    {% include 'meinberlin_projects/includes/project_participation.html' %}
```
vellip, line 42: Do you think we should either do this for all the sections or none of the sections?

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
```diff
@@ -55,16 +59,57 @@ def get_sorted_modules(context):
...
+        "budget_proposals": budget_proposal.objects.filter(module=module).count(),
+        "kiezkasse_proposals": kiezkasse_proposal.objects.filter(module=module).count(),
+        "votes": Vote.objects.filter(choice__question__poll__module=module)
+        .values("creator")
+        .distinct()
+        .count(),
+        "live_questions": LiveQuestion.objects.filter(module=module).count(),
+        "likes": Like.objects.filter(question__module=module).count(),
+    }
+
```
vellip, line 96: I might be misunderstanding, but this is for a single module, which means we will always only have one type of item, so either ideas, proposals, votes, topics, map ideas, map topics, etc – do we need to query everything else?

vellip: And, why not use the `ProjectInsight` model which already tracks most/all of that? You can do `project.insight` to access that.

vellip: So what I would suggest is checking `module.blueprint_type` and depending on that I would return `module.project.insight.written_ideas` for example for most of the blueprint types.

vellip (commented):
Thanks, @hom3mad3! I left some general questions/suggestions.

meinberlin/apps/projects/templates/meinberlin_projects/project_detail.html
hom3mad3: the template is starting to look quite messy, so i tend towards having templates for all the sections. my gauge is, if i need to scroll or make spaces in order to read a file, then snippets are called for.

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
hom3mad3: i'll have a look, thanks for the hint. i wasn't too sure how to get the specific amount for each type. i also noticed that what's in insights, doesn't always return the correct information for the individual module (i.e. vote, ratings) and needed to customize.

hom3mad3: the reason for the dictionary was to be able to name the specific module type and then assign to and icon, which would make the template easier to read and eventually debug

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
```diff
@@ -76,13 +121,9 @@ def get_num_entries(module):
...
-        .aggregate(comment_count=Count("pk") + Sum("child_comment_count"))[
-            "comment_count"
-        ]
-    )
-    if comment_count is None:
-        comment_count = 0
-    return item_count + comment_count
+
+    comment_count = count_comments(comment_filter)
+    if comment_count > 0:
```
hom3mad3, line 126: mmm might need to return zeros not to break other stuff and tests.. but for the tile insights, there should be a clean, DRY way to display only > 0 ...

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
goapunk: the project.insights won't work for ideas (or any other module content) because it's project-wide, so if there are two idea modules the count will be off. Comments will always be off if there's more than one module (as all allow comments in some way), right?

goapunk: however I agree with @vellip that we probably only want to query the relevant object of the module (e.g. idea or topic or proposal) not all of them

Issue #6104, opened 2025-02-18
#8827 [mB] Redesign frontend: Highlighted Example Project on Landing Page – Design


sblindow:
### 1. Tick Mark is too large 
The tick mark that appears when a featured project has concluded seems to be a little too large compared to the design.
##### Design

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/b0d825be-2653-476e-b00d-98d0c0ef380e" />

##### Current Implementation

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/e57056bc-4ddf-497c-8ec9-2531aeb62876" />

### 2. [Desktop] Text and button don't align with bottom of image when the project is concluded
When the featured project is concluded and has a tick mark, the section on the right with the (overwritten) title, teaser text and button do not align with the bottom of the image anymore. 

--> See above for screenshots

### 3. [Mobile] "show project" button is too short
The "show project" button is too short. According to the design it should span across the whole width of the screen.
##### Design

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/5fec95c9-b857-4f5e-8c31-a81dec1a95a5" />

##### Current Implementation

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/38fa799a-7ddb-40fd-be95-1ce695529eed" />

hom3mad3:
@MarleyMi could you confirm the blue panel issue has been fixed? i cannot reproduce anymore 🥳

MarleyMi:
Looks good

PR #6226, opened 2025-03-06
fix custom teaser

hom3mad3:
fixes https://github.com/liqd/a4-meinberlin/issues/6104

also fixed
- failing contrast for the badge
- deformed tile image


sevfurneaux (approved):

meinberlin/apps/cms/blocks.py
```diff
@@ -27,7 +27,7 @@ class ProjectTeaserBlock(blocks.StructBlock):
         help_text=_("Will be shown instead of the project title if set."),
     )
     custom_teaser = blocks.RichTextBlock(
-        max_length=500,
+        max_length=250,
```
vellip, line 30: Should we also truncate the project description in case no custom teaser is provided?

PR #6233, opened 2025-03-06
[#8866]: Add stats to module tiles

vellip:
**Describe your changes**
CleanShot 2025-03-06 at 12 49 30
Add stats to module tiles

**Tasks**
- [x] Changelog

### DIFF
--- a/meinberlin/apps/projects/templates/meinberlin_projects/includes/module-tile/module_insights.html
+++ b/meinberlin/apps/projects/templates/meinberlin_projects/includes/module-tile/module_insights.html
@@ -1,0 +1,5 @@
+{% load i18n meinberlin_project_tags %}
+
+<i class="module-insight__icon {{ icon }}" aria-hidden="true"></i>
+<span class="module-insight__count">{{ count }}</span>
+<span class="module-insight__label">{{ label }}</span>
--- a/meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
+++ b/meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
@@ -2,17 +2,16 @@ import itertools
 import re
 
 from django import template
-from django.db.models import Count
 from django.db.models import Q
-from django.db.models import Sum
 from django.utils.html import strip_tags
+from django.utils.translation import ngettext
 
 from adhocracy4.comments.models import Comment
 from adhocracy4.polls.models import Vote as Vote
-from meinberlin.apps.budgeting.models import Proposal as budget_proposal
+from adhocracy4.ratings.models import Rating
+from meinberlin.apps.budgeting.models import Proposal
 from meinberlin.apps.ideas.models import Idea
-from meinberlin.apps.kiezkasse.models import Proposal as kiezkasse_proposal
-from meinberlin.apps.likes.models import Like
+from meinberlin.apps.kiezkasse.models import Proposal as KKProposal
 from meinberlin.apps.livequestions.models import LiveQuestion
 from meinberlin.apps.mapideas.models import MapIdea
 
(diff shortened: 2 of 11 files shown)


meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
```diff
@@ -53,36 +52,70 @@ def get_sorted_modules(context):
...
-        | Q(chapter__module=module)
-        | Q(poll__module=module)
-    )
-    comment_count = (
-        Comment.objects.filter(comment_filter)
-        .annotate(child_comment_count=Count("child_comments__pk", distinct=True))
-        .aggregate(comment_count=Count("pk") + Sum("child_comment_count"))[
-            "comment_count"
-        ]
+@register.inclusion_tag("meinberlin_projects/includes/module-tile/module_insights.html")
```
m4ra, line 55: what's the benefit of adding the template here rather than in the parent template? I feel is more transparent and consitent to have the templates explicitly included within templates.

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
```diff
@@ -53,36 +52,70 @@ def get_sorted_modules(context):
...
+    for types, model in type_to_model_mapping.items():
+        if bp_type in types:
+            count = model.objects.filter(module=module).count()
+            if model == Proposal or model == KKProposal:
+                labels = ("Proposal", "Proposals")
+                icon = "fas fa-pen"
+            elif model == LiveQuestion:
+                labels = ("Question", "Questions")
+                icon = "fas fa-question"
+            break
```
m4ra, line 90: not sure why we need `break` here

meinberlin/apps/projects/templates/meinberlin_projects/includes/module-tile/module_insights.html
```diff
@@ -0,0 +1,7 @@
+{% load i18n meinberlin_project_tags %}
+
+{% if count >= 0 %}
+<i class="module-insight__icon {{ icon }}" aria-hidden="true"></i>
+<span class="module-insight__count">{{ count }}</span>
+<span class="module-insight__label">{{ label }}</span>
+{% endif %}
```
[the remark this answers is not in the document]
m4ra: Do the labels need to be translated? If that's the case, we need the `blocktranslate` here

tests/projects/test_project_tags.py
```diff
@@ -59,71 +64,131 @@ def test_is_a4_project(project, external_project, bplan):
...
     # no items or comments
     clear_query_cache()
-    with django_assert_num_queries(18):
-        count = get_num_entries(empty_module)
-        assert count == 0
+    with django_assert_num_queries(0):
+        data = render_module_insights(empty_module)
+        assert data["count"] == 0
+        assert data["label"] == ngettext("Idea", "Ideas", 0)
+        assert data["icon"] == "far fa-lightbulb"
```
m4ra, line 133: what's the empty module default? it's an idea? hence the icon and the label for ideas? if so can you add an inline comment when you create the empty_module.

tests/projects/test_project_tags.py
```diff
@@ -59,71 +64,131 @@ def test_is_a4_project(project, external_project, bplan):
...
-    phase, module, project, idea = setup_phase(
-        phase_factory, idea_factory, idea_phases.FeedbackPhase
+    phase_ic, module_ic, project_ic, idea = setup_phase(
+        phase_factory,
+        idea_factory,
+        idea_phases.FeedbackPhase,
+        module__blueprint_type="IC",
     )
+
     phase1, module1, project1, proposal = setup_phase(
```
m4ra, line 90: could you rename `module1` as you do with the other modules? `module_<type>`

m4ra (changes requested):
Thank you! Looks very good all in all! Just a few small things.

meinberlin/apps/projects/templates/meinberlin_projects/includes/module-tile/module_insights.html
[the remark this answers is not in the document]
vellip: I am translating the label inside the template_tag via `ngettext`. Good point with the `if` though, thanks!

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
vellip: You mean `{% include ... %}` vs `inclusion_tag`? I found it was a perfect case for an `inclusion_tag` because it always comes with its own `template` assigned, so it's more modular; in case you ever want to display this elsewhere, you could just call the tag.

Let me know however, if I should make the switch to `{% include %}`

meinberlin/apps/projects/templates/meinberlin_projects/includes/module-tile/module_insights.html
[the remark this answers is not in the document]
vellip: Thinking about it, since I initialize `count` with `0` it would always be true and I can also drop the `if`, currently the decision is we also want to show "0 Proposals".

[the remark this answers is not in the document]
m4ra: the ngettext `label` and `labels` are constructed dynamically though. When I generate the strings to be translated with `make po`, these `label/s` are not included.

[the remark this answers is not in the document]
m4ra: @vellip the blocktranslate tags don't help either. Might be that we need to include the labels' text in the template, either from the meta.class or hard coded, and use if/elif/else checks. Or if you have other ideas..

m4ra (changes requested):
translations are not generated for the dynamically created labels.

hom3mad3 (approved):
👍 love the clean templates, thanks for taking this over 🫶🏼

meinberlin/apps/projects/templatetags/meinberlin_project_tags.py
```diff
@@ -71,36 +70,72 @@ def has_ckeditor_content(value):
...
+        count = (
+            Vote.objects.filter(choice__question__poll__module=module)
+            .values("creator")
+            .distinct()
+            .count()
+        )
+        label = ngettext("Participant", "Participants", count)
+        icon = "fas fa-user-group"
+
+    return {"count": count, "label": label, "icon": icon}
```
m4ra, line 141: _noice_

PR #6242, opened 2025-03-07
filter-projects: fix bug where kiez filter only worked when there also was a district filter set

### LINKED ISSUE #6236: Kiezradar: filtering for a user created kiez doesn't work
**URL:** 
```/kiezradar/```

**user:** 
users with an account

**expected behaviour:** 
When users have created a personal Kiez in ```/account/kiezradar/``` they can use it in the Kiezradar to filter for projects in that area. 

**behaviour:** 
Selecting a user created kiez and applying the selection does not filter the list of projects and plans. Instead all projects and plans are shown, as if no filter has been applied.

### PR DESCRIPTION
**Pull request by vellip**
**Describe your changes**
fixes #6236

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/react/projects/filter-projects.js
+++ b/meinberlin/react/projects/filter-projects.js
@@ -28,21 +28,31 @@ export const filterProjects = (items, appliedFilters, kiezradars, topics, projec
         activeKiezradars.includes(kiezradar.name) &&
         getDistanceBetweenPoints(item.point.geometry.coordinates, kiezradar.point.geometry.coordinates) <= kiezradar.radius
       )
+    const hasKiezradarAndDistrict = activeKiezradars.length > 0 && districts.length > 0
 
-    return (activeTopics.length === 0 || activeTopics.some(topic => item.topics.includes(topic))) &&
-           (participations.length === 0 || participations.includes(item.participation)) &&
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by sevfurneaux — APPROVED** [review]

File: meinberlin/react/projects/filter-projects.js [hunk 1]
```diff
@@ -28,21 +28,31 @@ export const filterProjects = (items, appliedFilters, kiezradars, topics, projec
...
+      (search === '' ||
+        isInTitle(item.title, search) ||
+        isInTitle(item.district, search) ||
+        isInTitle(item.organisation, search) ||
+        isInTitle(item.description, search) ||
+        isInTopic(topics, item.topics, search)) &&
+      (projectState.includes(statusNames[item.status])) &&
+      (!plansOnly || item.type === 'plan') &&
+      (hasKiezradarAndDistrict
+        // if we have both active we want to include projects that are within kiez
```
sevfurneaux, line 46: @vellip Thanks for fixing this!

Appreciate the comment here, as admittedly I find this whole logic quite hard to read and understand 🧠. Maybe it can be encapsulated in a named variable in the future?

vellip: Definitely sounds like an idea!

Issue #6085, opened 2025-02-13
#8765 [mB] Redesign frontend: Opener: Image with Aside and detail Project – follow button is missing responsive behavior


sblindow:
**URL:** 
project detail page
https://meinberlin-design-dev.liqd.net/projekte/testing-opener-image-with-aside-and-detail-project/

**user:** 
Any user with an account.

**expected behaviour:** 
The follow button has the same behavior as the support button.
- black in default state
- turns dark green on hover
- light green when activated

See support button in participatory budgeting for reference: 
https://meinberlin-design-dev.liqd.net/budgeting/2025-01231/

See prototype here: 
https://www.figma.com/proto/vzwVPLittUtVRDfb8QBPR6/meinBerlin?node-id=3128-18454&p=f&t=VuvQ4v9XIGH69iCk-0&scaling=min-zoom&content-scaling=fixed&page-id=61%3A3106&starting-point-node-id=2698%3A21027&show-proto-sidebar=1

**behaviour:** 
The support button stays black when hovering and when activating.

Issue #6084, opened 2025-02-13
#8765 [mB] Redesign frontend: Opener: Image with Aside and detail Project – alert for follow button is missing


sblindow:
**URL:** 
project detail page
https://meinberlin-design-dev.liqd.net/projekte/testing-opener-image-with-aside-and-detail-project/

**user:** 
all users

**expected behaviour:** 
- Clicking on the “Follow” button triggers a success alert
- The alert when clicking on “Follow” should not interrupt navigation and should be easy to close

**behaviour:** 
success alert is missing

PR #6254, opened 2025-03-10
fix follow issues

hom3mad3:
fixes: https://github.com/liqd/a4-meinberlin/issues/6085 and https://github.com/liqd/a4-meinberlin/issues/6084


hom3mad3:
@vellip decided to not overwrite spacings above the alert, as to not affect the elements below. it was already enough work to get the alert outside the container 🙃 we can probably handle this after release if necessary

also, ended up not using a custom class, but left the enhancement in a4 anyways, since i think giving the option follows best practice.

vellip (approved):
Very kewl, thank you!

PR #6260, opened 2025-03-11
[#8956] style archived proposals

hom3mad3:
Screenshot 2025-03-11 at 14-25-10 Community Education   Idea Collection — meinBerlin

### DIFF
--- a/meinberlin/apps/contrib/templates/meinberlin_contrib/components/info-box.html
+++ b/meinberlin/apps/contrib/templates/meinberlin_contrib/components/info-box.html
@@ -1,11 +1,5 @@
 {% load i18n %}
-
-<div 
-  class="info-box" 
-  role="status" 
-  tabindex="0"
->
-  <p class="info-box__text">
-    {% translate info_message %}
-  </p>
+<div class="info-box" role="status" tabindex="0">
+    {% if info_title %}<h3 class="info-box__title">{{ info_title }}</h3>{% endif %}
+    <p class="info-box__text">{{ info_message }}</p>
 </div>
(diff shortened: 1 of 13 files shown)


meinberlin/apps/contrib/templates/meinberlin_contrib/components/info-box.html
```diff
@@ -5,6 +5,9 @@
   role="status" 
   tabindex="0"
 >
+  {% if info_title %}
+    <h3 class="info-box__title">{% translate info_title %}</h3>
+  {% endif %}
```
vellip, line 10: I think you have to do the `{% translate %}` in the template that uses this component. Otherwise `make po` will probably not understand that as a translateable string.

hom3mad3:
> LGTM, thank you so much! I just had one remark, and maybe something more general: I thought we are supposed to inline `html` files with 4 spaces, i think it says so in editorconfig, but you continued with 2, while I try to revert everything where i previously indented with 2 spaces. I guess I am not sure what do we want to use here.

oops, you're right. this is unintended.. i just automatically formatted w/ my editor and didn't bother to format it with our html linter.

meinberlin/apps/contrib/templates/meinberlin_contrib/components/info-box.html
hom3mad3: i have been adding translations to includes before, not sure what you mean.

hom3mad3: i think i get it, its because i'm translating the variable

Issue #6272, opened 2025-03-13
Card: shows 0 when there are no pills


hom3mad3:
**URL:** https://meinberlin-design-dev.liqd.net/projekte/module/themenpriorisierung-2/
**user:** *moderator/ initiator/ participants/ visitors (non-logged-in)/ admin/ cms-editors/ developers*
**expected behaviour:** no pills or 0's are shown where there are no pills in the card
**behaviour:** card displays a 0 at the top when there are no categories or labels

![Image](https://github.com/user-attachments/assets/abe59253-fca6-47a4-bf0a-8341a0b0aa1d)

PR #6274, opened 2025-03-13
react/contrib/Card.jsx: fix issue with undefined children

hom3mad3:
fixes: https://github.com/liqd/a4-meinberlin/issues/6272

### DIFF
--- a/meinberlin/react/contrib/card/Card.jsx
+++ b/meinberlin/react/contrib/card/Card.jsx
@@ -35,7 +35,7 @@ export const Card = (props) => {
           onMouseLeave={handleMouseLeave}
         >{item.name}
         </h3>
-        {children}
+        {children ? <>{children}</> : null}
       </header>
       <section className="card__footer--light">
         <CardStats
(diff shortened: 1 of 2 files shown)


meinberlin/react/contrib/card/Card.jsx
```diff
@@ -35,7 +35,7 @@ export const Card = (props) => {
           onMouseLeave={handleMouseLeave}
         >{item.name}
         </h3>
-        {children}
+        {children && <>{children}</>}
```
sevfurneaux, line 38: @hom3mad3 maybe you don't need the fragment `<></>` here? Does it work without it?

hom3mad3: I initially thought the fragment (<>...</>) might be unnecessary too, but in this case, removing it causes an issue. Since {children} might be null or undefined, wrapping it in a fragment ensures that React doesn’t break when rendering.

Also, I had to add an extra check in the card list to prevent a 0 from appearing above the card when there were no badges.

sevfurneaux (approved):

PR #6327, opened 2025-03-26
actions: fix error when TEST is not set in settings.

vellip:
**Describe your changes**
Had an error that appeared when creating an idea, when `TEST` is not set in settings.

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/notifications/signals.py
+++ b/meinberlin/apps/notifications/signals.py
@@ -17,8 +17,12 @@ User = get_user_model()
 
 @receiver(signals.post_save, sender=Action)
 def send_notifications(instance, created, **kwargs):
+    """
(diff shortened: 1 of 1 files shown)


meinberlin/apps/notifications/signals.py
```diff
@@ -18,7 +18,7 @@
 @receiver(signals.post_save, sender=Action)
 def send_notifications(instance, created, **kwargs):
     if created:
-        if settings.TEST:
+        if hasattr(settings, "TEST") and settings.TEST:
             send_action_notifications.delay(instance.pk)
```
m4ra, line 22: thanks. Also if settings has no attribute `TEST`, and we are in testing enviroment tests will fail because of the atomic transaction of the `delay_on_commit`. Maybe you can add a docstring to indicate that too?

PR #6336, opened 2025-03-28
[#9001] Add menu notification counter

sevfurneaux:
**Describe your changes**
This PR adds a menu notification counter.

### Page Header

Screenshot 2025-03-28 at 12 55 07

### Burger menu

Screenshot 2025-03-28 at 12 55 16

**Tasks**
- [ ] Changelog

### DIFF
--- a/meinberlin/apps/cms/templatetags/menu.py
+++ b/meinberlin/apps/cms/templatetags/menu.py
@@ -1,6 +1,7 @@
 from django import template
 
 from meinberlin.apps.cms import models as cms_models
+from meinberlin.apps.notifications.models import Notification
 
 register = template.Library()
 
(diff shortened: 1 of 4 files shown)


meinberlin/apps/cms/templatetags/menu.py
```diff
@@ -13,6 +14,21 @@ def get_header_menu(title):
...
 
+@register.simple_tag(takes_context=True)
+def get_notifications_count(context):
+    request = context.get("request")
+    if not request or not request.user.is_authenticated:
+        return 0
+
+    unread_count = (
+        Notification.objects.filter(recipient=request.user, read=False)
+        .exclude(action__actor=request.user)
```
vellip, line 25: I would propose to leave this out. In most cases it should not generate a notification in the first place if the user is the trigger.

vellip (approved):
Cool, cool! Looks good to me, just one proposal for change.

meinberlin/apps/cms/templatetags/menu.py
vellip: Otherwise we might run into inconsistencies.

sevfurneaux: @vellip Thanks! I just tested removing this:

`.exclude(action__actor=request.user)`

...and it seems to then be incorrect where it is showing 6 notifications rather than 4. So maybe we have to keep it in?

vellip: Okay, in that case leave it in

PR #6444, opened 2025-04-28
added command to clear cache

### LINKED ISSUE #6412: Add clearing of serializer caches to git workflow
When deploying changes, especially changes to any project or plan serializer, the cache keeps any old state across deployment. 

The git workflow should be changed in a way that it clears all the caches upon deploying to prevent any further headaches about weird issues or stale data.

Cache keys are:

Image

m4ra:
**Describe your changes**
fixes #6412 

**Tasks**
- [x] Changelog


m4ra:
@vellip we don't set cache for private projects, as far as I remember, and as far as I can see in the [tasks](https://github.com/liqd/a4-meinberlin/blob/dev/meinberlin/apps/projects/tasks.py) file. For prod and stage we need to add this command separately indeed.

PR #6511, opened 2025-07-23
apps/bplan: strip tags from description

partizipation:
Users making requests against bplan api sometimes include html tags for the description. This change strips tags, per [ST-780](https://app.asana.com/1/1175467295883992/inbox/1209159381077685/item/1210823403244953/story/1210845596960795)

For another day: The create/update methods that were already existing for this serializer are not DRY -- ideally the sanitation, truncation, etc. that is duplicated in these should be moved to the validation method I've created here (which create/update both receive validated data from), leaving side-effect-heavy logic (images, signals) in create/update.


glanzel (approved):
looks good

PR #6726, opened 2026-07-02
[ST-2252] add guest users

partizipation:
**Describe your changes**

**Tasks**
- [ ] Changelog

:.::..:::::: [Aligned Asana Task](https://app.asana.com/1/1175467295883992/project/1208559589495764/task/1215115726274530) ::.:.::.::::

### DIFF
--- a/meinberlin/apps/contrib/emails.py
+++ b/meinberlin/apps/contrib/emails.py
@@ -1,12 +1,42 @@
 from django.conf import settings
 
 from adhocracy4 import emails as a4_emails
+from adhocracy4.projects.guest_users import is_guest_user
 
 
 class Email(a4_emails.Email):
-    """Email base class with a configurable default language."""
+    """Email base class with a configurable default language.
+
+    Also drops guest users from the recipient list centrally: guests use
+    throwaway addresses and must never receive outbound mail. All meinBerlin
+    email classes inherit from this base, so filtering here covers every
+    outbound email (notifications, invites, newsletters, etc.).
+    """
 
     fallback_language = "en"
 
     def get_languages(self, receiver):
         return [settings.DEFAULT_LANGUAGE, self.fallback_language]
+
+    @staticmethod
+    def _exclude_guest_users(receivers):
+        """Return ``receivers`` without any guest users.
+
+        Receivers may be ``User`` instances or plain email strings; strings
+        (e.g. external notifications) are always kept.
+        """
+        return [
+            receiver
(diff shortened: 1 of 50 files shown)


meinberlin/apps/users/management/commands/delete_expired_guests.py
glanzel, line 1: should that maybe integrated as a celery beat at some point ?

glanzel (dismissed):

meinberlin/apps/users/management/commands/delete_expired_guests.py
partizipation: it's intended to be used with cron - was developed for roots but never actually flipped on. i made some improvements but need to discuss with @t-j-q

meinberlin/apps/contrib/emails.py
```diff
@@ -1,12 +1,42 @@
...
+    def _exclude_guest_users(receivers):
+        """Return ``receivers`` without any guest users.
+
+        Receivers may be ``User`` instances or plain email strings; strings
+        (e.g. external notifications) are always kept.
+        """
+        return [
+            receiver
+            for receiver in receivers
+            if not (hasattr(receiver, "email") and is_guest_user(receiver))
```
shn-liqd, line 31: why the first condition?

meinberlin/apps/projects/rules.py
```diff
@@ -33,7 +34,7 @@
     | is_initiator
     | is_moderator
     | is_prj_group_member
-    | ((is_public | is_project_member) & is_live),
+    | ((is_public | is_project_member) & is_live & guest_may_participate),
```
shn-liqd, line 36: Should it be an extra or condition?

partizipation: don't think so. `guest_may_participate` returns True for non-guests and project.allow_guest_users for guests, so the & only gates guests while leaving regular users unaffected.

the name for this is bad, would be better as something like `is_regular_user_or_guest_allowed` but would need change in a4 and a+ at this point. adding a comment at least.

meinberlin/apps/contrib/emails.py
partizipation: sometimes receiver can be a plain string, sometimes a user object, depending on the type of invite. this is meant as a way to avoid an attribute error if is_guest_user were to be called on a plain string

shn-liqd (approved):

